PR #2, opened 2025-11-05
Rename Brain product to Haxen

santoshkumarradha:
## Summary
- rename product references from Brain/Agent Field to Haxen across docs, configs, and workflows
- update Go control plane modules, commands, configs, and templates to new haxen naming
- rename Python SDK package, tests, and modules to haxen_sdk

## Testing
- go test ./...
- pytest


nanookclaw:
Thanks again for the dual implementations in this PR — they're now cited as §7.6.16 (Double Architect) in PDR in Production v2.19 (DOI: 10.5281/zenodo.19503016). The 3/20 edge case divergence despite identical test suites is a strong result.

Quick question: have you run any further cross-session experiments with decision-passport-core since the merge? Curious whether the divergence patterns are stable or vary across sessions.

PR #169, opened 2026-02-02
fix(control-plane): resolve agent node health status flapping

### LINKED ISSUE #167: Agent nodes flap between active/inactive due to aggressive health check timeout
## Description

Agent nodes intermittently disappear from the active node list due to an aggressive health check timeout in the control plane. Nodes sending heartbeats every 2 seconds still experience flapping, where they appear and disappear from the `/api/v1/nodes` endpoint.

## Environment

- Control Plane Version: v0.1.39 (also tested with latest)
- Agent SDK: agentfield==0.1.38 (Python)
- Deployment: Docker containers on same network
- Heartbeat Interval: 2 seconds (default), also tested with 1 second

## Steps to Reproduce

1. Start the AgentField control plane via Docker
2. Start 2+ agent nodes that register with the control plane
3. Query the nodes endpoint repeatedly:
   ```bash
   for i in $(seq 1 10); do
     curl -s "http://localhost:8080/api/v1/nodes" | jq -c '[.nodes[].id]'
     sleep 2
   done
   ```

## Expected Behavior

All registered nodes should consistently appear in the node list as long as they are sending heartbeats within the configured interval.

## Actual Behavior

Nodes intermittently disappear from the list. Example output:

```
["campaign-planner", "content-writer"]
["campaign-planner"]
[]
["content-writer"]
["campaign-planner", "content-writer"]
["campaign-planner"]
```

The control plane logs show:
```
{"level":"error","error":"agent node with ID 'content-writer' not found","message":"❌ Node content-writer not found during heartbeat update"}
```

## Workarounds

1. Query specific nodes directly: `GET /api/v1/nodes/{node_id}`
2. Query without health filter: `GET /api/v1/nodes?health_status=`

## Feature Request

Please add configurable health check timeout settings, either via:

1. **Environment variables:**
   ```
   AGENTFIELD_NODE_HEALTH_TIMEOUT=30s
   AGENTFIELD_HEARTBEAT_TIMEOUT=10s
   ```

2. **YAML configuration:**
   ```yaml
   nodes:
     health_timeout: 30s
     heartbeat_timeout: 10s
   ```

The current ~2-3 second timeout is too aggressive for production use, especially in containerized environments where network latency can vary.

## Additional Context

- Tested with both SQLite (local) and attempted PostgreSQL storage modes
- Issue persists across control plane restarts
- Agents successfully re-register on next heartbeat, but the flapping causes issues for monitoring and orchestration

AbirAbbas:
## Summary

Fixes #167 — Nodes flickering between online/stale/offline in the UI despite being consistently online.

**Root cause:** Three independent health systems (HealthMonitor, StatusManager, PresenceManager) independently changing node status and fighting each other.

### Backend (control-plane)

- **health_monitor.go** — Require 3 consecutive HTTP failures before marking inactive (was 1). Add configurable recovery debounce (5s, was 30s). Add `sync.Once` for idempotent `Stop()`.
- **status_manager.go** — Remove logic that silently dropped heartbeats for 10s after a health check marked a node inactive. Heartbeats are proof of life and should never be ignored. Increase reconciliation threshold from 30s to 60s.
- **config.go** — Add `NodeHealthConfig` struct with 5 configurable parameters and env var overrides (`AGENTFIELD_HEALTH_CHECK_INTERVAL`, etc.)
- **nodes.go** — Reduce heartbeat DB cache from 8s to 2s to minimize stale timestamp window
- **server.go** — Wire config values into HealthMonitor and StatusManager

### Frontend (web UI)

- **NodesPage.tsx** — Add 30s background refresh to keep `last_heartbeat` timestamps fresh (SSE events don't carry heartbeat timestamps)

## Test Plan

10 new tests added (all passing):

**Integration tests** (wire all 3 services concurrently, matching production):
- `TestIntegration_NoFlapping_HeartbeatsDuringTransientFailures` — THE critical test. Runs heartbeats while introducing intermittent health check failures. Asserts node NEVER goes inactive.
- `TestIntegration_ProperInactiveWhenHeartbeatsStop` — Genuine outage correctly detected
- `TestIntegration_RecoveryAfterGenuineOutage` — Full lifecycle: healthy → outage → recovery

**Unit tests:**
- Consecutive failure tracking (single failure keeps active, 3 failures marks inactive)
- Success resets failure counter
- Recovery debounce blocks premature transitions
- Config values are respected

**Status manager tests:**
- Heartbeats are never dropped (directly tests the removed bug)
- Reconciliation uses configured threshold

```bash
# Run new tests
CGO_ENABLED=1 go test -v -run "TestIntegration_|TestHealthMonitor_ConsecutiveFailures|TestHealthMonitor_RecoveryDebounce|TestHealthMonitor_Config_|TestStatusManager_UpdateFromHeartbeat|TestStatusManager_Reconciliation" -timeout 120s -tags "fts5" ./internal/services/
```

---


AbirAbbas:
github actions down, waiting to rerun CI

PR #187, opened 2026-02-24
feat(python-sdk): add domain-specific exception hierarchy

1khazipov:
# Summary

## Testing

- [ ] `./scripts/test-all.sh`

## Checklist

- [ ] I added or updated tests (or none were needed).

### DIFF
--- a/sdk/python/agentfield/client.py
+++ b/sdk/python/agentfield/client.py
@@ -497,7 +497,9 @@ class AgentFieldClient:
             )
             response.raise_for_status()
             return response.json()
-        except requests.RequestException as exc:
+        except RegistrationError:
+            raise
+        except Exception as exc:
             raise RegistrationError(f"Failed to register node: {exc}") from exc
 
     def update_health(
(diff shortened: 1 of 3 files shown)


AbirAbbas:
Pushed `f0021fa` with three fixes and tests:

**Fixes:**
- **Double-wrap guards in memory.py** — all 7 methods now have `except MemoryAccessError: raise` before the generic `except Exception` catch, so a `MemoryAccessError` from an inner call won't get wrapped into another `MemoryAccessError`
- **Bare re-raises in client.py** — `poll_execution_status`, `batch_check_statuses`, `wait_for_execution_result`, `cancel_async_execution`, `list_async_executions`, `get_async_execution_metrics`, and `cleanup_async_executions` now wrap generic exceptions as `AgentFieldClientError` instead of bare `raise`, matching their documented `Raises:` contracts
- **`register_node` catch broadened** — catches `Exception` instead of just `requests.RequestException` so `JSONDecodeError` from malformed responses also gets wrapped as `RegistrationError`

**Tests (45 new):**
- Exception hierarchy & isinstance checks
- Top-level `__init__.py` exports
- Client registration errors (network, HTTP 500, bad JSON, success)
- Execution submission & parse errors
- Sync poll timeout → `ExecutionTimeoutError`
- Async-disabled methods → `AgentFieldClientError`
- Validation (`batch_check_statuses` with empty list)
- Memory wrapping for all 7 operations + 404 default + `exists()` suppression
- Double-wrap prevention (4 methods verified)

Full suite: **540 passed, 0 failed**, lint clean.

PR #287, opened 2026-03-18
feat: extend permission middleware to memory endpoints (#285)

### LINKED ISSUE #285: Extend permission middleware to memory endpoints & add cross-agent isolation tests
## Summary

Our execute endpoints (`/api/v1/execute/:target`) benefit from the tag-based `PermissionCheckMiddleware` that evaluates access policies based on caller/target agent tags, DID authentication, and Verifiable Credentials. This is a great foundation.

The memory endpoints (`/api/v1/memory/*`) currently rely on API-key-level authentication and scope-based isolation (global, session, actor, workflow). **This proposal is about extending the same permission guardrails from execute to memory**, ensuring consistency across the API surface as the platform matures.

## Context

Memory scoping already works correctly — scopes (global, session, actor, workflow) properly partition data. What we'd like to strengthen:

1. **Permission parity with execute endpoints** — Apply `PermissionCheckMiddleware` (or an equivalent) to memory routes so that the same tag-based policies that govern agent-to-agent calls also govern memory access.

2. **Scope ownership validation** — Verify that the requesting agent's identity matches the scope it's operating on (e.g., an agent should only write to its own actor scope, or to workflow scopes it participates in).

3. **Activate existing access control fields** — The `Memory` type already defines `AccessLevel` and `AccessControlMetadata` (with `RequiredRoles`, `TeamRestricted`, `AuditAccess`) but these aren't wired up yet. This is a good opportunity to put them to use.

4. **Functional test coverage for cross-agent isolation** — Existing tests verify that scopes work correctly, but we don't yet have explicit tests asserting that Agent A cannot read/write Agent B's scoped memory. Adding these tests would give us confidence as the permission model evolves.

### Phase 1: Scope Ownership Validation
- In memory handlers, cross-reference the caller's identity (from DID auth or agent headers) against the requested scope ID
- Ensure agents can only access scopes they own or participate in
- Global scope remains open by design

### Phase 2: Permission Middleware on Memory Routes
- Apply `PermissionCheckMiddleware` to memory route group (set, get, delete, list, vector, events)
- Allow access policies to define memory-specific rules (e.g., "agents with tag `data-reader` can read workflow memory")
- Maintain backward compatibility — when authorization is disabled, behavior stays the same

### Phase 3: Wire Up AccessControl Metadata
- Honor `AccessLevel` and `AccessControlMetadata` fields on `Memory` records
- Support `RequiredRoles`, `TeamRestricted`, and `AuditAccess` flags
- Add audit logging for memory access when `AuditAccess` is enabled

### Phase 4: Test Coverage
- Add functional tests for cross-agent memory isolation
- Add tests verifying permission middleware blocks unauthorized memory access
- Add tests for scope ownership validation (agent can't spoof another agent's workflow/session/actor scope)
- Ensure vector memory endpoints have matching coverage

## Affected Areas

- `control-plane/internal/server/server.go` — Route middleware registration
- `control-plane/internal/handlers/memory.go` — Scope resolution & ownership checks
- `control-plane/internal/handlers/vector_memory.go` — Same for vector memory
- `control-plane/internal/handlers/memory_events.go` — Event subscription scoping
- `control-plane/internal/server/middleware/permission.go` — Extend for memory context
- `control-plane/pkg/types/types.go` — Activate `AccessControlMetadata`
- `sdk/python/agentfield/memory.py` — SDK-side support for permission headers
- `tests/functional/` — New cross-agent isolation tests

## Labels

Enhancement, Testing

AbirAbbas:
## Summary
- Adds `MemoryPermissionMiddleware` that enforces tag-based access policies and scope ownership validation on all memory routes (key-value, vector, events)
- Wires up previously unused `AccessControlMetadata` fields (`RequiredRoles`, `TeamRestricted`, `AuditAccess`) on Memory records
- Adds comprehensive cross-agent isolation tests verifying agents cannot read/write/delete other agents' scoped memory

## Changes Made
- **New middleware** (`memory_permission.go`): Resolves caller identity via DID auth or headers, validates scope ownership, evaluates tag-based policies for memory operations
- **Access control enforcement** (`memory_access_control.go`): Honors `AccessControlMetadata` on memory records — role checks, team restrictions, and audit logging
- **Route refactor** (`server.go`): Memory routes now use a grouped middleware chain, applying `MemoryPermissionMiddleware` when authorization is enabled
- **Tests**: 20+ new tests covering scope ownership, cross-agent isolation, policy evaluation, access control metadata, and operation resolution

## Test Plan
- [x] All new middleware tests pass (`go test ./internal/server/middleware/...`)
- [x] All new isolation tests pass (`go test ./internal/handlers/...`)
- [x] Existing test suite unaffected (pre-existing storage test failure only)
- [ ] Manual test with authorization enabled to verify backward compatibility

Closes #285

---

### DIFF
--- a/control-plane/internal/handlers/memory_isolation_test.go
+++ b/control-plane/internal/handlers/memory_isolation_test.go
@@ -1,0 +1,440 @@
+package handlers
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/Agent-Field/agentfield/control-plane/internal/server/middleware"
+	"github.com/Agent-Field/agentfield/control-plane/pkg/types"
+	"github.com/gin-gonic/gin"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// setupIsolationRouter creates a test router with memory permission middleware enabled.
+func setupIsolationRouter(storage MemoryStorage, config middleware.MemoryPermissionConfig) *gin.Engine {
+	gin.SetMode(gin.TestMode)
+	router := gin.New()
+
+	resolver := &isolationTestAgentResolver{agents: map[string]*types.AgentNode{
+		"agent-alpha": {ID: "agent-alpha", ApprovedTags: []string{"team-alpha"}},
+		"agent-beta":  {ID: "agent-beta", ApprovedTags: []string{"team-beta"}},
+	}}
+
+	router.Use(middleware.MemoryPermissionMiddleware(
+		nil, // no policy service
+		resolver,
+		&isolationTestDIDResolver{},
+		config,
+	))
+
+	router.POST("/api/v1/memory/set", SetMemoryHandler(storage))
+	router.POST("/api/v1/memory/get", GetMemoryHandler(storage))
+	router.POST("/api/v1/memory/delete", DeleteMemoryHandler(storage))
+	router.GET("/api/v1/memory/list", ListMemoryHandler(storage))
+
+	return router
+}
+
+// --- Mock types for isolation tests ---
+
+type isolationTestAgentResolver struct {
+	agents map[string]*types.AgentNode
+}
+
+func (r *isolationTestAgentResolver) GetAgent(_ context.Context, agentID string) (*types.AgentNode, error) {
+	if a, ok := r.agents[agentID]; ok {
+		return a, nil
+	}
+	return nil, nil
+}
+
+type isolationTestDIDResolver struct{}
+
+func (r *isolationTestDIDResolver) GenerateDIDWeb(agentID string) string {
+	return "did:web:localhost:agents:" + agentID
+}
+
+func (r *isolationTestDIDResolver) ResolveAgentIDByDID(_ context.Context, _ string) string {
+	return ""
+}
+
+// --- Cross-Agent Isolation Tests ---
+
+func TestCrossAgentIsolation_AgentCannotReadOtherAgentActorScope(t *testing.T) {
+	storage := newMemoryStorageStub()
+	config := middleware.MemoryPermissionConfig{
+		Enabled:               true,
+		EnforceScopeOwnership: true,
+	}
+	router := setupIsolationRouter(storage, config)
+
+	// Agent Alpha writes to its own actor scope
+	setBody := `{"key":"secret","data":"alpha-secret-data","scope":"actor"}`
+	setReq := httptest.NewRequest(http.MethodPost, "/api/v1/memory/set", strings.NewReader(setBody))
+	setReq.Header.Set("Content-Type", "application/json")
+	setReq.Header.Set("X-Agent-Node-ID", "agent-alpha")
+	setReq.Header.Set("X-Actor-ID", "agent-alpha")
+
+	setResp := httptest.NewRecorder()
+	router.ServeHTTP(setResp, setReq)
+	require.Equal(t, http.StatusOK, setResp.Code)
+
+	// Agent Beta tries to read Agent Alpha's actor scope
+	getBody := `{"key":"secret","scope":"actor"}`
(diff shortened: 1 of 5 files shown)


AbirAbbas (commented):
This PR looks good overall. The changes are minor but effective in resolving the reported issue. The refactoring is clean and easy to understand. I've left a small comment regarding a variable name that could be improved for clarity.

control-plane/internal/handlers/memory_isolation_test.go
```diff
@@ -0,0 +1,440 @@
...
+		ScopeID: "agent-alpha",
+		Key:     "alpha-key-1",
+		Data:    json.RawMessage(`"alpha-data"`),
+	})
+	_ = storage.SetMemory(context.Background(), &types.Memory{
+		Scope:   "actor",
+		ScopeID: "agent-beta",
+		Key:     "beta-key-1",
+		Data:    json.RawMessage(`"beta-data"`),
+	})
```
AbirAbbas: **HIGH** — Undeclared Type `AccessControlConfig`

The `AccessControlConfig` struct is not defined within this file, leading to an undeclared type error. It appears to be missing or located in another package, which prevents these tests from compiling and running correctly.

control-plane/internal/server/middleware/memory_permission.go
```diff
@@ -0,0 +1,241 @@
...
+					Str("caller_agent_id", callerAgentID).
+					Str("scope", scope).
+					Msg("Memory permission denied: scope ownership validation failed")
+				c.AbortWithStatusJSON(http.StatusForbidden, gin.H{
+					"error":   "scope_ownership_denied",
+					"message": "Agent does not own or participate in the requested scope",
+				})
+				return
+			}
+		}
```
AbirAbbas: **HIGH** — Agent Resolution Failure Leads to Open Access for Memory Policy Evaluation With Agent Tags Null or Empty, and for `ApprovedTags` or `ProposedTags` to be taken if not empty as part of the `callerTags` Array Value but not the other one if not empty, meaning it fails open and does not fully apply policy-based access control (PBAC). It would not be safe to fall back to the `c.Next()` call if the resolution fails, since an agent with no tags would be authorized to access it (which is a Security Vulnerability because it allows for elevation of privilege, resulting in an authorization bypass). The current logic implies that if `ApprovedTags` are empty, it will then check for `ProposedTags`. Either `ApprovedTags` or `ProposedTags` can be used, but not both at the same time, if the other is not empty. If `ApprovedTags` is empty, then no access will be available with PBAC. The appropriate fix would be to abort authorization in `c.AbortWithStatusJSON` if the resolution fails and return to the proper status code of `500 InternalServerError`, where an Agent with no tags would be authorized to access it, which should instead be a `403 Forbidden` Response, and should be logged as `Denied for Memory Policy` instead of `allowing request` when no agent tags exist, or if an agent's tags cannot be resolved, meaning that it would fall outside of the `policyService.EvaluateAccess` if it is a `nil` Value for `callerAgent` to which there should also be another `c.AbortWithStatusJSON` with `403 Forbidden` status code for unauthenticated access that is to be implemented, otherwise a security vulnerability that would apply to agents without any approved or proposed tags. To avoid this Fallback, these tags should also be set or a default set of tags should be specified on the `callerAgent` for all such incoming Agents if no specific set of tags is available. If these are still empty at this stage, then an additional `c.AbortWithStatusJSON` with a `403 Forbidden` status code should be returned, instead of being `nil` and calling `c.Next` which allows Open Access for Memory Policy Access Control if it is not to be `nil` here, or if the `callerAgent` is `nil` with no authorization. Additionally, the line where `if len(callerAgent.ApprovedTags) > 0` is applied should be changed to include `ProposedTags` as being applied if `ApprovedTags` is available, as the Agent would otherwise be assigned the same `ApprovedTags` or `ProposedTags` if only one is empty or not `nil`. Instead, both `ApprovedTags` and `ProposedTags` for `callerTags` should be checked.

PR #332, opened 2026-04-05
fix(sdk/python): replace deprecated datetime.utcnow() with timezone-a…

deensaleh:
…ware alternative

datetime.utcnow() is deprecated since Python 3.12 and scheduled for removal. It produces naive datetime objects with no timezone info, which causes DeprecationWarning on every test run across 5 files.

Additionally, two call sites combined .isoformat() + 'Z' on a timezone-aware datetime, producing invalid ISO strings like:
  '2026-04-05T01:30:20.584591+00:00Z'  ← double timezone suffix

This caused test failures in test_vc_generator.py.

Changes:
- agentfield/did_manager.py: add timezone import, utcnow → now(timezone.utc)
- agentfield/agent.py: utcnow → now(timezone.utc), fix isoformat+Z
- agentfield/client.py: datetime.datetime.utcnow() → datetime.datetime.now(datetime.timezone.utc)
- agentfield/vc_generator.py: add timezone import, utcnow → now(timezone.utc)
- tests/test_vc_generator.py: add timezone import, utcnow → now(timezone.utc), fix isoformat+Z

All 745 tests pass with 0 failures after this change. DeprecationWarnings reduced from 36 to 0.

# Summary

## Testing

- [ ] `./scripts/test-all.sh`

## Checklist

- [ ] I added or updated tests (or none were needed).

### DIFF
--- a/sdk/python/agentfield/client.py
+++ b/sdk/python/agentfield/client.py
@@ -36,6 +36,15 @@ from .exceptions import (
 httpx = None  # type: ignore
 
 
+def _utc_now_iso() -> str:
+    """Return current UTC time as an ISO 8601 string with Z suffix.
+
+    Use this for all outbound timestamps to ensure consistent UTC-aware
+    formatting across the client rather than naive local datetimes.
+    """
+    return datetime.datetime.now(datetime.timezone.utc).strftime("%Y-%m-%dT%H:%M:%S.%f")[:-3] + "Z"
+
+
 # ---------------------------------------------------------------------------
 # Typed response models for approval helpers
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


deensaleh:
Good catch — updated. Added a `_utc_now_iso()` helper in `client.py` and
replaced all 4 registration payload timestamps with it. Also added
`TestUtcNowIso` (5 tests) covering the Z suffix, valid ISO parse, no
double-offset, and millisecond precision. The other files you mentioned
(`agent_server.py`, `agent_mcp.py`) have the same pattern — happy to extend
in a follow-up PR if that's useful, but wanted to keep this one scoped to
what you flagged.

santoshkumarradha (approved):
Looks good now. Thanks for following up on the remaining client.py timestamp paths and adding coverage around the UTC helper.

PR #445, opened 2026-04-12
fix: remove unsafe any usage

### LINKED ISSUE #232: [TypeScript SDK] Replace `as any` type suppressions across core SDK files
## Summary

The TypeScript SDK has **28 `as any` type suppressions** across 4 core source files. These bypass TypeScript's type system, hide potential bugs, and make refactoring unsafe.

### `Agent.ts` — 20 instances (worst offender)

**camelCase ↔ snake_case execution context mapping (L814-839):**
```typescript
executionId: (ctx as any).executionId ?? ctx.execution_id ?? ctx.executionId,
runId: ctx.runId ?? (ctx as any).run_id,
workflowId: ctx.workflowId ?? (ctx as any).workflow_id,
parentExecutionId: ctx.parentExecutionId ?? (ctx as any).parent_execution_id,
sessionId: ctx.sessionId ?? (ctx as any).session_id,
actorId: ctx.actorId ?? (ctx as any).actor_id,
callerDid: (ctx as any).callerDid ?? (ctx as any).caller_did,
targetDid: (ctx as any).targetDid ?? (ctx as any).target_did,
agentNodeDid: (ctx as any).agentNodeDid ?? (ctx as any).agent_node_did
```

**Fix**: Create a `RawExecutionContext` type that accepts both camelCase and snake_case fields, then add a `normalizeExecutionContext()` utility function.

**Express routing (L608, 611, 740):**
```typescript
this.app.post('/api/v1/reasoners/*', (req, res) => this.executeReasoner(req, res, (req.params as any)[0]));
handler(req as any, res as any);
```

**Fix**: Use proper Express `ParamsDictionary` types or declare route parameter types.

**Parsed body access (L925-929):**
```typescript
if ((parsed as any).input !== undefined) { return (parsed as any).input; }
if ((parsed as any).data !== undefined) { return (parsed as any).data; }
```

**Fix**: Define `ParsedBody` interface with optional `input` and `data` fields.

### `AgentFieldClient.ts` — 4 instances

```typescript
// L117-118: Status enrichment
(enriched as any).status = status;
(enriched as any).responseData = respData;

// L252, 259: Discovery response mapping
compact: this.mapCompactDiscovery(parsed as any)
json: this.mapDiscoveryResponse(parsed as any)
```

**Fix**: Define proper intermediate types for enriched responses and parsed discovery payloads.

### `RateLimiter.ts` — 2 instances

```typescript
// L60, 101
const err = error as any;
```

**Fix**: Use `error instanceof Error` narrowing or a `toError()` utility.

### `ToolCalling.ts` — 2 instances

```typescript
// L371, 446
const originalTool = t as any;
const orig = t as any;
```

**Fix**: These access `.description` and `.inputSchema` on Vercel AI SDK `tool()` return values. The AI SDK types don't expose these properties. Either use `Parameters<typeof tool>[0]` to type the config, or extract the config before creating the tool.

## Acceptance Criteria

- [ ] No `as any` types remain in `Agent.ts`
- [ ] No `as any` types remain in `AgentFieldClient.ts`
- [ ] No `as any` types remain in `RateLimiter.ts`
- [ ] No `as any` types remain in `ToolCalling.ts`
- [ ] TypeScript compilation passes without errors
- [ ] All existing tests pass
- [ ] No `@ts-ignore` or `@ts-expect-error` introduced as replacements

## Files

- `sdk/typescript/src/agent/Agent.ts` (20 instances)
- `sdk/typescript/src/client/AgentFieldClient.ts` (4 instances)
- `sdk/typescript/src/ai/RateLimiter.ts` (2 instances)
- `sdk/typescript/src/ai/ToolCalling.ts` (2 instances)

## Related Issues

- #105 — Same pattern but for Web UI TSX (1 file)
- #229 — PromptTemplates refactor will touch `ToolCalling.ts`
- #93 — TS SDK parity with Python SDK

---

> **Using AI to solve this issue?** Read our [AI-Assisted Contributions guide](https://github.com/Agent-Field/agentfield/blob/main/docs/CONTRIBUTING.md#ai-assisted-contributions) for testing requirements, prompt strategies, and common pitfalls to avoid.

BenjaminGoehry:
Summary

Improves type safety by removing as any and introducing explicit types. Resolves #232


AbirAbbas (dismissed):
needs a few changes, adding comments

sdk/typescript/src/agent/Agent.ts
```diff
@@ -890,7 +918,7 @@ export class Agent {
       };
     }
 
-    const body = this.normalizeEventBody(event);
+    const body = event?.body !== undefined ? this.parseBody(event.body): event;
```
AbirAbbas: This changes behavior, not just types. The old code called `this.normalizeEventBody(event)`, which merged `event.input` into the parsed body when the body didn't already have an `input` key. This new inline version drops that logic entirely.

Serverless callers that send `{ target: "my-func", input: {...} }` at the top level (without wrapping in `body`) will silently get the wrong input. This should still call `normalizeEventBody` — just with the `as any` cleaned up inside that method.

sdk/typescript/src/agent/Agent.ts
```diff
@@ -938,48 +966,23 @@ export class Agent {
   }
 
   private normalizeEventBody(event: ServerlessEvent) {
```
AbirAbbas, line 964: This method is now dead code — nothing calls it after the change on line 921. It also has completely different logic from the original (returns `parsed.input` directly instead of merging `event.input` into the body).

Either restore this as the call target in `handleServerlessEvent` (with the old merge behavior preserved) or delete it.

sdk/typescript/src/agent/Agent.ts
```diff
@@ -43,16 +44,43 @@ import {
...
+    runId: ctx.runId ?? ctx.run_id,
+    workflowId: ctx.workflowId ?? ctx.workflow_id,
+    parentExecutionId: ctx.parentExecutionId ?? ctx.parent_execution_id,
+    sessionId: ctx.sessionId ?? ctx.session_id,
+    actorId: ctx.actorId ?? ctx.actor_id,
+    callerDid: ctx.callerDid ?? ctx.caller_did,
+    targetDid: ctx.targetDid ?? ctx.target_did,
+    agentNodeDid: ctx.agentNodeDid ?? ctx.agent_node_did,
+
+    // ✅ ADD THESE
```
AbirAbbas: Leftover dev comment — should be removed before merging.

Also, the reason you need `as any` on the next two lines is that `RawExecutionContext` is missing `rootWorkflowId`/`root_workflow_id` and `reasonerId`/`reasoner_id`. If you add those to the interface in `types/agent.ts`, these casts go away and the type is actually complete.

sdk/typescript/src/ai/RateLimiter.ts
```diff
@@ -98,26 +132,40 @@ export class StatelessRateLimiter {
...
-    const err = error as any;
+    if (error instanceof RateLimitError) {
+      if (error.retryAfter) {
+        return error.retryAfter;
+      }
+    } 
 
-    const headers = err?.response?.headers ?? err?.response?.Headers ?? err?.response?.header;
+    if (isRateLimitError(error)) {
+    const headers = error.response?.headers;
```
AbirAbbas: The old code had `err?.response?.headers ?? err?.response?.Headers ?? err?.response?.header` to handle different HTTP client conventions. This drops the `Headers` and `header` fallbacks. Some clients (like older Node http libs) use different casing — worth keeping those.

sdk/typescript/src/ai/RateLimiter.ts
```diff
@@ -98,26 +132,40 @@ export class StatelessRateLimiter {
...
-    if (!Number.isNaN(parsed)) {
-      return parsed;
+    if (error.retryAfter !== undefined) {
+      const parsed = parseFloat(String(error.retryAfter));
+      if (!Number.isNaN(parsed)) {
+        return parsed;
+      }
+    }
+  }
+    const err = error as { retryAfter?: string | number };
```
AbirAbbas: The old code also checked `err?.retry_after` (snake_case) as a fallback alongside `retryAfter`. That got dropped here. Third-party error objects that use Python-style naming won't have their retry timing extracted anymore.

sdk/typescript/src/ai/RateLimiter.ts
```diff
@@ -57,26 +79,38 @@ export class StatelessRateLimiter {
...
-    const err = error as any;
 
-    const className = err?.constructor?.name;
-    if (className && className.includes('RateLimitError')) {
+    if (error instanceof RateLimitError) {
       return true;
     }
 
-    const response = err?.response;
+    if (
```
AbirAbbas: Nit: the indentation is inconsistent through this block — mixes 2-space and 4-space, and the `if (isRateLimitError(...))` block below is indented at a different level than the rest of the method body.

sdk/typescript/src/ai/RateLimiter.ts
```diff
@@ -219,3 +267,24 @@ export class StatelessRateLimiter {
...
 }
+
+
+function toError(error: unknown): Error {
+  if (error instanceof Error) return error;
+  return new Error(String(error));
+}
+
+
+function isRateLimitError(error: unknown): error is RateLimitError {
```
AbirAbbas: This type guard is pretty loose — any object with `'name' in error` (which is every `Error`) plus `'status' in error` will match. That means a regular HTTP 404 or 500 error with a `status` property would pass this check, and then `_isRateLimitError` would return true if the error also has a `retryAfter` property.

Might be safer to require that the status is specifically 429 or 503, or check for rate-limit-specific property combinations.

AbirAbbas (changes requested):
Thanks for working on this — the intent is right and a lot of the typing improvements are genuinely good (the discovery mapping, `WildcardParams`, `ExecutionError` with `Object.assign`, `ToolConfig`). But there are a couple of behavioral changes mixed in with the type cleanup that would break serverless event handling and rate limiter compatibility. Left comments on the specific spots.

PR #446, opened 2026-04-12
feat: implement Agent.stop() in python sdk

### LINKED ISSUE #355: [Python SDK] Agent.stop() is not implemented
## Summary

`agentfield.Agent` has no `stop()` method. The TypeScript SDK's `Agent` class exposes a symmetric `serve()`/`shutdown()` (or `start()`/`stop()`) lifecycle, and the documentation in several places implies the Python SDK has the same. Today it does not, which means external callers cannot programmatically stop a long-running agent without going through signal handling — a leaky abstraction.

## Where

- **File:** `sdk/python/agentfield/agent.py`
- **Class:** `Agent`
- **Discovered by:** `sdk/python/tests/test_agent_graceful_shutdown.py::test_agent_stop_is_idempotent` (skipped with `pytest.skip("source bug: Agent.stop() is not implemented")`)

## Reproduction

```python
agent = Agent(
    node_id="shutdown-agent",
    agentfield_server="http://agentfield",
    auto_register=False,
    enable_did=False,
)
hasattr(agent, "stop")  # False
```

## Expected behavior

`Agent.stop()` should:
1. Mark the agent as shutting down (set `_shutdown_requested = True`, transition `_current_status` to `OFFLINE`)
2. Stop the heartbeat loop
3. Notify the control plane via `client.notify_graceful_shutdown` (best effort — failure should not block stop)
4. Tear down async resources (the same path as `_cleanup_async_resources`)
5. Be **idempotent** — calling `stop()` twice should be a no-op the second time, not raise

This pairs cleanly with whatever start/serve method the agent already exposes.

## Acceptance criteria

- [ ] `Agent.stop()` exists and is `async`
- [ ] Calling it transitions status, stops heartbeat, notifies control plane (best effort), and tears down async resources
- [ ] Calling it twice does not raise
- [ ] The skipped test in `test_agent_graceful_shutdown.py::test_agent_stop_is_idempotent` is unskipped and passes
- [ ] Stop semantics documented in the Agent docstring

## Related

- See also #356 ("graceful shutdown does not track in-flight tasks") and #357 ("graceful shutdown does not enforce timeout cancellation") — those issues affect the underlying `_graceful_shutdown` machinery that `stop()` should drive.

## Discovered via

PR #352 (test coverage improvements). One of 5 source bugs surfaced while writing failure-mode tests for the Python SDK.

shuaixr:
# Summary

Implements `Agent.stop()` method that performs a clean async shutdown of an agent instance:
- Marks agent as shutting down and transitions status to OFFLINE
- Stops heartbeat background worker
- Notifies AgentField control plane of graceful shutdown (best effort)
- Cleans up async execution resources, memory event clients, and connection managers
- Idempotent: repeated calls have no additional effect after the first
 
Closes: #355

## Testing

- [x] `./scripts/test-all.sh`

## Checklist

- [x] I updated `CHANGELOG.md` (or this change does not warrant a changelog entry).

### DIFF
--- a/sdk/python/agentfield/agent.py
+++ b/sdk/python/agentfield/agent.py
@@ -3457,6 +3457,103 @@ class Agent(FastAPI,_AgentSchemaMixin,_AgentDiscoveryMixin,_AgentServerlessMixin
         # Also clear from thread-local storage
         clear_current_agent()
 
+    async def stop(self) -> None:
+        """
+        Programmatically stop the agent and clean up resources.
+
+        This method performs a graceful shutdown by:
+        1. Marking the agent as shutting down and its status as OFFLINE.
+        2. Stopping the heartbeat background worker.
+        3. Notifying the AgentField control plane that the agent is shutting down.
+        4. Cleaning up resources and event subscriptions.
+
+        The method is idempotent; calling it multiple times has no additional effect.
+
+        Example:
+            ```python
+            app = Agent("my_agent")
+            # ... start agent in a background task or loop ...
+
+            # Later, shut down cleanly
+            await app.stop()
+            ```
+        """
+        if getattr(self, "_shutdown_requested", False):
+            # Already shutting down or stopped
+            return
+
+        self._shutdown_requested = True
+
+        from agentfield.types import AgentStatus
+
+        self._current_status = AgentStatus.OFFLINE
+
+        if hasattr(self, "agentfield_handler") and self.agentfield_handler:
+            try:
+                self.agentfield_handler.stop_heartbeat()
+            except Exception as e:
+                if self.dev_mode:
+                    from agentfield.logger import log_error
+
+                    log_error(f"Heartbeat stop error during stop(): {e}")
+
+        try:
+            if (
+                getattr(self, "agentfield_connected", False)
+                and hasattr(self, "client")
+                and self.client
+            ):
+                success = await self.client.notify_graceful_shutdown(self.node_id)
+                if self.dev_mode:
+                    from agentfield.logger import log_info
+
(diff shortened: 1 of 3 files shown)


sdk/python/agentfield/agent.py
```diff
@@ -4070,6 +4070,105 @@ def _clear_current(self) -> None:
...
+
+        try:
+            from agentfield.agent_registry import clear_current_agent
+
+            clear_current_agent()
+        except Exception as e:
+            if self.dev_mode:
+                from agentfield.logger import log_error
+
+                log_error(f"Registry clear error during stop(): {e}")
```
AbirAbbas, line 3550: This calls `clear_current_agent()` but doesn't clear `Agent._current_agent` (the class-level attr). The existing `_clear_current()` method right above does both — it deletes the class attr *and* calls `clear_current_agent()`.

After `stop()`, `Agent.get_current()` will still return the stopped agent. Should probably just call `self._clear_current()` here instead.

AbirAbbas (commented):
Nice addition — the try/except isolation around each cleanup step is the right call, and the idempotency guard works. A couple things worth addressing:

sdk/python/agentfield/agent.py
shuaixr: I changed it to use self._clear_current() and added a regression assertion covering Agent.get_current() is None after stop(). sorry for the slow reply, ive been a bit busy lately

AbirAbbas: all good taking a look!

AbirAbbas (approved):

PR #462, opened 2026-04-17
Feature/typescript didclient vc methods

### LINKED ISSUE #92: [TypeScript SDK] Add verifyCredential() and getWorkflowVcChain() to DID client
Python's VC generator has additional methods for verification and workflow audit trails. Add these to the TypeScript DID client.

## What we need

```typescript
// Verify a credential signature
const isValid = await did.verifyCredential(vcDocument, signature);

// Get complete audit trail for a workflow
const chain = await did.getWorkflowVcChain(workflowId);

// Aggregate multiple execution VCs into a workflow VC
const workflowVc = await did.createWorkflowVc(executionVcs);
```

## Reference implementation

- `sdk/python/agentfield/vc_generator.py` lines 163, 197, 226

## Files to modify

- `sdk/typescript/src/did/DidClient.ts`

---

## Documentation (internal team)

This adds new DID/VC methods. Update the SDK documentation to include:
- New methods: verifyCredential, getWorkflowVcChain, createWorkflowVc
- Method signatures and return types
- Examples for verification and audit trail retrieval
- Security considerations

Luffy2208:
# Summary

Adds verifyCredential, getWorkflowVcChain, and createWorkflowVc to the TypeScript DidClient, aligned with the Python VCGenerator API and behavior.

Also adds unit tests for the new DID client methods and extends integration (E2E) tests to cover the full VC flow through the SDK (generation, aggregation, retrieval, and verification).

## Testing

- [ ] `./scripts/test-all.sh`
- [x] Additional verification:

cd sdk/typescript && npm test -- tests/did_client_methods.test.ts tests/integration.e2e.test.ts  
cd sdk/typescript && npm run lint  

Unit tests (did_client_methods.test.ts):
- verifies credentials with the raw verifier payload
- verifies credential without detached signature
- gets the raw workflow VC chain response
- creates a workflow VC from workflow metadata and execution VC ids

Integration tests (integration.e2e.test.ts):
- registers DID identities and issues workflow VCs through ctx.did during live executions

## Checklist

- [ ] I updated CHANGELOG.md (or this change does not warrant a changelog entry).

N/A

## Related issues

Fixes #92


sdk/typescript/src/did/DidClient.ts
```diff
@@ -218,6 +218,68 @@ export class DidClient {
...
 
+
+  async verifyCredential(vcDocument: any, signature?: string): Promise<any> {
+    const payload: any = {
+      vc_document: vcDocument
+    };
+
+    // Only include signature if explicitly provided
+    if (signature !== undefined) {
+      payload.signature = signature;
```
AbirAbbas: The optional `signature` param is forwarded as `payload.signature`, but the server's `VCVerificationRequest` (`control-plane/pkg/types/did_types.go`) only binds `vc_document` — the field is silently dropped. The server still checks the embedded `proof.proofValue` so this isn't a verification bypass, but the API surface is misleading: callers passing a detached signature will think it was checked.

Either drop the parameter (the Python equivalent omits it) or add the binding server-side.

Luffy2208: ok i'll drop the parameter

AbirAbbas: Dropped the `signature` parameter in 6f6029af to match the Python `VCGenerator` surface — the server only binds `vc_document` so the detached signature path was never exercised.

PR #496, opened 2026-04-21
Feature /agentfield slash command + prompt-to-production README

santoshkumarradha:
## Summary

- Feature the skill-driven flow as the primary way to try AgentField in the README (replaces the old quick-start-first structure).
- Ship a `/agentfield` slash command for Claude Code so users can trigger the skill explicitly, on top of the existing auto-match behavior.
- Wire the slash command into the existing multi-target skill installer so every `af skill install --target claude-code` drops it into `~/.claude/commands/` alongside the skill symlink, and `af skill uninstall` cleans it up.

### Why /agentfield

Skills auto-match by description, but a visible, discoverable slash command:
- removes the "does it know what to do?" doubt for first-time users,
- gives us a concrete CTA to feature in the README and docs,
- is a single atomic installer artifact rather than a README-only instruction.

### What changed

**Skillkit (Go):**
- `skills/agentfield-multi-reasoner-builder/commands/agentfield.md` — slash-command markdown with frontmatter (`description`, `argument-hint`) and a body that invokes the skill and enforces its hard gate.
- `control-plane/internal/skillkit/target_claude_code.go` — `Install()` now symlinks every `<current>/commands/*.md` into `~/.claude/commands/`; `Uninstall()` resolves `commands/` via `CanonicalRoot()` and removes the installed links.
- `control-plane/internal/skillkit/embed.go` — extended `//go:embed` to include `commands/*.md` so built `af` binaries ship the slash command alongside SKILL.md and references.
- `control-plane/internal/skillkit/skill_data/.../commands/agentfield.md` — embed-mirror copy (kept in sync by `scripts/sync-embedded-skills.sh`).
- `control-plane/internal/skillkit/skillkit_edge_test.go` — new table-driven edge case covering install, idempotent re-install, `.md` filter, and uninstall cleanup.

**README:**
- `/agentfield` callout inside the "Prompt to production" section.
- End-state copy updated to match what the skill actually produces — a scaffolded docker-compose stack with a local `curl` to test — instead of the old "run `af server` + `python main.py`" instruction.

### Verification

Installed end-to-end into an isolated `$HOME` and confirmed:
- `~/.claude/skills/agentfield-multi-reasoner-builder` → canonical `current/` (symlink)
- `~/.claude/commands/agentfield.md` → canonical `current/commands/agentfield.md` (symlink)
- `current/` → `0.3.0/` (versioned)
- Re-install with `--force` leaves the link pointing at the correct target.
- `af skill uninstall --target claude-code` removes both the skill symlink and the slash-command link with no residue.

## Test plan

- [x] `go build ./...` in `control-plane/`
- [x] `go test ./internal/skillkit/...` — 47 tests pass (new `claude install exposes slash commands and uninstall removes them` subtest included)
- [x] End-to-end install via `af skill install --target claude-code` in an isolated `$HOME`
- [x] Idempotent re-install with `--force`
- [x] `af skill uninstall --target claude-code` cleans up `~/.claude/commands/agentfield.md`
- [ ] Try `/agentfield <prompt>` in a fresh Claude Code session (requires the user to restart Claude Code after install — the session loads commands at startup)

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,36 +73,30 @@ app.run()
 
 > **What you just saw:** `app.ai()` calls an LLM and returns structured output. `app.pause()` suspends for [human approval](https://agentfield.ai/docs/build/execution/human-in-the-loop?utm_source=github-readme&utm_campaign=github-readme&utm_id=github-readme-human-in-the-loop). `app.call()` routes to other agents through the control plane. `app.run()` auto-exposes everything as REST. [Read the full docs →](https://agentfield.ai/docs/learn?utm_source=github-readme&utm_campaign=github-readme&utm_id=github-readme-read-full-docs)
 
-## Build with Claude Code (recommended)
+## Prompt to production (recommended)
 
-**Describe the backend. Your coding agent ships it. AgentField runs it.**
+**Describe the system in one line. Get a production-ready multi-agent backend.**
+
+Works in Claude Code, Codex, Gemini CLI, OpenCode, Aider, Windsurf, and Cursor.
 
 ```bash
-# Installs the af CLI AND drops the agentfield-multi-reasoner-builder skill
-# into every coding agent on your machine (Claude Code, Codex, Gemini,
-# OpenCode, Aider, Windsurf, Cursor).
 curl -fsSL https://agentfield.ai/install.sh | bash
 ```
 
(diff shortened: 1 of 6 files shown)


AbirAbbas (approved):
lgtm

Issue #404, opened 2026-04-11
[Go SDK] Add tests for harness providers (claim one file if you want)


santoshkumarradha:
## Test coverage: Go SDK `harness/` — factory, cli, result, provider, providers

The harness package has a factory, CLI wrapper, result struct, options helpers, and four provider implementations (claudecode, codex, gemini, opencode). Four of these providers follow the same shape. Bundled into one issue because the testing patterns share infrastructure.

### Scope
New test files:
- `sdk/go/harness/factory_test.go`
- `sdk/go/harness/cli_test.go`
- `sdk/go/harness/result_test.go`
- `sdk/go/harness/provider_test.go`
- `sdk/go/harness/claudecode_test.go`
- `sdk/go/harness/opencode_test.go`
- `sdk/go/harness/gemini_test.go`
- `sdk/go/harness/codex_test.go`

### Test cases

**factory.go / `BuildProvider`:**
- `BuildProvider("opencode", "path")` returns `*OpenCodeProvider`
- `BuildProvider("claude-code", "path")` returns `*ClaudeCodeProvider`
- `BuildProvider("unknown", "")` returns a helpful error

**cli.go:**
- `RunCLI` executes a basic command (`echo`) and captures stdout
- `RunCLI` correctly merges env vars (and unsets when value is empty)
- `RunCLI` respects context cancellation / timeout
- `StripANSI` removes complex escape sequences
- `isExecNotFound` identifies "executable file not found" errors

**result.go:**
- `(r *Result) Text()` returns the raw result string
- Various `FailureType` values produce consistent serialization

**provider.go / `Options`:**
- `maxRetries()` returns set value or default (3)
- `timeout()` returns set value or default (600)
- Each default/override path covered

**claudecode.go / opencode.go / gemini.go / codex.go `.Execute`:**
- Constructs correct CLI flags per provider (model, max-turns, etc.)
- Parses session metrics from provider output
- Handles binary-not-found gracefully
- Captures stdout as result; handles non-zero exit codes

### Fixture / mock boundary
- Real `exec.Command` with short shell scripts (`sh -c 'echo ...'`) for `RunCLI`
- Mock `claude` / `codex` / `gemini` / `opencode` binaries via tiny shell scripts in `t.TempDir()`
- Table-driven subtests

### Acceptance criteria
- [ ] All 8 test files added
- [ ] `go test ./sdk/go/harness/...` passes
- [ ] Coverage on `sdk/go/harness` ≥80%

[...]

PR #505, opened 2026-04-28
fix(harness): limit opencode concurrency with global semaphore (#438)

### LINKED ISSUE #438: [Go SDK] OpenCode harness has no concurrency guard
## Summary

`OpenCodeProvider.Execute` in the Go SDK harness spawns a new `opencode` subprocess on every call with no concurrency cap, allowing bursty callers to launch unbounded parallel processes and trigger rate-limit flapping.

## Context

`harness/opencode.go:29` calls `RunCLI` immediately for every `Execute` invocation. A fan-out reasoner or a parallel workflow with 20 branches will launch 20 concurrent `opencode` processes, each establishing its own LLM API session. This saturates the backend's rate limits, causes 429 cascades, and consumes system resources (file descriptors, process slots) proportionally to fan-out width. The Python SDK's OpenCode provider uses a class-level semaphore to cap concurrency — the Go SDK has no equivalent.

### In Scope

- Add a `MaxConcurrent int` field on `OpenCodeProvider` with a default value (e.g. 4, configurable via `OpenCodeConfig`).
- Use a buffered channel or `golang.org/x/sync/semaphore` to gate entry into `RunCLI`.
- If `MaxConcurrent` is 0 or negative, use the default (do not disable limiting entirely without an explicit `NoLimit` sentinel).
- Callers blocked on the semaphore should respect context cancellation.

### Out of Scope

- Adding a global semaphore across multiple `OpenCodeProvider` instances — per-instance is sufficient.
- Changing the `opencode` CLI invocation or argument handling.
- Adding retry logic for 429 responses — that is a separate concern.

## Files

- `sdk/go/harness/opencode.go:29` — add semaphore acquisition before `RunCLI`, release in defer
- `sdk/go/harness/opencode.go` — add `MaxConcurrent int` to `OpenCodeProvider` struct and `OpenCodeConfig`; initialize semaphore in constructor
- `sdk/go/harness/opencode_test.go` — test: with `MaxConcurrent=2`, a burst of 5 concurrent `Execute` calls does not launch more than 2 subprocesses simultaneously; context cancellation unblocks a waiting caller

## Acceptance Criteria

- [ ] At most `MaxConcurrent` `opencode` subprocesses run simultaneously across concurrent `Execute` calls on the same provider instance
- [ ] A caller blocked waiting for a slot is unblocked immediately when the context is cancelled (returns `context.Canceled`)
- [ ] Default behavior (`MaxConcurrent` not set) limits to 4 concurrent executions

## Notes for Contributors

**Severity**: MEDIUM

`golang.org/x/sync/semaphore` is cleaner than a buffered channel for this use case because it supports weighted acquisition (useful if future work adds priority levels). Check if `golang.org/x/sync` is already a dependency in `sdk/go/go.mod`; if not, a buffered channel semaphore is acceptable to avoid adding a new dependency.

dinzz005:
# Summary

Adds a concurrency limit to OpenCodeProvider.Execute to prevent unbounded subprocess spawning under parallel workloads.

## Changes
- Introduced a package-level semaphore (buffered channel) to cap concurrent executions
- Lazy initialization via sync.Once
- Configurable using OPENCODE_MAX_CONCURRENT env var (default: 4)
- Semaphore acquisition happens before RunCLI and release in defer
- Waiting callers respect context cancellation

## Motivation
Previously, each Execute() call spawned a subprocess without any limit, causing:
- rate-limit issues (429)
- resource exhaustion (FDs, processes)
- unstable behavior under fan-out workloads

This change introduces backpressure and stabilizes execution.

## Notes
- Matches Python SDK behavior (process-level concurrency control)
- No changes to CLI invocation or retry logic

## Testing

- [X] `./scripts/test-all.sh`
- [X] Additional verification (please describe):

- Added concurrency test ensuring at most N executions run simultaneously
- Added context cancellation test to ensure blocked callers exit immediately

## Checklist

- [] I updated documentation where applicable.
- [x ] I added or updated tests (or none were needed).
- [x] I updated `CHANGELOG.md` (or this change does not warrant a changelog entry).


dinzz005:
If there are any issues or improvements needed, please let me know—I’ll fix them.

sdk/go/harness/opencode.go
```diff
@@ -55,6 +78,15 @@ func (p *OpenCodeProvider) Execute(ctx context.Context, prompt string, options O
 		env[k] = v
 	}
 
+    sem := getSemaphore()
```
AbirAbbas: Tiny nit: this line is indented with spaces instead of a tab, and there's an extra blank line a few lines below — `gofmt -l` flags the file. SDK Go CI doesn't run gofmt today so it won't block, but worth a quick `gofmt -w` while you're here.

sdk/go/harness/opencode.go
```diff
@@ -4,16 +4,39 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strconv"
 	"strings"
+	"sync"
 	"time"
 )
 
+var runCLI = RunCLI
```
AbirAbbas: Heads-up: introducing a package-level mutable `runCLI` purely so tests can swap it works, but it's a footgun — if anyone later adds a `t.Parallel()` test in this package that also calls `Execute`, the swap will race. The two new tests are sequential so it's safe today. If you want to avoid the global, an alternative is making `runCLI` a field on `OpenCodeProvider` (or passing it via an option). Happy to defer.

sdk/go/harness/opencode_test.go
```diff
@@ -0,0 +1,101 @@
...
+		<-block // never releases until we close it
+		return &CLIResult{}, nil
+	}
+
+	p := NewOpenCodeProvider("", "")
+
+	// occupy the slot
+	go func() {
+		_, _ = p.Execute(context.Background(), "test", Options{})
+	}()
```
AbirAbbas, line 84: This goroutine blocks on `<-block` and `block` is never closed, so the goroutine outlives the test. Doesn't affect correctness, but a `defer close(block)` at the top of the test would keep things tidy.

dinzz005: thank you for this. i will fix it and push it as soon as possible.

dinzz005:
if you have a time review and tell me if there is any changes to be made.

dinzz005:
hey guys if you check this pr and able to merge it then i can work on this issue (#404).

AbirAbbas:
> hey guys if you check this pr and able to merge it then i can work on this issue (#404).

taking a look!

AbirAbbas (approved):

PR #536, opened 2026-05-06
feat(runs): bottom-up cancel-tree endpoint and UI guard fix

AbirAbbas:
## Summary
- Add `POST /workflows/:run_id/cancel-tree` that walks the run DAG and marks every non-terminal execution cancelled bottom-up. Reuses the existing per-execution cancel mechanics (status transition, sibling-table sync, event publishing) via a local helper to keep blast radius small.
- Relax UI guards in `RunDetailPage` and `RunLifecycleMenu` so the Cancel button shows whenever aggregate `workflow_status` is non-terminal — previously gated on the root execution's status, which hid Cancel for runs with a terminated root and zombied children (fan-out reasoners that finished before their dispatched siblings).
- Wire the run-detail header Cancel, the per-row kebab Cancel, and the bulk-cancel toolbar to the new `useCancelWorkflowTree` mutation. Pause/Resume are unchanged — they remain scoped to the root execution since that's the row the user controls.

## Why
A run with depth-0 root `status='succeeded'` but children still flagged `running` could not be cancelled — the Cancel button was hidden, and the per-execution cancel endpoint refused on terminal root. `cancel-tree` walks the whole run, so the user has an escape hatch even when the root has terminated.

## Changes
- **Backend** (`internal/handlers/execute_cancel_tree.go`):
  - `CancelWorkflowTreeHandler` loads all executions for the run via `QueryExecutionRecords`, computes depth from `parent_execution_id` chains, sorts deepest-first, and applies the per-execution cancel transition to each non-terminal execution.
  - Returns a per-node summary (`{cancelled_count, skipped_count, error_count, nodes: [...]}`) so the UI can render an accurate notification.
  - Mounted at `/api/v1/workflows/:workflowId/cancel-tree` and `/api/ui/v1/workflows/:workflowId/cancel-tree`.
- **Frontend**:
  - New `cancelWorkflowTree()` service method and `useCancelWorkflowTree()` mutation hook.
  - `RunDetailPage`: Cancel button gates on `!isTerminalStatus(dag.workflow_status)` instead of root status; calls cancel-tree and displays cancelled count in the success notification.
  - `RunLifecycleMenu`: same guard change for the per-row kebab.
  - `RunsPage`: per-row `handleCancelRun` and bulk cancel both switched to cancel-tree per run; bulk inclusion gate switched from `root_execution_status` to aggregate `r.status`.

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/handlers/...` (all pass)
- [x] `tsc --noEmit` (clean)
- [x] `vitest run` (124 files / 631 tests pass)
- [ ] Manual: open a zombied run on the deployed control-plane, confirm Cancel button appears and cancel-tree flips remaining non-terminal executions

## Backwards compatibility
Purely additive on the backend. Per-execution `POST /executions/:execution_id/cancel` is unchanged. Frontend Cancel surfaces all switch to cancel-tree, but the new endpoint subsumes the old behavior (a run with one non-terminal execution gets exactly that one cancelled).

### DIFF
--- a/control-plane/internal/handlers/execute_cancel_tree.go
+++ b/control-plane/internal/handlers/execute_cancel_tree.go
@@ -1,0 +1,315 @@
+package handlers
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/Agent-Field/agentfield/control-plane/internal/events"
+	"github.com/Agent-Field/agentfield/control-plane/internal/logger"
+	"github.com/Agent-Field/agentfield/control-plane/internal/storage"
+	"github.com/Agent-Field/agentfield/control-plane/pkg/types"
+
+	"github.com/gin-gonic/gin"
+)
+
+// cancel-tree exists because per-execution cancel cannot reach a run whose
+// root execution has already terminated but whose children are still flagged
+// non-terminal (zombied workers, fan-out reasoners that finished before
+// their dispatched siblings, etc.). The user-facing model is "cancel the
+// run" — this endpoint walks the DAG bottom-up and cancels every
+// non-terminal execution under the run, leaving terminal ones alone.
+
+type cancelTreeRequest struct {
+	Reason string `json:"reason"`
+}
+
+type cancelTreeNodeResult struct {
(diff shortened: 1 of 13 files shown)


AbirAbbas:

PR #538, opened 2026-05-06
feat(workers): cancel-signal transport via HTTP callback

AbirAbbas:
## Summary
Bridge between the in-process ExecutionEventBus (where the per-execution cancel handler and PR #536's cancel-tree handler already publish ExecutionCancelledEvent) and remote SDK workers, so cancellation can propagate into in-flight reasoner code rather than only updating storage state.

## Transport
- New service \`services.CancelDispatcher\` subscribes to \`events.GlobalExecutionEventBus\` on startup.
- On \`ExecutionCancelledEvent\`, looks up the agent by \`agent_node_id\`, resolves \`BaseURL\`, and POSTs:
  \`POST {BaseURL}/_internal/executions/:exec_id/cancel\`
  Body: \`{execution_id, workflow_id, reason, emitted_at}\`
  Headers: \`X-Execution-ID\`, \`X-Workflow-ID\`, \`X-AgentField-Source: cancel-dispatcher\`
- Reuses the existing dispatch HTTP channel — no second long-lived connection per worker, no auth changes (network trust matches the dispatch path).

## Behaviour
- **Best-effort:** 5s timeout, no retries. The control plane's storage record of \`cancelled\` is authoritative; the callback only exists so workers can short-circuit.
- **Backwards compatible:** workers running older SDKs without the cancel endpoint return 404, which is logged at debug and treated as a no-op.
- **Lifecycle:** Started in \`server.Start()\` after \`cleanupService\`, stopped from \`server.Stop()\`. Subscribes/unsubscribes cleanly on the bus.

## Why callback over SSE
Workers already expose HTTP servers that the control plane dispatches into. Reusing that channel is cheaper than maintaining a long-lived control-plane→worker SSE connection per registered agent. SSE would also need reconnection/resubscription logic on both sides; a one-shot POST is easier to reason about and easier to test.

## Test plan
- [x] \`go vet ./...\`
- [x] \`go test ./internal/services/ -run TestCancelDispatcher\` (4/4 pass)
- [x] \`go test ./internal/services/...\` (full suite green)
- [ ] Manual: deploy + verify cancel from UI triggers callback to a registered worker

## Follow-ups (separate PRs)
- PR 3 — Go SDK: register \`/_internal/executions/:id/cancel\` handler, derive per-execution \`context.Context\`, cancel on receipt
- PR 4 — Python SDK: same handler, wrap reasoner invocation in \`asyncio.Task\` and \`task.cancel()\` on receipt
- PR 5 — TypeScript SDK: same handler, per-execution \`AbortController\`, \`abort()\` on receipt

### DIFF
--- a/control-plane/internal/server/server.go
+++ b/control-plane/internal/server/server.go
@@ -457,7 +457,9 @@ func NewAgentFieldServer(cfg *config.Config) (*AgentFieldServer, error) {
 	triggerHandlers := handlers.NewTriggerHandlers(storageProvider, triggerDispatcher, sourceManager)
 	handlers.SetTriggerSourceManager(sourceManager)
 
-	cancelDispatcher := services.NewCancelDispatcher(storageProvider, services.CancelDispatcherConfig{})
+	cancelDispatcher := services.NewCancelDispatcher(storageProvider, services.CancelDispatcherConfig{
+		InternalToken: cfg.Features.DID.Authorization.InternalToken,
+	})
 
 	return &AgentFieldServer{
 		storage:                storageProvider,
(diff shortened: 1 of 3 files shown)


AbirAbbas:

PR #540, opened 2026-05-06
feat(sdk-python): cooperative cancel via /_internal/executions/:id/cancel

AbirAbbas:
## Summary
Add cooperative cancellation to the Python SDK. Workers now register an \`asyncio.Task\` per in-flight reasoner invocation and expose \`POST /_internal/executions/:id/cancel\`. The control plane's cancel dispatcher (#538) calls into the route, which calls \`task.cancel()\` — raising \`asyncio.CancelledError\` into the reasoner coroutine. Async I/O libs (httpx, the official anthropic and openai SDKs, asyncpg, etc.) honor task cancellation natively, so most reasoners get cooperative cancellation without source changes.

Depends on #538 (cancel-signal transport). Functions correctly without it — the route is just dormant.

## Changes
- \`agentfield/cancel.py\` — new module:
  - \`register_execution_task\` / \`deregister_execution\` / \`cancel_execution\`
  - \`install_cancel_route(agent)\` — registers the FastAPI POST route
  - \`is_execution_cancelled(agent, execution_id)\` — sync helper for CPU-loop polling
- \`Agent.__init__\`:
  - \`self._cancel_tasks: Dict[str, asyncio.Task]\` and \`self._cancel_lock\`
  - Calls \`install_cancel_route(self)\` so the path is always available
- \`Agent\` reasoner endpoint (sync path): wraps \`run_reasoner()\` in \`asyncio.create_task\`, registers, awaits, catches \`CancelledError\` → 499 \`{status: cancelled, ...}\` response
- \`_execute_async_with_callback\`: catches \`CancelledError\`, emits a \`status: cancelled\` payload to the control-plane callback, deregisters in \`finally\`

## Backwards compatibility
Purely additive. Reasoners that swallow \`CancelledError\` (\`except Exception\` in 3.8+ does *not* — \`CancelledError\` is a \`BaseException\`) or use blocking sync code continue to run; their output is discarded by the control plane (existing per-execution cancel behaviour). The \`/_internal/\` route is harmless when the dispatcher isn't running in front of the worker.

## Test plan
- [x] \`pytest tests/test_cancel.py\` (9/9 pass)
- [x] Targeted agent suites (core, lifecycle, serverless, decorators, registry, resilience, helpers): 74/74 pass
- [x] Workflow + parent-child + approval + reasoner-path + handler suites: 107/107 pass when isolated from a known pre-existing flaky-ordering issue in \`test_agent_workflow.py\` (unrelated to this PR — confirmed by running \`test_workflow_parent_child.py\` alone; passes 9/9)
- [ ] Manual: register a Python-SDK agent on the deployed control-plane, trigger a long-running reasoner (e.g. \`asyncio.sleep(60)\`), hit Cancel, observe \`CancelledError\` raised and \`status: cancelled\` reported

### DIFF
--- a/sdk/python/agentfield/agent.py
+++ b/sdk/python/agentfield/agent.py
@@ -653,6 +653,21 @@ class Agent(FastAPI):
         # prevent GC of fire-and-forget async execution tasks
         self._background_tasks: set[asyncio.Task] = set()
 
+        # Cooperative cancel registry. The control plane's cancel dispatcher
+        # POSTs /_internal/executions/{id}/cancel to signal that the user's
+        # reasoner code should stop. We track the active asyncio.Task per
+        # execution_id so the cancel handler can call task.cancel() — that
+        # raises CancelledError into the user's coroutine, which any
+        # well-behaved async I/O (httpx, anthropic, openai) honors.
+        self._cancel_tasks: Dict[str, asyncio.Task] = {}
+        self._cancel_lock = asyncio.Lock()
+
+        # Install the /_internal/executions/{execution_id}/cancel route
+        # before any user-defined reasoners are registered so the path is
+        # always available for the control-plane callback.
+        from .cancel import install_cancel_route
+        install_cancel_route(self)
+
         # Initialize async execution manager (will be lazily created when needed)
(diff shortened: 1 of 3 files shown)


AbirAbbas:

PR #541, opened 2026-05-06
feat(sdk-typescript): cooperative cancel via /_internal/executions/:id/cancel

AbirAbbas:
## Summary
Add cooperative cancellation to the TypeScript SDK. Workers now register an \`AbortController\` per in-flight reasoner / skill invocation and expose \`POST /_internal/executions/:id/cancel\`. The control plane's cancel dispatcher (#538) calls into the route, which calls \`.abort()\` — surfacing as \`signal.aborted === true\` and rejecting any pending \`fetch\` / Anthropic-SDK / OpenAI-SDK request bound to \`ctx.signal\`.

Depends on #538 (cancel-signal transport). Functions correctly without it — the route is just dormant.

## Changes
- \`src/agent/cancel.ts\` — new module:
  - \`CancelRegistry\` with \`register / cancel / size\`. \`register\` returns \`{ controller, release }\` so the call site cleans up in \`finally\` regardless of outcome.
  - \`installCancelRoute(app, registry, logger?)\` registers \`POST /_internal/executions/:executionId/cancel\` on the Express app.
- \`src/agent/Agent.ts\`:
  - private \`cancelRegistry = new CancelRegistry()\`
  - \`installCancelRoute(this.app, this.cancelRegistry, ...)\` in the constructor (always-on)
  - \`runReasoner\` / \`runSkill\`: register a controller around the handler invocation, pass \`signal: controller.signal\` into the context, release in \`finally\`
- \`src/context/ReasonerContext.ts\` / \`src/context/SkillContext.ts\`: new \`readonly signal: AbortSignal\` field. Optional in the constructor (\`signal?: AbortSignal\`), defaults to a never-aborted signal so \`getCurrent*Context\` helpers and existing tests continue to work unchanged.

## Reasoner-author contract
\`\`\`ts
agent.reasoner('do-thing', async (ctx) => {
  const resp = await fetch(url, { signal: ctx.signal });        // aborts mid-flight
  const reply = await anthropic.messages.create(req, { signal: ctx.signal });
  for (const item of items) {
    if (ctx.signal.aborted) throw new Error('cancelled');        // CPU loop polling
  }
});
\`\`\`

## Backwards compatibility
Purely additive. Handlers that ignore \`ctx.signal\` continue to run; their output is discarded by the control plane (existing per-execution cancel behaviour). The \`/_internal/\` route is harmless when the dispatcher isn't running in front of the worker.

## Test plan
- [x] \`tsc --noEmit\` (clean)
- [x] \`vitest run tests/cancel.test.ts\` (14/14 pass)
- [x] \`vitest run\` (64 files / 595 tests pass)
- [ ] Manual: register a TS-SDK agent on the deployed control-plane, trigger a long-running reasoner using \`ctx.signal\` with \`fetch\`, hit Cancel, observe the request aborts and the handler reports a clean error

### DIFF
--- a/sdk/typescript/src/agent/Agent.ts
+++ b/sdk/typescript/src/agent/Agent.ts
@@ -14,6 +14,7 @@ import type {
 } from '../types/agent.js';
 import { ReasonerRegistry } from './ReasonerRegistry.js';
 import { SkillRegistry } from './SkillRegistry.js';
+import { CancelRegistry, installCancelRoute } from './cancel.js';
 import { AgentRouter } from '../router/AgentRouter.js';
 import type { ReasonerHandler, ReasonerOptions } from '../types/reasoner.js';
 import type { SkillHandler, SkillOptions } from '../types/skill.js';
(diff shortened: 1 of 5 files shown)


AbirAbbas:

PR #564, opened 2026-05-10
fix(sdk): pause-aware overdue check in async polling task

AbirAbbas:
## Bug

github-buddy's `app.call("swe-planner.build", ...)` timed out at **exactly 21600s wallclock** on Railway run `run_1778268481826_8c9dd544` and again on `run_1778346573033_dafddc40` despite v0.1.81's pause-aware `wait_for_result`. The parent reasoner (`implement_from_issue`) caught the spurious `ExecutionTimeoutError`, returned a result dict carrying the error, and the control plane marked the parent as `succeeded` — while SWE-AF's `build` execution stayed alive in `waiting` for hours afterward, orphaned.

Symptom in the UI: parent step renders **green/succeeded** with `result.error = "ExecutionTimeoutError: Execution timed out after 21600.0 seconds"`, while a child step is still **yellow/waiting**. Both are technically what the data says, but the parent should never have given up — its 6h budget was supposed to ignore time the child spent paused on a hax-sdk human approval.

## Root cause

`AsyncExecutionManager` runs a background polling task (`_poll_active_executions`) that calls `execution.is_overdue` on every cycle:

```python
if execution.is_overdue:
    execution.timeout_execution()        # → status = TIMEOUT
    self.metrics.timeout_executions += 1
    self.metrics.active_executions -= 1
    continue
```

`is_overdue` is pure wallclock:

```python
@property
def is_overdue(self) -> bool:
    if self.timeout is None:
        return False
    return self.age > self.timeout       # age = time.time() - submit_time
```

After ``timeout`` seconds of wallclock — regardless of how much of that wallclock was spent in the awaited child's `waiting` state — it flips `execution.status` to TIMEOUT. The pause-aware `wait_for_result` loop then reads that TIMEOUT on its next iteration and raises:

```python
elif execution.status == ExecutionStatus.TIMEOUT:
    raise ExecutionTimeoutError(
        f"Execution timed out after {execution.timeout} seconds"
    )
```

That's exactly the error string we see. v0.1.81 made the wait loop's own elapsed-time check pause-aware, but missed the **second** pause-unaware timeout check that runs in parallel inside the polling task. The polling task wins the race at exactly the wallclock budget.

## Fix

Two-commit split for reviewability:

1. **`ExecutionState`** gets an optional `_pause_clock` field (private, `init=False`); `is_overdue` subtracts `pause_clock.total_paused()` from age before comparing against the timeout. With no clock attached, behaviour is identical to before.
2. **`wait_for_result`** attaches the caller-supplied `pause_clock` to the awaited `ExecutionState` for the duration of the wait, then restores the previous value in `finally` so future waits don't consume a stale parent clock.

The polling task and the wait loop now read paused-time from the same source.

## Validation Contract

Behaviours pinned by the new tests:

- A paused execution must NOT be flipped TIMEOUT by `_poll_active_executions` while wallclock > timeout but `paused-time > (wallclock - timeout)`.
  Pinned by [`test_poll_active_executions_respects_attached_pause_clock`](sdk/python/tests/test_async_execution.py).
- An execution with NO pause_clock attached must still time out at wallclock — same legacy behaviour.
  Pinned by `test_poll_active_executions_still_times_out_without_pause_clock`.
- The clock must be attached for the duration of the wait and detached on every exit (success / timeout / exception), so a future wait on the same execution does not consume a stale parent clock.
  Pinned by `test_wait_for_result_attaches_pause_clock_to_execution_state`.
- `is_overdue` with a clock claiming more paused time than the wallclock age returns False.
  Pinned by `test_is_overdue_subtracts_attached_pause_clock_from_age`.
- A clock that raises from `total_paused()` falls back to wallclock so executions still time out eventually.
  Pinned by `test_is_overdue_pause_clock_failure_falls_back_to_wallclock`.
- Without `_pause_clock` set, `is_overdue` is unchanged.
  Pinned by `test_is_overdue_no_pause_clock_matches_legacy_behavior`.

## Test plan

- [x] `pytest --no-cov` — **1488 passed, 4 skipped** (integration tests requiring external infra), 0 failures.
- [x] `ruff check .` — all checks passed.
- [x] 7 new tests cover the contract above and the failure-mode regression.

## Why this didn't fail in prior tests

v0.1.81's tests drove `wait_for_result` directly with a paused child, but they did not run the polling loop alongside it. The bug only surfaces when both run concurrently — which is the production configuration. Test `test_poll_active_executions_respects_attached_pause_clock` now drives the production data path: status updated by polling, timeout check by polling, wait loop reads the same `_executions[id].status`.

### DIFF
--- a/sdk/python/agentfield/async_execution_manager.py
+++ b/sdk/python/agentfield/async_execution_manager.py
@@ -563,6 +563,26 @@ class AsyncExecutionManager:
         # Polling is unconditional, so this works whether or not the SSE
         # event stream is enabled.
         child_pause_active = False
+        # Attach the pause-clock to the execution state so the polling task's
+        # ``is_overdue`` check is also pause-aware. Without this, the polling
+        # loop's wallclock-based overdue check pre-empts the pause-aware loop
+        # below: it marks the execution TIMEOUT at exactly ``timeout`` seconds
+        # of wallclock — even when most of that wallclock was spent in the
+        # child's ``waiting`` state — and we surface that as a spurious
+        # ExecutionTimeoutError. Snapshot the previous value so we restore it
+        # in the finally block (concurrent waiters on the same execution_id
+        # are unusual but supported by the manager API).
+        previous_pause_clock = None
+        attached_pause_clock = False
+        if pause_clock is not None:
+            async with self._execution_lock:
+                execution = self._executions.get(execution_id)
+                if execution is not None:
+                    previous_pause_clock = getattr(
+                        execution, "_pause_clock", None
+                    )
+                    execution._pause_clock = pause_clock
+                    attached_pause_clock = True
         try:
             # Wait for completion
             while _active_elapsed() < wait_timeout:
@@ -639,6 +659,13 @@ class AsyncExecutionManager:
                     pause_clock.end_pause()
                 except Exception:
                     pass
+            # Restore the previous ``_pause_clock`` so future polling passes
+            # don't keep subtracting paused time from a stale parent clock.
+            if attached_pause_clock:
+                async with self._execution_lock:
+                    execution = self._executions.get(execution_id)
+                    if execution is not None:
+                        execution._pause_clock = previous_pause_clock
 
     async def cancel_execution(
         self, execution_id: str, reason: Optional[str] = None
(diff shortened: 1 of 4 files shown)


santoshkumarradha:
Ah damn, good catch on this !

Issue #421, opened 2026-04-12
[Security] Execution notes read endpoint has no authorization (IDOR)


santoshkumarradha:
## Summary

`GetExecutionNotesHandler` returns notes for any execution ID without verifying that the requesting caller owns or is authorized to read that execution.

## Context

The read endpoint at `execution_notes.go:135` fetches execution notes by `:execution_id` path parameter and returns them with no ownership check. Any authenticated caller can enumerate notes for any execution in the system by iterating execution IDs, leaking potentially sensitive workflow data and reasoning traces across agent/tenant boundaries in multi-tenant deployments.

### In Scope

- Add a caller-identity authorization check in `GetExecutionNotesHandler` that validates the requesting agent/session has read access to the target execution.
- Return HTTP 403 (not 404) when an execution exists but the caller is unauthorized; returning 404 is also acceptable if information-hiding is preferred — be consistent with how other protected resources behave.

### Out of Scope

- Fixing the write endpoint — that is tracked separately (S2).
- Implementing shared-read access or delegation — basic ownership match is sufficient.
- Changing the notes storage schema.

## Files

- `control-plane/internal/handlers/execution_notes.go:135` — add authorization check in `GetExecutionNotesHandler`
- `control-plane/internal/handlers/execution_notes_test.go` — tests: owner read succeeds, cross-agent read returns 403/404, unauthenticated request returns 401

## Acceptance Criteria

- [ ] Caller authenticated as agent A cannot read notes for executions belonging to agent B
- [ ] Caller authenticated as agent A can read notes for their own executions
- [ ] Unauthenticated requests are rejected before reaching the notes fetch (existing auth middleware handles this — just verify it's applied)
- [ ] Tests pass (`go test ./control-plane/...`)
- [ ] Linting passes (`make lint`)

## Notes for Contributors

**Severity**: HIGH

The ownership-check pattern added for S2 (write endpoint) should be directly reusable here. Decide whether to return 403 or 404 for unauthorized reads and apply that choice consistently with the rest of the handler layer.

Luffy2208:
I can work on this if assigned.

Luffy2208:
I’m planning to handle this after #420 lands, since the read-side fix should reuse the ownership/caller-identity helpers being added there for the write endpoint.
Once #420 is merged, I’ll rebase from latest `main` and implement the read authorization on top of those shared helpers instead of duplicating the DID/caller resolution logic. That should keep the #421 patch smaller and easier to review.

PR #575, opened 2026-05-19
fix(control-plane): authorize execution note writes(#420)

### LINKED ISSUE #420: [Security] Execution notes write endpoint has no authorization (IDOR)
## Summary

`AddExecutionNoteHandler` performs no ownership or tenant check before writing notes to an arbitrary execution, allowing any authenticated caller to add notes to another agent's executions.

## Context

The handler reads `executionID` from context, the `X-Execution-ID` header, or a query parameter, then writes notes to that execution with no verification that the caller owns or has access to it. This is a classic IDOR (Insecure Direct Object Reference): any API key holder can mutate execution data belonging to a different agent or tenant, making audit trails untrustworthy and enabling data poisoning in multi-tenant deployments.

### In Scope

- Add an ownership/authorization check in `AddExecutionNoteHandler` that validates the authenticated agent's identity matches the execution's owning agent before permitting the write.
- Return a 403 with a clear error message when the check fails.
- Ensure the check works for both API-key and DID-authenticated callers.

### Out of Scope

- Changing the notes data model or storage schema.
- Fixing the read endpoint — that is tracked in a separate issue (S3).
- Adding roles or a full RBAC system — basic ownership check is sufficient here.

## Files

- `control-plane/internal/handlers/execution_notes.go:45` — add caller-vs-execution ownership assertion in `AddExecutionNoteHandler`
- `control-plane/internal/services/execution_service.go` (or storage layer) — expose a method to retrieve the owning agent ID for an execution, if not already available
- `control-plane/internal/handlers/execution_notes_test.go` — add tests: owner write succeeds, non-owner write returns 403

## Acceptance Criteria

- [ ] A caller authenticated as agent A cannot add notes to executions owned by agent B (returns HTTP 403)
- [ ] A caller authenticated as agent A can add notes to their own executions (returns 200/201)
- [ ] The check works for both API-key and DID-auth callers

## Notes for Contributors

**Severity**: HIGH

Check how other write handlers (e.g. execution stop/cancel) verify ownership — replicate that pattern. The authenticated agent identity should already be available in the Gin context from the auth middleware; retrieve it with the same context key used elsewhere.

Luffy2208:
# Summary

Fixes an IDOR in the execution notes write endpoint by enforcing execution ownership before appending a note.

File-specific changes:

- `control-plane/internal/handlers/execution_notes.go`
  - Resolves the caller agent identity before writing a note.
  - Compares the caller agent ID with the execution owner, `execution.AgentNodeID`.
  - Returns `403 execution_ownership_mismatch` when the caller does not own the execution.
  - Supports DID-authenticated callers by resolving the verified caller DID to an agent ID.

- `control-plane/internal/server/middleware/auth.go`
  - Stores API-key caller identity in Gin context using `CallerAgentIDKey`.
  - Uses `X-Caller-Agent-ID` first, with `X-Agent-Node-ID` as fallback.

- `control-plane/internal/handlers/execution_notes_test.go`
  - Adds coverage for owner write success.
  - Adds coverage for non-owner API-key write returning `403`.
  - Adds coverage for DID-authenticated owner and non-owner behavior.

- `control-plane/internal/server/middleware/auth_test.go`
  - Adds coverage that API-key auth populates caller identity in Gin context.

- `control-plane/internal/handlers/coverage_handlers_90_test.go`
  - Updates the existing successful note-write coverage test to include matching caller identity.

## Testing

- [ ] `./scripts/test-all.sh`
- [x] Additional verification:
  - `cd control-plane && go test ./internal/handlers ./internal/server/middleware`
  - `cd control-plane && go test ./internal/handlers ./internal/server/middleware -coverprofile=/tmp/issue-420.coverprofile`
  - `cd control-plane && golangci-lint run --new-from-rev=upstream/main ./internal/handlers ./internal/server/middleware`
  - Manual E2E curl verification:
    - Agent A adding a note to Agent A’s execution returns `200 OK`
    - Agent A adding a note to Agent B’s execution returns `403 Forbidden`
    - Agent B’s fresh execution remains with `notes: []` after the blocked write

Note: full repo lint currently reports pre-existing unrelated Go lint issues outside this PR’s changed files. Changed-line lint for the touched packages reports `0 issues`.

## Checklist

Not UI-related.

## Related issues

Fixes #420


Luffy2208:
I fixed the patch coverage failure by adding targeted tests in `control-plane/internal/handlers/execution_notes_test.go`.

What was added:

- Tests for missing execution owner returning `403`
- Tests for missing caller identity returning `403`
- Tests for DID resolver failure returning `500`
- Tests for unresolved DID fail-closed behavior returning `403`
- Tests for caller identity resolution from:
  - Gin context `CallerAgentIDKey`
  - `X-Caller-Agent-ID`
  - `X-Agent-Node-ID`
  - DID list fallback
  - DID lookup error fallback
- Direct coverage for `executionNoteAuthorizationError.Error()`

control-plane/internal/handlers/execution_notes.go
```diff
@@ -130,6 +165,71 @@ func AddExecutionNoteHandler(storageProvider ExecutionNoteStorage) gin.HandlerFu
...
+		return &executionNoteAuthorizationError{message: "caller agent identity is required to add notes to this execution"}
+	}
+	if callerAgentID != ownerAgentID {
+		return &executionNoteAuthorizationError{message: "this execution does not belong to the requesting agent"}
+	}
+
+	return nil
+}
+
+func executionNoteCallerAgentID(ctx context.Context, c *gin.Context, storageProvider ExecutionNoteStorage) (string, error) {
```
santoshkumarradha, line 207: 🔴 **[CRITICAL] Raw-header fallback becomes sole identity source under default config**

`executionNoteCallerAgentID` has 3 tiers: (1) verified DID, (2) `CallerAgentIDKey` context value, (3) raw `X-Caller-Agent-ID`/`X-Agent-Node-ID` headers. Tiers 1 & 2 are config-gated. Under defaults (`APIKey=""`, `did_auth_enabled=false`), both are skipped — tier 3 accepts attacker-controlled headers with zero validation, flowing directly to `ensureExecutionNoteOwnership`.

Evidence:
- `routes_middleware.go:77` — DID middleware not installed when disabled
- `auth.go:26-28` — APIKeyAuth no-ops when `APIKey==""`, never sets context key
- `execution_notes.go:196-201` — raw header read with no validation
- When APIKeyAuth *does* run, it reads the same headers (`auth.go:118-124`), so the fallback is either dead code or active bypass — never legitimate

**Fix**: Delete the raw-header fallback. Add a startup assertion in `routes_middleware.go` that refuses to register write routes when both auth methods are disabled.

---
*`Compound Analysis` · confidence 95%*

control-plane/internal/handlers/execution_notes.go
```diff
@@ -130,6 +165,71 @@ func AddExecutionNoteHandler(storageProvider ExecutionNoteStorage) gin.HandlerFu
...
+			return strings.TrimSpace(info.AgentNodeID), nil
+		}
+	}
+
+	return "", nil
+}
+
 // GetExecutionNotesHandler handles GET /api/v1/executions/:execution_id/notes
 // Retrieves notes for a specific execution with optional tag filtering
 func GetExecutionNotesHandler(storageProvider ExecutionNoteStorage) gin.HandlerFunc {
```
santoshkumarradha, line 268: 🟠 **[IMPORTANT] Read path leaks execution notes — no ownership check**

PR fixes IDOR on write path but `GetExecutionNotesHandler` (line 235) remains open: any API-key-authenticated caller can read any execution's notes by ID. `storageProvider.GetExecutionRecord()` is called at line 256 with no caller identity resolution or comparison against `execution.AgentNodeID` — same IDOR pattern just fixed on write.

Notes carry workflow state — phase progress, intermediate results, confidence reasoning. See `examples/python_agent_nodes/agentic_rag/main.py:912,914,963,1097`. The UI details endpoint `GetExecutionDetailsGlobalHandler` (ui/executions.go:558, lines 784-785) also exposes `NotesCount`/`LatestNote` without ownership enforcement.

No code comment, test, or PR description text explains whether the open read is intentional. Likely oversight given the write-side fix.

**Fix**: Mirror write path — resolve caller via `executionNoteCallerAgentID`, then `ensureExecutionNoteOwnership`, return 403 on mismatch. If intentional, add a code comment + a test confirming non-owner reads are the intended contract.

---
*`Authorization Asymmetry` · confidence 90%*

control-plane/internal/handlers/execution_notes.go
santoshkumarradha, line 207: 🟠 **[IMPORTANT] DID→agent-ID resolution is duplicated in two places with subtly different fallback chains**

`executionNoteCallerAgentID` (execution_notes.go:185-186) calls a local `resolveExecutionNoteAgentIDByDID` (line 206) instead of reusing the central `DIDWebService.ResolveAgentIDByDID` (did_web_service.go:53-64) that `MemoryPermissionMiddleware` uses via `DIDResolverInterface` (permission.go:22-27).

Both paths happen to query `GetDIDDocument` first today, but the fallbacks diverge:
- Handler path falls back to `ListAgentDIDs()` over the `agent_dids` SQL table.
- `DIDWebService` path falls back to `DIDService.ResolveAgentIDByDID` (did_service.go:500) which scans the in-memory `registry.AgentNodes`.

For did:key callers (where `GetDIDDocument` returns not-found because did:key isn't stored as a did:web document), the two paths can resolve to different identities during agent registration/unregistration windows when SQL and the in-memory registry are out of sync. The handler also returns a `nil` agent ID silently if no `executionNoteAgentDIDLister` is implemented (line 215), while the central path returns the in-memory result.

**Fix**: Have `executionNoteCallerAgentID` delegate to the same `DIDResolverInterface` used elsewhere, or inject a single resolver into the handler. Eliminating the duplicated resolution function removes the fork.

---
*`Duplicated DID resolution path` · confidence 75%*

control-plane/internal/handlers/execution_notes.go
```diff
@@ -130,6 +165,71 @@ func AddExecutionNoteHandler(storageProvider ExecutionNoteStorage) gin.HandlerFu
...
+		return agentID, nil
+	}
+	if agentID := strings.TrimSpace(c.GetHeader("X-Agent-Node-ID")); agentID != "" {
+		return agentID, nil
+	}
+
+	return "", nil
+}
+
+func resolveExecutionNoteAgentIDByDID(ctx context.Context, storageProvider ExecutionNoteStorage, callerDID string) (string, error) {
```
santoshkumarradha, line 236: 🟠 **[IMPORTANT] DID resolution reads differently-named columns from independent tables**

`resolveExecutionNoteAgentIDByDID` (line 206) has two paths returning what should be the same identifier:

- Path 1 (line 209): `record.AgentID` from `did_documents.agent_id` (migration 019)
- Path 2 (line 226): `info.AgentNodeID` from `agent_dids.agent_node_id` (migration 002)

Both compared against `execution.AgentNodeID`. They converge today only because `nodes_register.go:1027,1044` populates both from `newNode.ID` — pure code coincidence. No FK, unique constraint, or comment ties them. A future migration touching one but not the other silently breaks resolution, causing false 403s.

The naming divergence (`AgentID` vs `AgentNodeID`) obscures the invariant.

**Fix**: (1) Document the equivalence on `DIDDocumentRecord.AgentID`. (2) Rename to `AgentNodeID` for consistency. (3) Add invariant test that creates entries in both tables and verifies both paths return the same result.

---
*`Field-name divergence` · confidence 85%*

control-plane/internal/handlers/execution_notes.go
```diff
@@ -130,6 +165,71 @@ func AddExecutionNoteHandler(storageProvider ExecutionNoteStorage) gin.HandlerFu
...
+
+	return nil
+}
+
+func executionNoteCallerAgentID(ctx context.Context, c *gin.Context, storageProvider ExecutionNoteStorage) (string, error) {
+	if callerDID := strings.TrimSpace(middleware.GetVerifiedCallerDID(c)); callerDID != "" {
+		return resolveExecutionNoteAgentIDByDID(ctx, storageProvider, callerDID)
+	}
+
+	if callerID, exists := c.Get(string(middleware.CallerAgentIDKey)); exists {
```
santoshkumarradha, line 212: 🟠 **[IMPORTANT] Type-unsafe `CallerAgentIDKey` enables silent fallback to attacker headers**

`c.Set(string, any)` accepts any value type — only convention enforces `string`. The handler at line 190 does `id, ok := callerID.(string)`; on a non-string value, `ok=false`, `id=""`, the inner block is skipped, and control falls through to raw `X-Caller-Agent-ID` / `X-Agent-Node-ID` headers (lines 196-201) with no log or error.

Attack path: an authenticated request hits a future middleware bug that writes a non-string to the key → identity silently discarded → handler reads attacker-controlled headers → ownership check passes if attacker knows the execution ID.

Proof: `context_helpers_additional_test.go:19-20` shows `c.Set(string(CallerAgentIDKey), 42)` compiles cleanly and `GetCallerAgentID` returns `""` — the language permits the bug.

**Fix**: (1) Typed setter `SetCallerAgentID(c, agentID string)` + `forbidigo` lint rule banning raw `c.Set` on this key. (2) When `exists==true && ok==false`, log WARN and return error — never fall to raw headers. (3) Remove the raw-header fallback entirely (see related comment).

---
*`Compound Analysis` · confidence 70%*

control-plane/internal/handlers/execution_notes.go
```diff
@@ -130,6 +165,71 @@ func AddExecutionNoteHandler(storageProvider ExecutionNoteStorage) gin.HandlerFu
...
+	if agentID := strings.TrimSpace(c.GetHeader("X-Agent-Node-ID")); agentID != "" {
+		return agentID, nil
+	}
+
+	return "", nil
+}
+
+func resolveExecutionNoteAgentIDByDID(ctx context.Context, storageProvider ExecutionNoteStorage, callerDID string) (string, error) {
+	if lookup, ok := storageProvider.(executionNoteDIDDocumentLookup); ok {
+		if record, err := lookup.GetDIDDocument(ctx, callerDID); err == nil && record != nil {
```
santoshkumarradha: 🟡 **[MEDIUM] DID resolution lookup doesn't filter revoked records**

`resolveExecutionNoteAgentIDByDID` (line 208) accepts any non-error `GetDIDDocument` result. `LocalStorage.GetDIDDocument` (local.go:8305-8334) returns records with `RevokedAt` populated and `err==nil` — no `revoked_at IS NULL` filter, unlike `GetDIDDocumentByAgentID` (local.go:8345). `ListAgentDIDs` (local.go:6864-6917) similarly returns rows regardless of `agent_dids.status` (which can be `active|inactive|revoked` per migration 002).

**Defense-in-depth concern, not a direct bypass for did:web**: `DIDAuthMiddleware` → `VerifyDIDOwnership` → `ResolveDID` (did_web_service.go:128-149) already returns `nil` document for revoked did:web, so the auth layer rejects with 401 before reaching this handler. The handler's missing revocation check is a secondary gap — it matters if:
- The caller is a did:key whose entry in `agent_dids` was marked `revoked` (auth still passes because did:key is self-verifying)
- Revocation happens between auth and handler execution (narrow race)
- This resolver is later reused in a code path that doesn't run `DIDAuthMiddleware` first

**Fix**: After the `GetDIDDocument` call, `if record.IsRevoked() { /* fall through or error */ }`. For `ListAgentDIDs`, skip entries where `info.Status == AgentDIDStatusRevoked`. Or push the filters into the queries as `GetDIDDocumentByAgentID` does.

---
*`Defense-in-depth: revocation filter` · confidence 80%*

control-plane/internal/handlers/execution_notes.go
```diff
@@ -130,6 +165,71 @@ func AddExecutionNoteHandler(storageProvider ExecutionNoteStorage) gin.HandlerFu
...
+	}
+	if agentID := strings.TrimSpace(c.GetHeader("X-Agent-Node-ID")); agentID != "" {
+		return agentID, nil
+	}
+
+	return "", nil
+}
+
+func resolveExecutionNoteAgentIDByDID(ctx context.Context, storageProvider ExecutionNoteStorage, callerDID string) (string, error) {
+	if lookup, ok := storageProvider.(executionNoteDIDDocumentLookup); ok {
```
santoshkumarradha, line 237: **[MEDIUM] `GetDIDDocument` errors swallowed; transient failures masked as not-found**

`resolveExecutionNoteAgentIDByDID` (lines 207-211) checks `err == nil && record != nil`. `GetDIDDocument` (local.go:8323-8328) returns the same error type for both not-found and DB failures with no sentinel. Transient DB failure → falls through to `ListAgentDIDs` → if that also returns empty, client gets 403 "caller agent identity is required" while the real cause is infrastructure. Error info is lost.

**Fix**: (a) sentinel `ErrDIDNotFound` so only genuine not-found triggers fallback; or (b) log the `GetDIDDocument` error before falling through so operators can detect degradation.

---
*`Reliability` · confidence 85%*

santoshkumarradha (dismissed):
## 🔴 PR-AF Review — **Needs Major Rework**

> **10 findings** · 🔴 3 critical · 🟠 4 important · 🔵 0 suggestions · ⚪ 0 nitpicks

<details>

# Summary

Fixes an IDOR in the execution notes write endpoint by enforcing execution ownership before appending a note.

File-specific changes:

- `control-plane/internal/handlers/execution_notes.go`
  - Resolves the caller agent identity before writing a note.
  - Compares the caller agent ID with the execution owner, `execution.AgentNodeID`.
  - Returns `403 execution_ownership_mismatch` when the caller does not own the execution.
  - Supports DID-authenticated callers by resolving the verified caller DID to an agent ID.

- `control-plane/internal/server/middleware/auth.go`
  - Stores API-key caller identity in Gin context using `CallerAgentIDKey`.
  - Uses `X-Caller-Agent-ID` first, with `X-Agent-Node-ID` as fallback.

- `control-plane/internal/handlers/execution_notes_test.go`
  - Adds coverage for owner write success.
  - Adds coverage for non-owner API-key write returning `403`.
  - Adds coverage for DID-authenticated owner and non-owner behavior.

- `control-plane/internal/server/middleware/auth_test.go`
  - Adds coverage that API-key auth populates caller identity in Gin context.

- `control-plane/internal/handlers/coverage_handlers_90_test.go`
  - Updates the existing successful note-write coverage test to include matching caller identity.

## Testing

- [ ] `./scripts/test-all.sh`
- [x] Additional verification:
  - `cd control-plane && go test ./internal/handlers ./internal/server/middleware`
  - `cd control-plane && go test ./internal/handlers ./internal/server/middleware -coverprofile=/tmp/issue-420.coverprofile`
  - `cd control-plane && golangci-lint run --new-from-rev=upstream/main ./internal/handlers ./internal/server/middleware`
  - Manual E2E curl verification:
    - Agent A adding a note to Agent A’s execution returns `200 OK`
    - Agent A adding a note to Agent B’s execution returns `403 Forbidden`
    - Agent B’s fresh execution remains with `notes: []` after the blocked write

Note: full repo lint currently reports pre-existing unrelated Go lint issues outside this PR’s changed files. Changed-line lint for the touched packages reports `0 issues`.

## Checklist

Not UI-related.

## Related issues

Fixes #420

</details>

### Key Findings

**7 issue(s) should be addressed before merge:**

- 🔴 **Complete ownership enforcement bypass when APIKey is empty and DID auth is off — the default configuration** (`control-plane/internal/server/middleware/auth.go:24`) — When `AuthConfig.APIKey` is empty (the default in ALL deployment configurations: `config/agentfield.yaml`, Docker Compose at `deployments/docker/docker-compose.yml`, and Helm at `deployments/helm/agen…
- 🔴 **Three-tier identity fallback in execution notes handler has no fail-closed mechanism: raw-header tier silently becomes primary identity source when all upstream auth middleware is configuration-disabled** (`control-plane/internal/handlers/execution_notes.go:184`) — The execution notes handler's executionNoteCallerAgentID implements a three-tier identity resolution cascade: (1) verified DID from DIDAuthMiddleware, (2) CallerAgentIDKey context value from APIKeyAut…
- 🔴 **Fixing the default authentication bypass by enabling DID auth silently activates diverging DID resolution paths — a 'fix F4, expose F1' trap** (`control-plane/internal/handlers/execution_notes.go:184`) — F4 establishes that under the default configuration (APIKey empty AND did_auth_enabled false), the execution notes handler accepts unvalidated raw headers as caller identity — a complete ownership enf…
- 🟠 **GetExecutionNotesHandler leaks execution notes to any authenticated caller with no ownership enforcement** (`control-plane/internal/handlers/execution_notes.go:235`) — The PR fixes an IDOR on the write path (AddExecutionNoteHandler) by enforcing execution ownership, but the read path (GetExecutionNotesHandler) remains completely open: any API-key-authenticated calle…
- 🟠 **DID auth middleware provides no defense against no-auth bypass — attacker can simply omit DID headers** (`control-plane/internal/server/middleware/did_auth.go:177`) — The assignment question asks: "Does the DIDAuthMiddleware at routes_middleware.go:77-88 provide any defense when API key auth is off?"  **No, not in any meaningful way.** If `DID.Enabled && DIDAuthEna…
- 🟠 **Two-tier DID resolution reads semantically-different field names (AgentID vs AgentNodeID) from independent tables with no structural equivalence guarantee** (`control-plane/internal/handlers/execution_notes.go:206`) — The function `resolveExecutionNoteAgentIDByDID` at line 206 resolves a caller DID to an agent identifier through two independent code paths that read **differently-named columns from different tables*…
- 🟠 **Type-unsafe CallerAgentIDKey enables silent reversion to attacker-controlled raw-header identity when any middleware writes a non-string value — turning a compile-time type error into a runtime authentication bypass** (`control-plane/internal/handlers/execution_notes.go:189`) — The combination of F2 (CallerAgentIDKey accepts any value type via Gin's `c.Set(key string, value any)`) and F4 (executionNoteCallerAgentID falls through to raw X-Caller-Agent-ID / X-Agent-Node-ID hea…

**Files with findings:** `control-plane/internal/handlers/execution_notes.go`, `control-plane/internal/server/middleware/auth.go`, `control-plane/internal/server/middleware/did_auth.go`

<details>

#### 🔴 Critical (3)

- **Complete ownership enforcement bypass when APIKey is empty and DID auth is off — the default configuration** `control-plane/internal/server/middleware/auth.go:24`
- **Three-tier identity fallback in execution notes handler has no fail-closed mechanism: raw-header tier silently becomes primary identity source when all upstream auth middleware is configuration-disabled** `control-plane/internal/handlers/execution_notes.go:184`
- **Fixing the default authentication bypass by enabling DID auth silently activates diverging DID resolution paths — a 'fix F4, expose F1' trap** `control-plane/internal/handlers/execution_notes.go:184`

#### 🟠 Important (4)

- **GetExecutionNotesHandler leaks execution notes to any authenticated caller with no ownership enforcement** `control-plane/internal/handlers/execution_notes.go:235`
- **DID auth middleware provides no defense against no-auth bypass — attacker can simply omit DID headers** `control-plane/internal/server/middleware/did_auth.go:177`
- **Two-tier DID resolution reads semantically-different field names (AgentID vs AgentNodeID) from independent tables with no structural equivalence guarantee** `control-plane/internal/handlers/execution_notes.go:206`
- **Type-unsafe CallerAgentIDKey enables silent reversion to attacker-controlled raw-header identity when any middleware writes a non-string value — turning a compile-time type error into a runtime authentication bypass** `control-plane/internal/handlers/execution_notes.go:189`

</details>

<details>

**Dimensions Analyzed (15):**

- **No-Auth Mode Identity Spoofing Bypass** — 4 file(s)
- **CallerAgentIDKey Context Semantics Collision** — 3 file(s)
- **GET/POST Execution Notes Authorization Asymmetry** — 2 file(s)
- **DID Resolution Silent Degradation — Error vs Not-Found Conflation** — 3 file(s)
- **Context.Background() in GET Handler Bypasses Request Timeout** — 2 file(s)
- **DID resolution field-name divergence: AgentID vs AgentNodeID** — 3 file(s)
- **CallerAgentIDKey context type contract: non-string write → silent fallback** — 3 file(s)
- **No-auth middleware bypass: mechanical trace from empty APIKey to unauthenticated header reads** — 4 file(s)
- **Storage error propagation contract: errors.As dependency on UpdateExecutionRecord fidelity** — 2 file(s)
- **Context.Background() drift in GET handler: deadline propagation gap vs POST handler** — 2 file(s)
- **No-auth bypass of execution ownership enforcement via raw header fallback** — 4 file(s)
- **Triplicated caller resolution logic with diverged priority chains sharing one context namespace** — 3 file(s)
- **Auth error classification breaks if production storage provider wraps or replaces closure errors** — 2 file(s)
- **APIKeyAuth global broadcast of CallerAgentIDKey changes semantics for ALL authenticated routes** — 2 file(s)
- **DID-to-agent ID resolution produces different identifiers depending on resolution path (AgentID vs AgentNodeID field mismatch)** — 3 file(s)

- **Semantic** — 5 dimension(s), 92% coverage confidence
- **Mechanical** — 5 dimension(s), 92% coverage confidence
- **Systemic** — 5 dimension(s), 92% coverage confidence

- **6** compound finding(s) synthesized
- **12** finding(s) adversarially tested: 12 confirmed, 0 challenged

</details>

<details>

| Metric | Value |
|--------|-------|
| Duration | 4643.7s |
| Agent invocations | 65 |
| Coverage iterations | 0 |
| Budget exhausted | No |
| PR type | bugfix |
| Complexity | standard |

</details>

Review ID: `rev_b6e41625c18a`

<br>
<div align="right">
    AgentField PR-AF
  </a>
</div>

santoshkumarradha:
@Luffy2208, just a heads-up that we're currently evaluating the review quality of https://github.com/Agent-Field/pr-af. If you notice any of these automated findings are off-base, noisy, or unhelpful, please let us know! Your feedback would be super helpful.

control-plane/internal/handlers/execution_notes.go
AbirAbbas: Speculative — both DID resolution paths converge in practice today (both populated from `newNode.ID`), and divergence requires a did:key caller hitting a registration race. Worth unifying as cleanup but not blocking the IDOR fix. Resolving.

control-plane/internal/handlers/execution_notes.go
AbirAbbas: Naming inconsistency is real but both fields are populated from `newNode.ID` ([nodes_register.go:1027,1044](control-plane/internal/services/nodes_register.go)). No bug today — a code comment would suffice. Not blocking. Resolving.

control-plane/internal/handlers/execution_notes.go
AbirAbbas: Speculative compound risk — requires a future middleware bug writing a non-string value AND the raw-header fallback existing. The real fix (removing the raw-header fallback, tracked in the no-auth bypass thread) eliminates this path entirely. Resolving.

control-plane/internal/handlers/execution_notes.go
AbirAbbas: Minor reliability nit, not a security issue. Worth a follow-up to add a sentinel `ErrDIDNotFound` but not blocking the IDOR fix. Resolving.

control-plane/internal/handlers/execution_notes.go
Luffy2208: Agreed. Even though did:web revocation is already enforced by DID auth, the handler resolver should fail closed too. I’ll reject revoked DIDDocumentRecords, skip non-active AgentDIDInfo entries, and add coverage for revoked DID resolution.
Should i do this in this same PR?

control-plane/internal/handlers/execution_notes.go
Luffy2208: Agreed. The read path should mirror the write path here. I’ll add caller resolution + ensureExecutionNoteOwnership to GetExecutionNotesHandler, return 403 on ownership mismatch, and add tests for owner vs non-owner reads.
There's already an issue raised(#421 ) for this 
I will fix it in thats issue's PR.

control-plane/internal/handlers/execution_notes.go
Luffy2208: Agreed. This is a real bypass under the default config. I’ll remove the raw-header fallback, require caller identity to come from verified DID or authenticated middleware context only, and add tests for the spoofed-header/no-auth case. I’ll also add a startup/config guard so the write route fails closed when both API key auth and DID auth are disabled.
in this same PR right?

AbirAbbas: Pushed a fix for this. Dropped the raw-header fallback so the handler only trusts a verified DID or the API-key-validated middleware context — spoofed `X-Caller-Agent-ID`/`X-Agent-Node-ID` no longer resolve to an identity.

One deviation from "refuse to register the route when both auth methods are off": the default config ships with no API key and DID auth disabled, and `app.note()` relies on that path in local/dev, so hard-failing the route closed would break the feature for everyone running without auth. Instead ownership is enforced whenever an auth method is active, and skipped (with a loud startup warning) when the server is fully unauthenticated — which matches the no-auth posture of every other endpoint. Added tests for the spoofed-header and no-auth cases.

control-plane/internal/handlers/execution_notes.go
AbirAbbas: Done in this PR. The resolver now skips revoked `DIDDocumentRecord`s (`IsRevoked()`) and `agent_dids` entries with status `revoked`, so a revoked DID fails closed even if the auth layer let the request through. Added coverage for revoked-document, revoked-registry-entry, and active cases.

AbirAbbas:
merging this pr to avoid it getting stale, pushed up the fixes, just waiting on CI

AbirAbbas (approved):

PR #618, opened 2026-06-05
Tech debt

pocesar:
## Type of change

- [x] Bug fix
- [ ] New feature
- [x] Refactor / cleanup
- [ ] Docs only
- [ ] Tests only
- [ ] CI / tooling
- [ ] Breaking change

## Test plan

- [ ]
- [ ]

## Test coverage

  **85%**.

- [ ] If I removed code, I updated `coverage-baseline.json` in

> the PR summary.

## Checklist

- [ ] I have read [CONTRIBUTING.md](CONTRIBUTING.md) (if present) and
- [ ] I have linked any related issues (Fixes #123 / Closes #456).


control-plane/migrations/034_workflow_execution_retry_count.sql
```diff
@@ -0,0 +1,9 @@
+-- 034_workflow_execution_retry_count.sql
+-- Adds retry_count to workflow_executions for stale-execution auto-retry.
+-- RetryStaleWorkflowExecutions increments this column when resetting stuck runs.
+
+-- +goose Up
+ALTER TABLE workflow_executions ADD COLUMN IF NOT EXISTS retry_count INTEGER NOT NULL DEFAULT 0;
```
santoshkumarradha: This migration breaks the current SQLite test path. `ALTER TABLE ... ADD COLUMN IF NOT EXISTS` is not accepted by the SQLite version our suite uses, so `go test ./...` fails during storage initialization with `near "EXISTS": syntax error`. Please switch this to the same compatibility pattern we use in earlier migrations so both fresh SQLite DBs and already-migrated DBs work.

compose.yaml
```diff
@@ -0,0 +1,78 @@
...
+    build:
+      context: .
+      dockerfile: deployments/docker/Dockerfile.python-agent
+    command:
+      - python
+      - -c
+      - from agent import agent;
+        agent.run()
+    volumes:
+      - ./agent.py:/app/agent.py
```
santoshkumarradha: This compose example points at `./agent.py`, but there is no `agent.py` at the repo root. In a fresh checkout the bind mount source is missing, so the `agent` and `team` services cannot start as written. If this is meant to reuse the triggers demo, please wire it to the real file under `examples/triggers-demo/` or add the missing entrypoint alongside this compose file.

santoshkumarradha (dismissed):
Thanks for pushing this cleanup through. I found a couple of concrete regressions while validating locally, so I’m requesting changes before this can land.

I ran `cd control-plane && go test ./...` in a fresh checkout of this PR and the run fails during SQLite setup because the new migration uses `IF NOT EXISTS` on `ALTER TABLE ... ADD COLUMN`, which SQLite does not support in the version our tests exercise. I also checked the new root `compose.yaml`, and the agent services point at `./agent.py`, but this repo does not have that file at the root, so the example stack will not start as written.

Also worth noting: I only see the CLA check on the PR right now, so the normal CI gates do not appear to have run yet.

santoshkumarradha:
Thanks for the cleanup pass. I did a security-focused read because this PR touches a lot of sensitive surfaces. I don’t see evidence of a supply-chain/backdoor style issue, but I think this should be split up and fixed before merge.

Main blockers:
- The new storage migration breaks fresh SQLite/local startup (`migration 016: near "EXISTS": syntax error`), and it also appears to add then drop `retry_count` in the embedded migration.
- The root `compose.yaml` hardcodes `postgres/postgres`, `abc123`, and `admin123`, exposes ports, and references a root `agent.py` that does not exist.
- Telemetry/docs changes are not just cleanup for us. Anonymous OSS telemetry and its public disclosure are critical, so flipping telemetry off and removing the docs/privacy context needs a dedicated discussion.
- The new frontend `cpClient` should not attach API/admin tokens to arbitrary absolute URLs; it should stay same-origin/path-only before it becomes a shared helper.
- Identity/DID/VC and admin gRPC removals are security/audit surface changes, not tech debt, and need their own PR with rationale and replacement plan.

I’d recommend reducing this into small focused PRs: `.DS_Store` cleanup, migration fix, compose example, frontend API client refactor, and any identity/admin removals separately.

pocesar:
@santoshkumarradha sorry about the sloppy PR, I was in a rush yesterday. I'll split them properly. Rest assured it's not an account takeover 😆

AbirAbbas:
Picked this back up since it had been idle for a while and was conflicting with main. Rather than land the full change, I've reduced it to the subset that's a clean, zero-regression cleanup and rebased it onto current main (it's now mergeable):

**Kept** (genuinely dead code, ~1.5k lines):
- `ui/identity.go` + its tests — the DID Explorer/Credentials backend. The frontend service that called it (`identityApi.ts`) has no remaining consumers, and the DID Explorer pages were already removed (App.tsx redirects `/identity/dids` and `/identity/credentials` to `/settings`).
- `identityApi.ts` + its test — orphaned frontend service.
- `config/docker-perf.yaml` — unreferenced.

**Left out** because they'd be regressions and deserve their own focused PRs:
- Removing the `/admin/public-key` alias — all three SDKs still call it for offline VC verification (`af verify` would break).
- Removing `/actions/claim` + node start/stop/lifecycle — documented live API used by poll-mode agents.
- Removing the legacy reasoner execute endpoints — no in-repo caller, but external SDK versions may still hit them; needs a deprecation cycle first.
- The root `compose.yaml` (references a non-existent `agent.py`/`team.py` and hardcodes creds) and the storage-mode/telemetry config default flips.

Thanks for the original cleanup pass @pocesar — credited on the commit.

AbirAbbas (approved):

PR #631, opened 2026-06-06
fix: Updated the startup and shutdown events, now they are using life…

### LINKED ISSUE #628: [Python SDK] Using deprecated shutdown and startup events
## Describe the bug

FastAPI warns during the startup that the `AgentServer` class is using deprecated `startup` and `shutdown` events, but should be using `lifespan` instead

https://fastapi.tiangolo.com/advanced/events/?h=life#lifespan

## Steps to reproduce

2. Run '...'
3. See error

## Environment

- Control plane version:

DebanKsahu:
## Summary

I changed the startup and shutdown events from python sdk. Previously it was using deprecated `on_event()` method but I replaced it with lifespan which handle both shutdown and startup event.

## Type of change

- [x] Bug fix
- [ ] New feature
- [x] Refactor / cleanup
- [ ] Docs only
- [ ] Tests only
- [ ] CI / tooling
- [ ] Breaking change

## Test plan

- [x] `./scripts/test-all.sh`
- [x] `./scripts/coverage-summary.sh` was run locally, but it currently stops at sdk-python because the local `.venv` is missing `pip`.
- [x] Verified the branch rebases cleanly on latest `upstream/main`.

## Checklist

## Related issues / PRs
Fixes #628

### DIFF
--- a/sdk/python/agentfield/agent_server.py
+++ b/sdk/python/agentfield/agent_server.py
@@ -1,3 +1,5 @@
+from contextlib import asynccontextmanager
+from fastapi import FastAPI
 import asyncio
 import importlib.util
 import os
@@ -48,7 +50,10 @@ class AgentServer:
             if not node_logs.logs_enabled():
                 return JSONResponse(
                     status_code=404,
-                    content={"error": "logs_disabled", "message": "Process logs API is disabled"},
+                    content={
+                        "error": "logs_disabled",
+                        "message": "Process logs API is disabled",
+                    },
                 )
             auth = request.headers.get("authorization") or request.headers.get(
                 "Authorization"
@@ -56,7 +61,10 @@ class AgentServer:
             if not node_logs.verify_internal_bearer(auth):
                 return JSONResponse(
                     status_code=401,
-                    content={"error": "unauthorized", "message": "Valid Authorization Bearer required"},
+                    content={
+                        "error": "unauthorized",
+                        "message": "Valid Authorization Bearer required",
+                    },
                 )
             qp = request.query_params
             try:
@@ -107,7 +115,9 @@ class AgentServer:
                     name = t.get_name()
                 except Exception:
                     name = "?"
-                buf.write(f"=== Task {name} done={t.done()} cancelled={t.cancelled()} ===\n")
+                buf.write(
+                    f"=== Task {name} done={t.done()} cancelled={t.cancelled()} ===\n"
+                )
                 try:
                     coro = t.get_coro()
                     buf.write(f"coro: {coro!r}\n")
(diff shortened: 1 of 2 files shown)


sdk/python/agentfield/agent_server.py
```diff
@@ -766,8 +785,18 @@ def serve(
...
+        async def lifespan(app: FastAPI):
+            # Add startup event handler for resilient lifecycle
+            await startup_resilient_lifecycle()
+            try:
+                yield
+            finally:
+                # Add shutdown event handler for cleanup
+                await shutdown_cleanup()
+
+        self.agent.router.lifespan_context = lifespan
```
santoshkumarradha: This assignment replaces any lifespan callback the caller already passed into `Agent(..., **kwargs)`, because `Agent` forwards those kwargs into `FastAPI` in its constructor. Today those callers can use FastAPI lifecycle hooks alongside the SDK. With this change, their lifespan stops running as soon as `serve()` is called.

Can we wrap the existing `self.agent.router.lifespan_context` instead of overwriting it, so the deprecation fix keeps the current API behavior intact?

santoshkumarradha (changes requested):
Thanks for taking this on. Moving off the deprecated startup/shutdown hooks is the right direction. I left one blocker inline because this version replaces the app lifespan outright, which can break callers that already pass their own FastAPI lifespan into Agent.

sdk/python/agentfield/agent_server.py
DebanKsahu: Yeah, you are right, my current way will override the lifespan provided by callers. sorry for not considering this edge case, but I have a suggestion, if we could merge the 2 lifespan (use's and python sdk's) then it could solve the problem. `contextlib` have a util named `AsyncExitStack` which generally used for this type of merging, which also remember both entry and exit of all the lifespan. I will check the resource and push a new commit.

sdk/python/agentfield/agent_server.py
```diff
@@ -766,8 +786,27 @@ def serve(
...
+                await shutdown_cleanup()
+
+        existing_lifespan = self.agent.router.lifespan_context
+
+        @asynccontextmanager
+        async def merged_lifespan(app: FastAPI):
+            async with AsyncExitStack() as stack:
+                await stack.enter_async_context(internal_lifespan(app))
+                await stack.enter_async_context(existing_lifespan(app))
+            yield
```
santoshkumarradha: Thanks for the follow-up. There is still one blocker here: `yield` runs after the `AsyncExitStack` block, so both lifespan contexts are entered and exited before FastAPI starts serving. That means startup resources are torn down immediately instead of staying alive for the request lifecycle. The merged lifespan needs to `yield` while the stack is still open, then let the stack unwind during shutdown.

DebanKsahu: Sorry, for the mistake, I will resolve the issue now.

DebanKsahu: I made the changes, you can check.

santoshkumarradha (approved):
Thanks for iterating on this. I added a small regression test for the merged lifespan path and the targeted SDK checks pass locally.

santoshkumarradha:
@CLAassistant recheck

PR #635, opened 2026-06-08
Fix get_logger returning agentfield.result_cache for all loggers

AbirAbbas:
## Summary

- Fixed  to return loggers with correct names instead of always returning 
- Implemented dictionary-based cache () to store loggers by their name
- Maintained shared configuration via  for consistent log level management

## Changes

- : Replaced global singleton pattern with dictionary-based caching
- : Updated logger imports
- : Added unit tests for new caching behavior

## Test plan

- All 5 acceptance criteria verified passing
- Unit tests in  pass
- Each unique logger name produces a logger with that name
-  properly propagates to all cached and new loggers

🤖 Built with [AgentField SWE-AF](https://github.com/Agent-Field/SWE-AF)
🔌 Powered by [AgentField](https://github.com/Agent-Field/agentfield)


AbirAbbas (commented):
## 🟡 PR-AF Review — **Minor Issues**

> **6 findings** · 🔴 0 critical · 🟠 3 important · 🔵 1 suggestions · ⚪ 2 nitpicks

<details>

## Summary

- Fixed  to return loggers with correct names instead of always returning 
- Implemented dictionary-based cache () to store loggers by their name
- Maintained shared configuration via  for consistent log level management

## Changes

- : Replaced global singleton pattern with dictionary-based caching
- : Updated logger imports
- : Added unit tests for new caching behavior

## Test plan

- All 5 acceptance criteria verified passing
- Unit tests in  pass
- Each unique logger name produces a logger with that name
-  properly propagates to all cached and new loggers

🤖 Built with [AgentField SWE-AF](https://github.com/Agent-Field/SWE-AF)
🔌 Powered by [AgentField](https://github.com/Agent-Field/agentfield)

</details>

### Key Findings

**3 issue(s) should be addressed before merge:**

- 🟠 **Race condition in get_logger() dictionary access** (`sdk/python/agentfield/logger.py:465`) — The `_logger_cache` dictionary is accessed without any locking mechanism.
- 🟠 **Thread unsafe iteration in set_log_level()** (`sdk/python/agentfield/logger.py:476`) — The `set_log_level()` function iterates over `_logger_cache.values()` (line 481) while other threads could be modifying the cache (e.g., in `get_logger()`).
- 🟠 **Thread unsafe iteration in set_cp_client()** (`sdk/python/agentfield/logger.py:485`) — Same issue as set_log_level().

**3 suggestion(s) and style note(s):**

- 🔵 Test file directly imports private implementation details (`sdk/python/tests/test_logger.py:3`)
- ⚪ Test imports and resets private module variables - no runtime impact (`sdk/python/tests/test_logger.py:8`)
- ⚪ set_log_level() behavior change not fully documented (`sdk/python/agentfield/logger.py:476`)

**Files with findings:** `sdk/python/agentfield/logger.py`, `sdk/python/tests/test_logger.py`

<details>

#### 🟠 Important (3)

- **Race condition in get_logger() dictionary access** `sdk/python/agentfield/logger.py:465`
- **Thread unsafe iteration in set_log_level()** `sdk/python/agentfield/logger.py:476`
- **Thread unsafe iteration in set_cp_client()** `sdk/python/agentfield/logger.py:485`

#### 🔵 Suggestion (1)

- **Test file directly imports private implementation details** `sdk/python/tests/test_logger.py:3`

#### ⚪ Nitpick (2)

- **Test imports and resets private module variables - no runtime impact** `sdk/python/tests/test_logger.py:8`
- **set_log_level() behavior change not fully documented** `sdk/python/agentfield/logger.py:476`

</details>

<details>

**Dimensions Analyzed (2):**

- **Logger name parameter contract violation** — 1 file(s)
- **Test file coupling to private implementation details** — 1 file(s)

- **Semantic** — 4 dimension(s), 90% coverage confidence
- **Mechanical** — 5 dimension(s), 85% coverage confidence
- **Systemic** — 3 dimension(s), 85% coverage confidence

- **6** finding(s) adversarially tested: 6 confirmed, 0 challenged

</details>

<details>

| Metric | Value |
|--------|-------|
| Duration | 1636.0s |
| Agent invocations | 13 |
| Coverage iterations | 2 |
| Budget exhausted | No |
| PR type | bugfix |
| Complexity | trivial |

</details>

Review ID: `rev_9697f5f51f23`

<br>
<div align="right">
    AgentField PR-AF
  </a>
</div>

sdk/python/agentfield/logger.py
```diff
@@ -454,28 +455,37 @@ def log_execution(
...
-# Global logger instance
-_global_logger = None
+# Global logger cache: name -> AgentFieldLogger instance
+_logger_cache: Dict[str, AgentFieldLogger] = {}
+
+# Global log level override (set via set_log_level)
+_global_log_level: Optional[str] = None
 
 
 def get_logger(name: str = "agentfield") -> AgentFieldLogger:
```
AbirAbbas, line 481: 🟠 **[IMPORTANT] Race condition in get_logger() dictionary access**

The `_logger_cache` dictionary is accessed without any locking mechanism. The check-then-act pattern at lines 468-473 (`if name not in _logger_cache: ... _logger_cache[name] = logger`) creates a race condition where two concurrent calls to `get_logger()` with the same name could both pass the `not in` check before either inserts, resulting in duplicate logger instances or one being silently discarded. Under realistic concurrent load (e.g., multiple agents initializing simultaneously), this can cause inconsistent logger instances and potential data corruption.

---

> Line 468: `if name not in _logger_cache:` - check without lock
> Line 472: `_logger_cache[name] = logger` - insert without lock
> Concurrent callers could both pass the check before either insert.

**💡 Suggested Fix**

Add a threading lock to protect all accesses to `_logger_cache`. Example: Use `threading.Lock()` to wrap the check-and-set operation in `get_logger()`, or use a thread-safe dict implementation. The lock should also protect `set_log_level()` and `set_cp_client()` which iterate over the cache.

---
*`Thread safety and implementation gap review` · confidence 85%*

sdk/python/agentfield/logger.py
```diff
@@ -454,28 +455,37 @@ def log_execution(
...
-    return _global_logger
+    if name not in _logger_cache:
+        logger = AgentFieldLogger(name)
+        if _global_log_level is not None:
+            logger.set_level(_global_log_level)
+        _logger_cache[name] = logger
+    return _logger_cache[name]
 
 
 def set_log_level(level: str):
```
AbirAbbas, line 495: 🟠 **[IMPORTANT] Thread unsafe iteration in set_log_level()**

The `set_log_level()` function iterates over `_logger_cache.values()` (line 481) while other threads could be modifying the cache (e.g., in `get_logger()`). This can cause `RuntimeError: dictionary changed size during iteration` or lead to missed updates if loggers are added during iteration. The global `_global_log_level` assignment (line 480) is also not atomic with respect to reads in `get_logger()`.

---

> Line 481: `for logger in _logger_cache.values():` - iterates without lock
> Other threads calling `get_logger()` could modify `_logger_cache` during this iteration.

**💡 Suggested Fix**

Protect the iteration with the same lock used in `get_logger()`. Acquire the lock before reading `_logger_cache` and hold it for the duration of the iteration to prevent concurrent modification.

---
*`Thread safety and implementation gap review` · confidence 80%*

sdk/python/agentfield/logger.py
```diff
@@ -454,28 +455,37 @@ def log_execution(
...
+    """Set log level for all logger instances at runtime (e.g., 'DEBUG', 'INFO', 'WARN', 'ERROR')"""
 
-    get_logger().set_level(level)
+    global _global_log_level
+    _global_log_level = level
+    for logger in _logger_cache.values():
+        logger.set_level(level)
 
 
 def set_cp_client(client: Optional["AgentFieldClient"]) -> None:
```
AbirAbbas, line 514: 🟠 **[IMPORTANT] Thread unsafe iteration in set_cp_client()**

Same issue as set_log_level(). The `set_cp_client()` function iterates over `_logger_cache.values()` (line 487) without locking, creating the same race condition risk. While the assignment to `logger._cp_client` is a simple operation, the iteration itself is unsafe.

---

> Line 487: `for logger in _logger_cache.values():` - iterates without lock while other threads may modify the cache.

**💡 Suggested Fix**

Protect with the same lock used elsewhere. Acquire the lock before iterating over cached loggers.

---
*`Thread safety and implementation gap review` · confidence 80%*

sdk/python/tests/test_logger.py
```diff
@@ -0,0 +1,91 @@
+import pytest
+
+from agentfield.logger import (
```
AbirAbbas, line 7: 🔵 **[SUGGESTION] Test file directly imports private implementation details**

The test file imports `_logger_cache` and `_global_log_level` directly (lines 8-9) to reset state between tests. This creates tight coupling to internal implementation details. If the caching mechanism is refactored (e.g., to use a class-based solution or external cache), these tests will break. The fixture at lines 13-21 directly mutates these private globals.

---

> Line 8: `from agentfield.logger import (... _logger_cache, _global_log_level,)`
> Line 16: `_logger_cache.clear()`
> Line 17-18: `global _global_log_level; _global_log_level = None`

**💡 Suggested Fix**

Consider exposing a `reset_logger_cache()` function in the public API (or a test-specific API) that encapsulates the reset logic. This would allow internal refactoring without breaking tests. Alternatively, document that these are test-only internals with no backward compatibility guarantees.

---
*`Thread safety and implementation gap review` · confidence 75%*

sdk/python/tests/test_logger.py
```diff
@@ -0,0 +1,91 @@
...
+    AgentFieldLogger,
+    get_logger,
+    log_info,
+    set_log_level,
+    _logger_cache,
```
AbirAbbas: ⚪ **[NITPICK] Test imports and resets private module variables - no runtime impact**

The test file directly imports `_logger_cache` and `_global_log_level` (lines 8-9) and the fixture resets them (lines 13-21). However, **no runtime issues exist**: (1) The fixture runs in isolated test context with autouse=True, resetting state before/after each test; (2) Production code never resets these variables - `_global_log_level` is only assigned in `set_log_level()` at logger.py:480; (3) There's no scenario where test resets could affect production code since they run in separate contexts. The coupling is a design concern (tests break if implementation changes), not a functional runtime bug.

---

> Production code grep shows only one assignment to _global_log_level in production: logger.py:480 in set_log_level(). No clear() or reset operations exist in production. Test fixture is self-contained.

---
*`Test coupling to private module variables` · confidence 95%*

sdk/python/agentfield/logger.py
AbirAbbas, line 495: ⚪ **[NITPICK] set_log_level() behavior change not fully documented**

The PR context notes that `set_log_level()` behavior changed: 'Old code would create a logger if none existed when called; new code only updates existing cached loggers.' The new implementation at lines 476-482 does NOT create a logger if none exists—it only stores the global level and updates existing loggers. This is semantically correct (new loggers get the level when created via get_logger lines 470-471), but the behavior difference could surprise callers who previously relied on the side effect of logger creation.

---

> Lines 479-482 set _global_log_level and iterate over existing cache, but do not create a logger if cache is empty.

**💡 Suggested Fix**

Add a docstring note about this behavioral change, or consider whether backward compatibility requires creating a logger if none exists.

---
*`Thread safety and implementation gap review` · confidence 60%*

sdk/python/agentfield/logger.py
AbirAbbas: Fixed in `a0265076`. `get_logger()` now performs the check-and-insert under a reentrant `_logger_cache_lock`, so two concurrent calls for the same name can no longer both pass the `not in` check.

sdk/python/agentfield/logger.py
AbirAbbas: Fixed in `a0265076`. `set_log_level()` now snapshots `list(_logger_cache.values())` under the lock, then applies the level outside it — so a concurrent `get_logger()` insert cannot mutate the dict mid-iteration (no more `RuntimeError`), and `_global_log_level` is assigned under the same lock.

sdk/python/agentfield/logger.py
AbirAbbas: Fixed in `a0265076`. `set_cp_client()` uses the same snapshot-under-lock pattern as `set_log_level()`.

sdk/python/tests/test_logger.py
AbirAbbas: Addressed in `46418462`, with a deliberate deviation from the suggestion: rather than add a public `reset_logger_cache()` to the production API solely for test setup (production surface for a test-only need), the fixture now references the internals via the module (`agentfield.logger._logger_cache`) and documents them as test-only. Note the original `global _global_log_level = None` reset was silently broken — it rebound this test module's copy, not the real module global — that is now fixed too. Happy to add a public test helper instead if you prefer.

sdk/python/tests/test_logger.py
AbirAbbas: Acknowledged — no runtime impact, as noted. Coupling is reduced to a module reference and documented as test-only in `46418462`. Resolving since the functional concern is nil.

sdk/python/agentfield/logger.py
AbirAbbas: Documented in `e1673a0d` — added a docstring note that `set_log_level()` no longer implicitly creates the default logger; loggers created later via `get_logger()` pick up the stored level on creation.

huangyoufeng76-debug (approved):

sdk/python/tests/test_logger.py
```diff
@@ -0,0 +1,157 @@
...
+
+import agentfield.logger as logger_module
+from agentfield.logger import (
+    get_logger,
+    log_info,
+    set_log_level,
+)
+
+
+@pytest.fixture(autouse=True)
```
AbirAbbas, line 15: 🟠 **[IMPORTANT] Cross-test-suite propagate state leakage — reset_logger_state fixture exists only in test_logger.py, not in shared conftest.py**

The `reset_logger_state` autouse fixture in `test_logger.py:15-56` snapshots and restores `logging.Logger.propagate` flags and handlers across every test in **this module only**. Its docstring (lines 26-30) explicitly documents the risk:

> *'creating the "agentfield" logger here would stop "agentfield.cancel" records from reaching root handlers (pytest's caplog), breaking unrelated tests like test_cancel.py.'*

This fixture is necessary because `AgentFieldLogger._setup_logger()` mutates stdlib `logging.Logger` objects (adding handlers, setting `propagate=False`) — mutations that **outlive the test** and leak across the entire pytest session. **No equivalent fixture exists in `conftest.py`** — the shared conftest has `_isolate_execution_context` (conftest.py:71-115) for ContextVar state, and `_EnvPatcher` for environment variables, but nothing for stdlib logging state.

Consequences:
1. `test_cancel.py:138-213` had to switch from pytest's `caplog` to a **direct handler** attached to `"agentfield.cancel"` (lines 185-201) because caplog (root-handler-based) broke when `test_logger.py` ran first and leaked `propagate=False`.
2. Any future test file that creates `AgentFieldLogger` instances without a similar reset fixture will corrupt the stdlib logging state for **all subsequent tests** in the session, causing order-dependent failures that are extremely hard to debug.
3. If `test_logger.py` ever moves or the fixture is accidentally removed, `test_cancel.py` will fail silently (its direct-handler approach now works independently, but other tests won't).

---

> Step 1: test_logger.py:14-56 defines `reset_logger_state` as an autouse fixture, saved at line 34: `name: (lgr.propagate, list(lgr.handlers))`.
> Step 2: test_logger.py:14 — `@pytest.fixture(autouse=True)` — this only applies to tests in test_logger.py.
> Step 3: conftest.py (full 781 lines) — no `reset_logger_state` fixture; no logging restore logic.
> Step 4: test_cancel.py:146-152 docstring: "once any AgentField logger is created, agentfield.cancel records no longer reach the root logger where caplog's handler lives — which made this assertion silently order-dependent."
> Step 5: test_cancel.py:185-201 — manually attaches a `_Capture` handler directly to `logging.getLogger("agentfield.cancel")` instead of using caplog.
> Step 6: Any test in a future file that calls `get_logger()` will mutate `logging.Logger.propagate` and `logging.Logger.handlers` in the stdlib global `manager.loggerDict` — these persist for the remainder of the pytest session.

**💡 Suggested Fix**

Move the `reset_logger_state` fixture from `test_logger.py` into `conftest.py` as a session-scoped or autouse fixture, so all tests in the suite are protected from propagating state leakage. Alternatively, at minimum add it to `conftest.py` as an `@pytest.fixture(autouse=True)` fixture that covers the entire test session. This complements the existing `_isolate_execution_context` and `env_patch` fixtures that already handle other state isolation concerns.

---
*`Propagation Semantics Change` · confidence 90%*

sdk/python/agentfield/logger.py
```diff
@@ -454,28 +459,54 @@ def log_execution(
...
+    # Snapshot under the lock so a concurrent get_logger() can't mutate the dict
+    # mid-iteration; apply levels outside the lock to avoid holding it during I/O.
+    with _logger_cache_lock:
+        _global_log_level = level
+        loggers = list(_logger_cache.values())
+    for logger in loggers:
+        logger.set_level(level)
 
 
 def set_cp_client(client: Optional["AgentFieldClient"]) -> None:
```
AbirAbbas, line 514: 🟠 **[IMPORTANT] verification.py logger missed by set_cp_client causes silent structured log loss when local_verification=True**

**Root cause**: `set_cp_client()` at `logger.py:504-509` snapshots currently-cached loggers and sets `_cp_client` as an instance attribute on each. Loggers created afterward inherit the class-level `_cp_client = None` (logger.py:56). `_dispatch_to_cp()` at `logger.py:187-191` silently returns when `_cp_client is None` — no warning, no error.

**Affected module**: `verification.py:17` runs `logger = get_logger(__name__)` at module import time. This import happens at `agent.py:779` (`from agentfield.verification import LocalVerifier`), which is **89 lines after** `set_cp_client(self.client)` at `agent.py:690`. Condition: only when `Agent(node_id=..., local_verification=True)`.

**Consequence**: The `agentfield.verification` logger will never forward structured execution logs to the control plane. All 20+ log calls in verification.py (including `logger.debug(...)` at line 95, `logger.warning(...)` at lines 97, 100, etc.) that run within an execution context will silently drop telemetry. This includes cache refresh failures, policy fetch errors, signature verification warnings — all invisible to the control plane.

**The other 7 modules** (`client.py:123`, `did_manager.py:15`, `vc_generator.py:16`, `async_execution_manager.py:33`, `result_cache.py:20`, `did_auth.py:17`, `http_connection_manager.py:21`) are imported via agent.py's top-level imports (lines 33, 41, 42, 58) or their cascade imports (client.py → result_cache, did_auth, async_execution_manager → http_connection_manager), which all execute **before** `Agent.__init__` begins, so they ARE in the cache when `set_cp_client()` fires and receive the client correctly.

---

> Step 1: `Agent.__init__` calls `set_cp_client(self.client)` at `agent.py:690`, distributing the client to the 7 currently-cached loggers (`agentfield`, `agentfield.client`, `agentfield.did_manager`, `agentfield.vc_generator`, `agentfield.async_execution_manager`, `agentfield.result_cache`, `agentfield.did_auth`, `agentfield.http_connection_manager`).
> 
> Step 2: `agent.py:779`: `from agentfield.verification import LocalVerifier` executes (when `local_verification=True`), importing verification.py which triggers module-level `logger = get_logger(__name__)` at `verification.py:17`.
> 
> Step 3: `get_logger("agentfield.verification")` creates a new `AgentFieldLogger` instance at `logger.py:478`. The `_cp_client` attribute lookup on this instance falls through to the class-level `_cp_client = None` at `logger.py:56` because `set_cp_client` never set an instance attribute on it.
> 
> Step 4: When verification code calls e.g. `logger.warning("Failed to fetch policies: {e}")` (verification.py:100) within an execution context, `_emit_optional_structured` → `_emit_structured_record` → `_dispatch_to_cp()` at `logger.py:187-191` checks `client = self._cp_client` → `None` → `return` (silent no-op).
> 
> Step 5: The test fixture at `test_trigger_param_binding.py:50-76` explicitly documents this class of problem (lines 55-58): "Agent.__init__ calls set_cp_client(self.client), repointing the GLOBAL logger's control-plane client at this test's fake client...".

**💡 Suggested Fix**

**Option A (targeted)**: Store `_cp_client` as a module-level variable (like `_global_log_level`) and apply it to new loggers in `get_logger()`:
```python
_global_cp_client: Optional["AgentFieldClient"] = None

def set_cp_client(client):
    global _global_cp_client
    with _logger_cache_lock:
        _global_cp_client = client
        loggers = list(_logger_cache.values())
    for logger in loggers:
        logger._cp_client = client

def get_logger(name="agentfield"):
    with _logger_cache_lock:
        if name not in _logger_cache:
            logger = AgentFieldLogger(name)
            if _global_log_level is not None:
                logger.set_level(_global_log_level)
            if _global_cp_client is not None:
                logger._cp_client = _global_cp_client
            _logger_cache[name] = logger
        return _logger_cache[name]
```

**Option B (structural)**: Move the `_cp_client` from a per-logger instance attribute to a module-level singleton that `_dispatch_to_cp()` reads directly — eliminating the per-instance distribution entirely. This matches the semantics of the old single-singleton model while preserving name-keyed loggers.

**Option C (immediate mitigation)**: In `Agent.__init__`, reverse the order — import verification.py before calling `set_cp_client()`. However this only fixes verification.py and leaves future modules vulnerable to the same ordering bug.

---
*`set_cp_client Initialization Ordering Dependency` · confidence 85%*

sdk/python/agentfield/logger.py
```diff
@@ -75,8 +75,9 @@ def __init__(self, name: str = "agentfield"):
 
     def set_level(self, level: str):
         """Set log level at runtime (e.g., 'DEBUG', 'INFO', 'WARN', 'ERROR')"""
-
-        self.logger.setLevel(_LEVEL_TO_LOGGING.get(level.upper(), logging.INFO))
+        level_upper = level.upper()
+        self.log_level = level_upper
+        self.logger.setLevel(_LEVEL_TO_LOGGING.get(level_upper, logging.INFO))
 
     def _setup_logger(self):
```
AbirAbbas, line 82: 🟠 **[IMPORTANT] Unbounded logger cache entries compound propagate=False damage — each dynamic-name AgentFieldLogger permanently silences its stdlib namespace branch from root handlers**

Two independently-identified issues combine into cumulative semantic degradation of the stdlib logging hierarchy that grows over the process lifetime:

**Finding A**: `_logger_cache` (logger.py:463) is a plain `Dict` with no size limit, TTL, or eviction — downstream integrators passing dynamic names like `get_logger(f"workflow.{workflow_id}")` would grow it without bound.

**Finding B**: `_setup_logger()` (logger.py:90) unconditionally sets `self.logger.propagate = False` on the underlying `logging.Logger` for every AgentFieldLogger instance.

**Compound mechanism**: Each cache entry creates a `logging.Logger` via `logging.getLogger(name)` (logger.py:59), adds a `StreamHandler` (logger.py:86-89), and sets `propagate=False` (logger.py:90). The cache holds a reference to the `AgentFieldLogger` which holds the `logging.Logger`, preventing GC. As the cache grows with dynamic names, an ever-increasing portion of the stdlib logging namespace tree is permanently severed from root handlers. Records emitted to these namespaces — whether through the AgentFieldLogger or through raw `logging.getLogger()` calls sharing the same name — silently never reach root-attached handlers (production monitoring, structured log collectors, pytest caplog). This is not merely memory pressure (Finding A's concern) nor a single fixed logger blocking one other (Finding B's concern), but a progressive, unbounded degradation where the set of silently-suppressed namespaces grows monotonically with every dynamic cache insertion.

**Existing evidence**: After `get_logger('agentfield')` is called, records from `logging.getLogger('agentfield.cancel').info(...)` produce zero records at root — this demonstrates the mechanism works for any sub-namespace. With dynamic names, this effect would expand to every dynamically-named branch.

**Production modules at risk**: Four modules use raw `logging.getLogger()` under the 'agentfield' namespace (cancel.py:35, opencode.py:23, claude.py:151, _runner.py:24). If a dynamically-named AgentFieldLogger silently sets propagate=False on an intermediate namespace node, these raw loggers downstream would silently lose root-handler delivery.

**Current state**: The cache is bounded in practice — all 8 production callers use static `__name__` values. The risk is that the API design tacitly invites dynamic names without any guardrail, and the damage from unbounded growth is semantic (propagation loss) not merely resource (memory).

---

> Step 1: logger.py:463 — `_logger_cache: Dict[str, AgentFieldLogger] = {}` — plain unbounded dict.
> Step 2: logger.py:477-481 — `get_logger()` inserts unconditionally on cache miss with no guardrail.
> Step 3: logger.py:59 — `AgentFieldLogger.__init__` calls `logging.getLogger(name)`, creating a stdlib Logger in the global registry.
> Step 4: logger.py:90 — `_setup_logger()` sets `self.logger.propagate = False` unconditionally on every instance.
> Step 5: logger.py:86-89 — `_setup_logger()` adds a `StreamHandler` to the stdlib Logger.
> Step 6: Verified live: after `get_logger('agentfield')`, records from `logging.getLogger('agentfield.cancel')` produce 0 records at root handler — propagate=False stops traversal at the agentfield parent.
> Step 7: test_execution_logger.py:170-171 — the `base_logger` fixture manually works around propagate=False with `logger.logger.propagate = True`, confirming the mutation occurs and requires manual correction.
> Step 8: Each cache entry = one more stdlib Logger with propagate=False + StreamHandler that lives for the process lifetime (cache prevents GC).

**💡 Suggested Fix**

Implement two changes together: (1) Bound the cache with an LRU policy (e.g., max 256 entries, log a warning on eviction). (2) In `_setup_logger()`, set `propagate=False` only on the root 'agentfield' logger (i.e., when `self.logger.name == 'agentfield'`), leaving child loggers with default propagate=True so raw stdlib loggers under the same namespace can still reach root handlers. Fixing only the cache still leaves the propagation severance on whatever names remain cached. Fixing only propagation still leaves unbounded cache growth as a resource leak.

---
*`Compound Analysis` · confidence 70%*

sdk/python/tests/test_logger.py
```diff
@@ -0,0 +1,157 @@
...
+       test file is order-independent.
+    """
+    manager = logging.root.manager
+    saved = {
+        name: (lgr.propagate, list(lgr.handlers))
+        for name, lgr in manager.loggerDict.items()
+        if isinstance(lgr, logging.Logger)
+    }
+
+    logger_module._logger_cache.clear()
```
AbirAbbas, line 41: 🔵 **[SUGGESTION] `reset_logger_state` fixture directly mutates private `_logger_cache` and `_global_log_level` module internals**

The `reset_logger_state` fixture at lines 39-40 and 44-45 directly accesses `logger_module._logger_cache.clear()` and assigns `logger_module._global_log_level = None`. These are `_`-prefixed module-level variables at `logger.py:463` and `logger.py:466` — private by Python convention. There is no public `reset()` or `clear()` function on the logger module.

This creates tight coupling between test code and implementation details. If the `_logger_cache` representation changes — for example, switching from a plain `dict` to a custom cache class with an internal index, or splitting into per-process dictionaries — the fixture silently breaks at `.clear()` or silently misses state. The tests may pass but no longer actually reset state.

By contrast, `ResultCache.clear()` at `result_cache.py:271` is a **public** method on a class — the module owns its own reset contract. Similarly, `test_trigger_param_binding.py:70` restores the control-plane client via the **public** `set_cp_client(prev_cp_client)` API rather than mutating `AgentFieldLogger._cp_client` directly (even though line 64 reads that private attribute for the snapshot).

No behavioral defect exists today — `_logger_cache` is a plain `dict` and `.clear()` works. The risk is future fragility: any internal refactoring of the logger module's state representation forces a corresponding test update, even though public behavior hasn't changed.

---

> Step 1: `reset_logger_state` fixture in `sdk/python/tests/test_logger.py:39` calls `logger_module._logger_cache.clear()`.
> Step 2: `_logger_cache` is defined at `sdk/python/agentfield/logger.py:463` as `_logger_cache: Dict[str, AgentFieldLogger] = {}` — a `_`-prefixed private module variable.
> Step 3: `reset_logger_state` at line 40 assigns `logger_module._global_log_level = None` — `_global_log_level` is also `_`-prefixed at `logger.py:466`.
> Step 4: No public `reset()` or `clear()` function exists anywhere in `agentfield/logger.py` that would abstract over both of these private internals.
> Step 5: Contrast with `ResultCache.clear()` at `result_cache.py:271` which is a public method on a public class — the module controls its own reset contract.

**💡 Suggested Fix**

Add a public `reset_logger_state()` function to `agentfield/logger.py` that encapsulates both `_logger_cache.clear()` and `_global_log_level = None`. Then the fixture calls that function instead of reaching into private module internals. Optionally, also provide a public `get_cp_client()` accessor so fixtures like `test_trigger_param_binding.py:64` don't need to read `._cp_client` directly. This gives the logger module ownership over its own reset semantics and allows internal representation changes without breaking test isolation.

---
*`Test Fixture Isolation: Private Internals & Fragmentation` · confidence 85%*

sdk/python/agentfield/logger.py
```diff
@@ -454,28 +459,54 @@ def log_execution(
         return self._emit_structured_record(record)
 
 
-# Global logger instance
-_global_logger = None
+# Global logger cache: name -> AgentFieldLogger instance
+_logger_cache: Dict[str, AgentFieldLogger] = {}
```
AbirAbbas, line 463: 🔵 **[SUGGESTION] Cache has no size limit or eviction policy — currently bounded but API invites unbounded growth**

The `_logger_cache` at line 463 is a plain `Dict[str, AgentFieldLogger]` with no maximum size, TTL, or LRU eviction. `get_logger()` at lines 473-482 inserts unconditionally on cache miss with no guardrail. **Currently this is bounded** — all 8 production callers (vc_generator.py:16, verification.py:17, http_connection_manager.py:21, result_cache.py:20, did_manager.py:15, did_auth.py:17, client.py:123, async_execution_manager.py:33) use `get_logger(__name__)` with static module names. No examples/ callers exist. The only dynamic pattern is `test_logger.py:144` (`get_logger(f"concurrent.{i}.{j}")`) which creates 400 entries but cleans them via `reset_logger_state`.

However, the public API accepts an arbitrary `name: str` with no validation. A downstream integrator passing `get_logger(f"workflow.{workflow_id}")` would grow the cache without bound. Each entry holds an `AgentFieldLogger`, its underlying `logging.Logger` (persisted in stdlib's global `manager.loggerDict`), and a `StreamHandler` — roughly 5-10 KB per entry. The `StreamHandler` and `logging.Logger` references in stdlib's manager also prevent GC even if the `AgentFieldLogger` wrapper is discarded.

**Risk assessment**: Theoretical given current usage patterns. No dynamic `__name__`, no plugin loading, no `importlib`-based dynamic module creation produces unbounded names today. The gap is in the API design — it tacitly invites dynamic names without any guardrail.

---

> Step 1: `_logger_cache` declared as plain `Dict` at logger.py:463 with no size constraint.
> Step 2: `get_logger()` at logger.py:473-482 does `if name not in _logger_cache: ... _logger_cache[name] = logger` — unconditional insert on miss.
> Step 3: All 8 production callers use `get_logger(__name__)` with static values (e.g., `'agentfield.client'`, `'agentfield.did_manager'`). Grep confirms zero dynamic-name patterns in production or examples.
> Step 4: `test_logger.py:144` demonstrates the pattern that would trigger unbounded growth (`get_logger(f"concurrent.{i}.{j}")`) but `reset_logger_state` clears the cache after each test.
> Step 5: No code path in current production or planned usage creates dynamic logger names — making this a design gap rather than an active bug.

**💡 Suggested Fix**

Add a bounded LRU cache (e.g., `functools.lru_cache` or a capped `OrderedDict`) with a reasonable ceiling (256 entries) and log a warning when evicting. This preserves the name-keyed behavior for bounded static use while preventing silent memory leaks from dynamic callers. Alternatively, document that `get_logger()` names must be static module names and raise `ValueError` for names containing variable-like patterns.

---
*`Unbounded _logger_cache growth with no eviction policy` · confidence 80%*

sdk/python/tests/test_logger.py
AbirAbbas, line 15: 🔵 **[SUGGESTION] Fragmentation of test isolation patterns across modules — no unified global-state reset**

Four different test modules each define their own ad-hoc `autouse` fixtures targeting different slices of SDK global state:

1. **`conftest.py:71-115`** — `_isolate_execution_context()` resets `agent_registry._current_agent` and `execution_context._context_manager._context_var` ContextVars, and drains leaked asyncio tasks.
2. **`test_trigger_param_binding.py:50-75`** — `_restore_global_agent_state()` snapshots/restores `AgentFieldLogger._cp_client` (via `get_logger()._cp_client` / `set_cp_client()`) and `Agent._current_agent`.
3. **`test_logger.py:14-56`** — `reset_logger_state()` clears `_logger_cache` / `_global_log_level` AND snapshots/restores the stdlib `logging.root.manager.loggerDict`.
4. **`test_memory_client_core.py:33-35`** — `mute_debug_logs()` monkeypatches `log_debug` to a no-op.

Each fixture is file-scoped (`autouse=True` in a test file), so other test files don't benefit from the isolation. A test author writing a new test module that uses logger or execution context must either discover and add a similar fixture or risk order-dependent failures (`test_cancel.py:147-152` explicitly documents this hazard). The `test_logger.py` fixture at lines 32-56 exists *because* the stdlib logger state leaks — it's compensating for a design gap rather than following a deliberate pattern.

A centralized `reset_all_sdk_state()` fixture in `conftest.py` would give every test module baseline isolation without requiring each author to understand which internal state each SDK module touches. Individual fixtures could still exist for module-specific fine-grained control.

---

> Step 1: `conftest.py:71` defines `_isolate_execution_context()` — resets ContextVars, drains asyncio tasks.
> Step 2: `test_trigger_param_binding.py:50` defines `_restore_global_agent_state()` — resets `_cp_client` and `Agent._current_agent`.
> Step 3: `test_logger.py:14` defines `reset_logger_state()` — resets `_logger_cache`, `_global_log_level`, and stdlib logging state.
> Step 4: `test_memory_client_core.py:33` defines `mute_debug_logs()` — monkeypatches `log_debug`.
> Step 5: Each fixture is file-scoped. No `conftest.py` fixture covers logger module state or stdlib logging state.
> Step 6: `test_cancel.py:147-152` documents that creating an `AgentFieldLogger` in one test breaks `caplog` in later tests — the hazard these fixtures exist to prevent.

**💡 Suggested Fix**

Consider a centralized `reset_all_sdk_state()` autouse fixture in `conftest.py` that aggregates the common reset operations (log-level reset, logger cache clear, ContextVar reset, asyncio task drain). Individual module fixtures can remain for module-specific state but all test files get baseline protection without duplication. This would also make the reset sequence explicit and auditable in one place rather than scattered across files. If a new SDK module adds global state, the centralized fixture is the obvious place to add its reset — no discovery burden on test authors.

---
*`Test Fixture Isolation: Private Internals & Fragmentation` · confidence 80%*

AbirAbbas (commented):
## 🟡 PR-AF Review — **Minor Issues**

> **7 findings** · 🔴 0 critical · 🟠 3 important · 🔵 4 suggestions · ⚪ 0 nitpicks

<details>

## Summary

- Fixed  to return loggers with correct names instead of always returning 
- Implemented dictionary-based cache () to store loggers by their name
- Maintained shared configuration via  for consistent log level management

## Changes

- : Replaced global singleton pattern with dictionary-based caching
- : Updated logger imports
- : Added unit tests for new caching behavior

## Test plan

- All 5 acceptance criteria verified passing
- Unit tests in  pass
- Each unique logger name produces a logger with that name
-  properly propagates to all cached and new loggers

🤖 Built with [AgentField SWE-AF](https://github.com/Agent-Field/SWE-AF)
🔌 Powered by [AgentField](https://github.com/Agent-Field/agentfield)

</details>

### Key Findings

**3 issue(s) should be addressed before merge:**

- 🟠 **Cross-test-suite propagate state leakage — reset_logger_state fixture exists only in test_logger.py, not in shared conftest.py** (`sdk/python/tests/test_logger.py:14`) — The `reset_logger_state` autouse fixture in `test_logger.py:15-56` snapshots and restores `logging.Logger.propagate` flags and handlers across every test in **this module only**.
- 🟠 **verification.py logger missed by set_cp_client causes silent structured log loss when local_verification=True** (`sdk/python/agentfield/logger.py:504`) — **Root cause**: `set_cp_client()` at `logger.py:504-509` snapshots currently-cached loggers and sets `_cp_client` as an instance attribute on each.
- 🟠 **Unbounded logger cache entries compound propagate=False damage — each dynamic-name AgentFieldLogger permanently silences its stdlib namespace branch from root handlers** (`sdk/python/agentfield/logger.py:82`) — Two independently-identified issues combine into cumulative semantic degradation of the stdlib logging hierarchy that grows over the process lifetime:  **Finding A**: `_logger_cache` (logger.py:463) i…

**4 suggestion(s) and style note(s):**

- 🔵 `reset_logger_state` fixture directly mutates private `_logger_cache` and `_global_log_level` module internals (`sdk/python/tests/test_logger.py:39`)
- 🔵 Cache has no size limit or eviction policy — currently bounded but API invites unbounded growth (`sdk/python/agentfield/logger.py:463`)
- 🔵 Fragmentation of test isolation patterns across modules — no unified global-state reset (`sdk/python/tests/test_logger.py:14`)
- 🔵 base_logger fixture must manually set propagate=True after construction — workaround for wrong default, not documentation (`sdk/python/tests/test_execution_logger.py:158`)

**Files with findings:** `sdk/python/agentfield/logger.py`, `sdk/python/tests/test_execution_logger.py`, `sdk/python/tests/test_logger.py`

<details>

#### 🟠 Important (3)

- **Cross-test-suite propagate state leakage — reset_logger_state fixture exists only in test_logger.py, not in shared conftest.py** `sdk/python/tests/test_logger.py:14`
- **verification.py logger missed by set_cp_client causes silent structured log loss when local_verification=True** `sdk/python/agentfield/logger.py:504`
- **Unbounded logger cache entries compound propagate=False damage — each dynamic-name AgentFieldLogger permanently silences its stdlib namespace branch from root handlers** `sdk/python/agentfield/logger.py:82`

#### 🔵 Suggestion (4)

- **`reset_logger_state` fixture directly mutates private `_logger_cache` and `_global_log_level` module internals** `sdk/python/tests/test_logger.py:39`
- **Cache has no size limit or eviction policy — currently bounded but API invites unbounded growth** `sdk/python/agentfield/logger.py:463`
- **Fragmentation of test isolation patterns across modules — no unified global-state reset** `sdk/python/tests/test_logger.py:14`
- **base_logger fixture must manually set propagate=True after construction — workaround for wrong default, not documentation** `sdk/python/tests/test_execution_logger.py:158`

</details>

<details>

**Dimensions Analyzed (4):**

- **Propagation semantics: does AgentFieldLogger.propagate=False on every named child break downstream log consumers?** — 2 file(s)
- **set_cp_client initialization ordering: do loggers created after set_cp_client silently lose structured log dispatch?** — 2 file(s)
- **_logger_cache has no eviction policy — unbounded growth with dynamic names** — 1 file(s)
- **Test isolation coupling to private logger internals** — 1 file(s)

- **Semantic** — 3 dimension(s), 88% coverage confidence
- **Mechanical** — 3 dimension(s), 85% coverage confidence
- **Systemic** — 2 dimension(s), 75% coverage confidence

- **3** compound finding(s) synthesized
- **5** finding(s) adversarially tested: 5 confirmed, 0 challenged

</details>

<details>

| Metric | Value |
|--------|-------|
| Duration | 17760.3s |
| Agent invocations | 18 |
| Coverage iterations | 0 |
| Budget exhausted | Yes (timeout: 17760s > 1800s limit) |
| PR type | bugfix |
| Complexity | standard |

</details>

Review ID: `rev_38fa8f7f6512`

<br>
<div align="right">
    AgentField PR-AF
  </a>
</div>

sdk/python/agentfield/logger.py
```diff
@@ -454,28 +459,54 @@ def log_execution(
...
 
 
 def set_cp_client(client: Optional["AgentFieldClient"]) -> None:
-    """Attach a control-plane client so structured logs are forwarded."""
-    get_logger()._cp_client = client
+    """Attach a control-plane client so structured logs are forwarded to all loggers."""
+    with _logger_cache_lock:
+        loggers = list(_logger_cache.values())
+    for logger in loggers:
+        logger._cp_client = client
```
santoshkumarradha, line 529: This changes the cache shape but it drops the old global `set_cp_client()` contract for any logger created after Agent installs the control-plane client. I reproduced it locally on this branch with `set_cp_client(client); get_logger("agentfield.cancel")._cp_client is client` and it returns `False`, so named loggers silently stop forwarding structured logs. We need to persist the client as shared logger state, then apply it both to cached loggers and to any new logger created later.

santoshkumarradha (changes requested):
Thanks for pushing this through. The thread-safety fixes and the test cleanup helped, but there is still one behavior regression before this is safe to merge: after `Agent` calls `set_cp_client(self.client)`, any named logger created later no longer inherits that client on this branch.

I reproduced it locally with the current head by calling `set_cp_client(client)` and then `get_logger("agentfield.cancel")`, and the new logger comes back with `_cp_client is None`. That means structured logs from non-default named loggers can silently stop forwarding to the control plane.

Once the control-plane client is stored as shared logger state and applied to future loggers as well as cached ones, I can recheck this quickly.

sdk/python/agentfield/logger.py
AbirAbbas: Fixed in ef6d7354. Persisted the client as shared logger state (module-level `_global_cp_client`, set under the cache lock in `set_cp_client`) and apply it to new loggers in `get_logger()` — mirroring the existing `_global_log_level` pattern. Your exact repro now passes: `set_cp_client(client); get_logger("agentfield.cancel")._cp_client is client` → `True`, and `set_cp_client(None)` detaches globally. Added regression coverage in `test_logger.py` (`test_logger_created_after_set_cp_client_forwards_to_cp`, plus the already-cached and None-reset paths).

sdk/python/agentfield/logger.py
AbirAbbas: Confirmed and fixed in ef6d7354. Root cause exactly as described — the asymmetry between `set_log_level` (which had `_global_log_level` applied in `get_logger`) and `set_cp_client` (which had no module-level equivalent). Took Option A: added `_global_cp_client`, set it under the cache lock, and apply it to loggers created later. The `agentfield.verification` logger (lazily imported in `Agent.__init__` after `set_cp_client`) now forwards correctly under `local_verification=True`. Regression test added.

sdk/python/tests/test_logger.py
AbirAbbas: The fixture already snapshots and restores the stdlib `logging` registry (propagate flags + handlers) on teardown, so it does not leak `propagate=False` state into other test files — `test_cancel.py` is order-independent because of that restore, not because it switched to a direct handler. Extended it here to also reset the new `_global_cp_client`. Centralizing a `reset_all_sdk_state()` in `conftest.py` is a reasonable hygiene improvement but is broader than this fix and orthogonal to the bug; leaving it out of scope here.

sdk/python/agentfield/logger.py
AbirAbbas: Out of scope for this fix. As the analysis itself notes, the cache is bounded in practice — all callers use static `__name__`, so there is no unbounded-growth or progressive-propagation path today. `propagate=False` is pre-existing intended behavior (the SDK owns its own stdout handler), not introduced here. If we ever expose dynamic logger names we can add an LRU bound + scope `propagate=False` to the root `agentfield` logger then; not warranted now.

sdk/python/tests/test_logger.py
AbirAbbas: Acknowledged — the test reaches into `_logger_cache` / `_global_log_level` / `_global_cp_client` by design, accessed via `logger_module` so the reset hits the real globals. A public `reset_logger_state()` on the module would be cleaner but is a test-ergonomics improvement, not part of this bug fix. Leaving as-is for now.

sdk/python/agentfield/logger.py
AbirAbbas: Same disposition as the compound-analysis thread: theoretical given all callers use static module names (the comment concedes this). No dynamic-name caller exists in production or examples, so no eviction policy is needed for this change. Can revisit if/when dynamic names are introduced.

sdk/python/tests/test_logger.py
AbirAbbas: Fair observation. A centralized `reset_all_sdk_state()` in `conftest.py` is a sensible follow-up, but consolidating the four module-scoped fixtures is a test-architecture change independent of this bug fix — out of scope here.

PR #637, opened 2026-06-08
fix(harness): preserve original goal on non-crash schema retries

### LINKED ISSUE #636: Harness schema-retry drops the original goal on non-crash retries (agents retry blind)
## Summary

`HarnessRunner._handle_schema_with_retry` (`sdk/python/agentfield/harness/_runner.py`) drops the original goal/task on **non-crash** schema retries. When a reasoner's first attempt fails schema validation in a non-crash way (output file written-but-invalid, or not written without a crash), the retry prompt is built from `build_followup_prompt(error_detail, cwd, schema)` alone — which contains only the schema-correction text ("PREVIOUS ATTEMPT FAILED… here is the schema… write the file") and **not the original goal**.

The crash branch already preserves the goal via `options["_original_prompt"]`; the non-crash branch does not.

```python
# _runner.py, in the retry loop:
if is_crash:
    original_prompt = options.get("_original_prompt", "")
    retry_prompt = original_prompt if original_prompt else build_followup_prompt(...)
else:
    error_detail = diagnose_output_failure(output_path, schema)
    retry_prompt = build_followup_prompt(error_detail, cwd, schema)   # <-- drops _original_prompt
```

`build_followup_prompt` (`_schema.py`) does not include the original prompt, and opencode's per-call fresh `XDG_DATA_HOME` means the `resume_session_id` set on the retry often can't actually restore the prior session — so the agent retries genuinely goal-blind.

## Impact

We hit this with a multi-agent build. A Product Manager reasoner's first attempt failed to write its output file; the retry, receiving only "previous attempt failed + schema", produced a **placeholder PRD** ("no product goal was supplied… awaiting user input"). That placeholder then propagated to every downstream reasoner (architect → sprint planner → merger), all of which planned against an empty goal. The whole run silently produced nothing useful, with no obvious error — the hardest kind of failure to trace.

## Fix

Prepend `options["_original_prompt"]` (already populated earlier in `run()`) to the followup on the non-crash branch, so retries carry **both** the original goal and the schema correction. A regression test is included: drive a retry whose first attempt writes schema-invalid output, and assert the retry prompt still contains the original goal.

PR incoming.

tha-hammer:
## What

Fixes the non-crash schema-retry path in `HarnessRunner._handle_schema_with_retry` so retries keep the original goal instead of sending only the schema-correction followup.

Closes #636.

## Why

When a reasoner's first attempt fails schema validation in a non-crash way (output file invalid, or not written without a crash), the retry prompt was built from `build_followup_prompt(error_detail, cwd, schema)` alone — which contains only "PREVIOUS ATTEMPT FAILED… here is the schema… write the file" and **not the original goal**. The crash branch already preserved the goal via `options["_original_prompt"]`; this branch didn't.

In practice the agent retried goal-blind and could emit placeholder output (we saw a PM produce a "no goal supplied" PRD) that then poisoned every downstream reasoner — a silent, hard-to-trace failure.

## Change

```python
else:
    error_detail = diagnose_output_failure(output_path, schema)
    _orig = options.get("_original_prompt", "")
    _followup = build_followup_prompt(error_detail, cwd, schema)
    retry_prompt = (_orig + "\n\n" + _followup) if _orig else _followup
```

`_original_prompt` is already populated earlier in `run()`, so this is a minimal, local change.

## Test

Adds `test_schema_retry_preserves_original_goal_in_prompt`: drives a non-crash schema-invalid first attempt and asserts the retry prompt sent to the provider still contains the original goal. Verified **red** before the fix (retry prompt was only `"PREVIOUS ATTEMPT FAILED…"`) and **green** after; the full `test_harness_runner.py` suite (15 tests) passes.

### DIFF
--- a/sdk/python/agentfield/harness/_runner.py
+++ b/sdk/python/agentfield/harness/_runner.py
@@ -429,7 +429,12 @@ class HarnessRunner:
                 )
             else:
                 error_detail = diagnose_output_failure(output_path, schema)
-                retry_prompt = build_followup_prompt(error_detail, cwd, schema)
+                _orig = options.get("_original_prompt", "")
+                _followup = build_followup_prompt(error_detail, cwd, schema)
+                # Keep the original goal/task on non-crash retries. Without this
+                # the agent retries with only the schema-correction text and loses
+                # the goal (e.g. PM emits a placeholder PRD that poisons the run).
+                retry_prompt = (_orig + "\n\n" + _followup) if _orig else _followup
 
             detail_for_log = diagnose_output_failure(output_path, schema)
 
--- a/sdk/python/tests/test_harness_runner.py
+++ b/sdk/python/tests/test_harness_runner.py
@@ -384,3 +384,44 @@ async def test_schema_retry_accumulates_cost(tmp_path, monkeypatch):
     assert result.parsed.name == "ok"
     assert result.cost_usd == pytest.approx(0.03)  # 0.01 + 0.02
     assert result.num_turns == 2
+
+
+@pytest.mark.asyncio
+async def test_schema_retry_preserves_original_goal_in_prompt(tmp_path, monkeypatch):
+    """Regression for the data-flow bug: on a non-crash schema retry the retry
+    prompt MUST still carry the original goal, not only the schema-correction
+    followup. Previously the PM lost its goal on retry and emitted a placeholder
+    PRD that poisoned every downstream reasoner (architect/sprint/merger)."""
+    GOAL = "IMPLEMENT_24_ENDPOINTS_SENTINEL_GOAL"
+    prompts: list[str] = []
+
+    class BadJsonCapturingProvider(MockProvider):
+        async def execute(self, prompt, options):
+            prompts.append(prompt)
+            output_path = get_output_path(str(options.get("cwd", ".")))
+            with open(output_path, "w", encoding="utf-8") as file_obj:
+                file_obj.write('{"name": "ok"}')  # missing 'count' -> invalid, non-crash
+            return await super().execute(prompt, options)
+
+    async def _no_repair(*args, **kwargs):
+        return None
+
+    monkeypatch.setattr("agentfield.harness._runner._ai_schema_repair", _no_repair)
+
+    provider = BadJsonCapturingProvider()
+    runner = HarnessRunner()
+    with patch("agentfield.harness._runner.build_provider", return_value=provider):
+        await runner.run(
+            GOAL,
+            provider="codex",
+            schema=DemoSchema,
+            cwd=str(tmp_path),
+            schema_max_retries=2,
+        )
+
+    assert len(prompts) >= 2, f"expected a retry; got {len(prompts)} call(s)"
+    retry_prompt = prompts[1]
+    assert GOAL in retry_prompt, (
+        "retry prompt dropped the original goal -- the agent retries blind. "
+        "First 400 chars:\n" + retry_prompt[:400]
+    )


sdk/python/agentfield/harness/_runner.py
```diff
@@ -429,7 +429,12 @@ async def _handle_schema_with_retry(
                 )
             else:
                 error_detail = diagnose_output_failure(output_path, schema)
```
AbirAbbas, line 431: 🔵 **[SUGGESTION] Non-crash schema retry prompt inlines the JSON schema twice for small schemas**

For schemas under `LARGE_SCHEMA_TOKEN_THRESHOLD` (4000 tokens, `_schema.py:26`), the non-crash retry prompt at `_runner.py:431-437` contains the full JSON schema **twice** — once embedded in `_original_prompt` (which was built at line 246 via `prompt + build_prompt_suffix(schema, output_dir)` and inlines the schema at `_schema.py:91-98`) and once embedded in `_followup` (which calls `build_followup_prompt` and inlines the schema at `_schema.py:341`).

This wastes context-window tokens proportional to the schema size. For a schema near the 4000-token threshold, the combined retry prompt carries ~8000 tokens of schema content alone, plus wrapping text and the original user prompt. When session resume is active (line 449), the provider additionally replays the conversation history containing the original prompt with the inlined schema, making this effectively a **triple** presentation.

The wasted tokens increase the risk of context overflow or output truncation for providers with tighter context windows, and the duplicate schema could be misinterpreted by the model as a variant requirement or revision instruction rather than a redundant restatement.

This duplication affects only the **small-schema** path. Large schemas (>4000 tokens) are not affected because both `build_prompt_suffix` (line 79-89) and `build_followup_prompt` (line 328-339) reference the schema file path instead of inlining.

---

> **Step 1** (`_runner.py:244-247`): `effective_prompt = prompt + build_prompt_suffix(schema, output_dir)` — for a small schema, `build_prompt_suffix` returns text containing `"Required JSON Schema:\n{schema_json}\n\n"` (per `_schema.py:91-98`). `_original_prompt` is set to this value.
> 
> **Step 2** (`_runner.py:430-437`): On non-crash retry, `_orig = options.get("_original_prompt", "")` retrieves the prompt **with** the inlined schema. `_followup = build_followup_prompt(error_detail, cwd, schema)` calls `_schema.py:341` which appends `"The JSON MUST conform to this schema:\n{schema_json}\n\n"`.
> 
> **Step 3** (`_runner.py:437`): `retry_prompt = (_orig + "
> 
> " + _followup)` — the concatenated prompt contains `schema_json` from `_orig` **and** `schema_json` from `_followup`, i.e., the schema JSON string appears at two distinct locations within the same message.
> 
> **Step 4** (`_runner.py:449`): `retry_options["resume_session_id"] = last_session_id` — the provider replays the original conversation, which already contains the full `_original_prompt` (with schema) as the first user message. The model now sees the schema in the resumed history + in the retry's prefixed `_orig` + in the retry's `_followup`.

**💡 Suggested Fix**

The simplest fix: before concatenating `_orig` with `_followup`, strip the schema suffix from `_orig`. `build_prompt_suffix` always starts with `\n\n---\nCRITICAL OUTPUT REQUIREMENTS:` for small schemas — locate that delimiter and truncate. This preserves the original user goal while letting the followup carry the schema-correction context.

Alternative: store the raw user `prompt` (before `build_prompt_suffix` is applied) in a separate key like `options["_original_goal"]` at line 244, and use that for the retry prefix instead of the full `_original_prompt`. This separates concerns: the goal text from the schema/suffix text.

```python
# At line 244-247:
options["_original_goal"] = prompt
if schema is not None:
    effective_prompt = prompt + build_prompt_suffix(schema, output_dir)
options["_original_prompt"] = effective_prompt

# At line 432:
_orig_goal = options.get("_original_goal", "")
retry_prompt = (_orig_goal + "

" + _followup) if _orig_goal else _followup
```

---
*`Schema duplication in retry prompt for small schemas` · confidence 95%*

sdk/python/tests/test_harness_runner.py
```diff
@@ -384,3 +384,44 @@ async def execute(self, prompt: str, options: dict) -> RawResult:
...
+            return await super().execute(prompt, options)
+
+    async def _no_repair(*args, **kwargs):
+        return None
+
+    monkeypatch.setattr("agentfield.harness._runner._ai_schema_repair", _no_repair)
+
+    provider = BadJsonCapturingProvider()
+    runner = HarnessRunner()
+    with patch("agentfield.harness._runner.build_provider", return_value=provider):
```
AbirAbbas, line 413: 🔵 **[SUGGESTION] Test ignores HarnessResult return value — no assertion on eventual failure mode**

The test calls `await runner.run(...)` at line 414 but discards the returned `HarnessResult`. The `BadJsonCapturingProvider` always writes `{"name": "ok"}` (missing the required `count` field) on every invocation — including all retries. After exhausting `schema_max_retries=2`, the runner should return an error `HarnessResult` with `is_error=True` and `failure_type=FailureType.SCHEMA`.

Not asserting the return value means the test won't catch regressions where the runner silently accepts invalid output after retries (e.g., if a parse bug causes `validated` to be non-None when it shouldn't be). The existing test `test_schema_validation_failure_returns_error_result` (line 217) covers the no-retry case, but this test exercises the retry-exhausted case and should verify the error propagation through the retry loop.

---

> Step 1: `BadJsonCapturingProvider.execute()` at line 402-403 writes `'{"name": "ok"}'` to the output file — always missing the required `count` field.
> Step 2: `_handle_schema_with_retry` loop at `_runner.py:414-491` will call this provider on every retry attempt.
> Step 3: With `schema_max_retries=2` and `_ai_schema_repair` patched to return None, all `parse_and_validate` calls will return None.
> Step 4: The loop exhausts all retries and falls through to `_runner.py:493-510`, returning `HarnessResult(is_error=True, failure_type=FailureType.SCHEMA)`.
> Step 5: The test at line 414 does `await runner.run(...)` without capturing the return value — the error result is discarded.

**💡 Suggested Fix**

Capture and assert the return value:
```python
result = await runner.run(
    GOAL,
    provider="codex",
    schema=DemoSchema,
    cwd=str(tmp_path),
    schema_max_retries=2,
)
assert result.is_error is True, f"expected error after all retries exhausted; got {result}"
```
This provides defense-in-depth: if a future change causes the retry loop to incorrectly accept invalid output, this assertion would catch it.

---
*`Coverage Gap Review - Cluster sdk/python/tests` · confidence 70%*

sdk/python/agentfield/harness/_runner.py
```diff
@@ -429,7 +429,12 @@ async def _handle_schema_with_retry(
                 )
             else:
```
AbirAbbas, line 430: 🔵 **[SUGGESTION] Claude Code retry sends original goal redundantly when resuming session**

When the Claude Code provider performs a non-crash schema retry, the model receives the original task goal **twice** — once in the resumed conversation history and once in the new retry prompt. This occurs because two independent mechanisms in `_handle_schema_with_retry` both preserve the original prompt without coordinating:

1. **`resume_session_id`** (line 450): Resumes the prior conversation, which already contains the full `effective_prompt` (goal + schema suffix) from the first attempt.
2. **`_orig` prefix** (lines 432-437): Prepends `_original_prompt` (identical to `effective_prompt`) to the retry prompt.

Since the Claude SDK's `query(prompt=..., options={resume: session_id})` **appends** the new prompt as a fresh user message on top of the loaded conversation history (confirmed from upstream `claude_agent_sdk` source), the combined context is:

```
[System prompt]
User: <original goal + schema suffix>           ← from resumed history
Assistant: <first attempt response>             ← from resumed history
User: <original goal + schema suffix AGAIN>     ← from _orig prefix
      + followup (schema correction)            ← from build_followup_prompt
```

The redundant goal wastes context window tokens that could be used for the correction task itself. For schemas near the `LARGE_SCHEMA_TOKEN_THRESHOLD` (4000 tokens), the wasted space is substantial. While the Claude SDK performs automatic compaction when approaching context limits, this wasted space reduces the model's effective working memory for making the correction.

Additionally, the model could misinterpret the repeated task instructions as a new or separate request rather than a correction of the prior attempt, although the `"PREVIOUS ATTEMPT FAILED"` prefix in the followup mitigates this risk.

---

> Step 1: `_runner.py:246` — `effective_prompt = prompt + build_prompt_suffix(schema, output_dir)` builds the full original prompt with goal + schema suffix.
> Step 2: `_runner.py:247` — `options["_original_prompt"] = effective_prompt` stores it for retry use.
> Step 3: `_runner.py:251-253` — First `_execute_with_retry(provider_instance, effective_prompt, options)` sends `effective_prompt` to Claude SDK, which stores it in conversation history.
> Step 4: `_runner.py:449-450` — `retry_options["resume_session_id"] = last_session_id` enables session resume.
> Step 5: `_runner.py:432` — `_orig = options.get("_original_prompt", "")` retrieves the full `effective_prompt`.
> Step 6: `_runner.py:437` — `retry_prompt = (_orig + "\n\n" + _followup)` prepends the full original prompt to the correction followup.
> Step 7: `_runner.py:452-454` — `_execute_with_retry(provider, retry_prompt, retry_options)` calls Claude SDK.
> Step 8: `claude.py:60-62` — `agent_options["resume"] = str(resume_sid)` passes resume to SDK.
> Step 9: Claude SDK loads full conversation history (including `effective_prompt` from Step 3) and appends `retry_prompt` (which includes `effective_prompt` again from Step 6) as a new user message. Model sees duplicate goal.

**💡 Suggested Fix**

When `resume_session_id` is set (i.e., the session is being resumed), omit `_original_prompt` from the retry prompt since the resumed conversation already contains it. The retry prompt should be only the followup/correction message. When `resume_session_id` is NOT set (other providers, or session ID unavailable), keep the current behavior of prepending `_original_prompt`.

Example fix at lines 430-437:
```python
else:
    error_detail = diagnose_output_failure(output_path, schema)
    if last_session_id:
        # Session will be resumed; goal is already in history.
        retry_prompt = build_followup_prompt(error_detail, cwd, schema)
    else:
        _orig = options.get("_original_prompt", "")
        _followup = build_followup_prompt(error_detail, cwd, schema)
        retry_prompt = (_orig + "\n\n" + _followup) if _orig else _followup
```

This preserves the goal-preservation fix from this PR for non-resuming providers while eliminating the redundant duplication for Claude.

---
*`Claude resume duplicate goal in retry` · confidence 90%*

sdk/python/tests/test_harness_runner.py
```diff
@@ -384,3 +384,44 @@ async def execute(self, prompt: str, options: dict) -> RawResult:
     assert result.parsed.name == "ok"
     assert result.cost_usd == pytest.approx(0.03)  # 0.01 + 0.02
     assert result.num_turns == 2
+
+
+@pytest.mark.asyncio
```
AbirAbbas, line 389: 🔵 **[SUGGESTION] Test verifies goal preservation but does not verify followup/schema-correction feedback is retained**

The test assert at line 424 checks only `GOAL in retry_prompt`. This catches the original bug (retry prompt lacked the goal entirely) but does **not** verify that the schema-correction followup (from `build_followup_prompt()`) is also present in the retry prompt.

The crash branch (`_runner.py:422-429`) uses `_original_prompt` as a **replacement** for `build_followup_prompt` — i.e. when `_original_prompt` is set, the followup is discarded and only the original prompt is sent. The non-crash branch (this PR's fix) uses `_original_prompt` as a **prefix** before the followup. If a future contributor unifies the branches by applying the crash-branch pattern (`retry_prompt = _orig if _orig else build_followup_prompt(...)`) to the non-crash path, the test would **still pass** because `_original_prompt` always contains the goal (set unconditionally at `_runner.py:247`). But the schema-correction error detail would be silently dropped, reverting to a different form of goal-blind retry.

The test should additionally assert that the retry prompt contains distinctive followup text (e.g. `"PREVIOUS ATTEMPT FAILED"`) to verify the full contract: goal + schema-correction feedback.

---

> Step 1: `_runner.py:247` unconditionally sets `options["_original_prompt"] = effective_prompt` which always contains the goal.
> Step 2: `_runner.py:422-429` (crash branch) uses `_original_prompt` as a REPLACEMENT: `retry_prompt = original_prompt if original_prompt else build_followup_prompt(...)` — the followup is discarded when `_orig` is set.
> Step 3: `_runner.py:430-437` (non-crash branch) uses `_original_prompt` as a PREFIX: `retry_prompt = (_orig + "\n\n" + _followup) if _orig else _followup` — the followup is appended.
> Step 4: The test at `test_harness_runner.py:424` only asserts `GOAL in retry_prompt`.
> Step 5: Since `_original_prompt` always contains the goal, applying the crash-branch pattern to the non-crash branch would make the retry prompt = `_original_prompt` alone (goal present, followup dropped) and the test would still pass.

**💡 Suggested Fix**

Add a second assertion verifying the followup text is present in the retry prompt:
```python
assert "PREVIOUS ATTEMPT FAILED" in retry_prompt, (
    "retry prompt lost the schema-correction followup text. "
    "First 400 chars:\n" + retry_prompt[:400]
)
```
This ensures both components of the fix are verified: the original goal is preserved AND the followup feedback is appended, not replaced.

---
*`Coverage Gap Review - Cluster sdk/python/tests` · confidence 85%*

AbirAbbas (commented):
## 🟡 PR-AF Review — **Minor Issues**

> **10 findings** · 🔴 0 critical · 🟠 2 important · 🔵 8 suggestions · ⚪ 0 nitpicks

<details>

## What

Fixes the non-crash schema-retry path in `HarnessRunner._handle_schema_with_retry` so retries keep the original goal instead of sending only the schema-correction followup.

Closes #636.

## Why

When a reasoner's first attempt fails schema validation in a non-crash way (output file invalid, or not written without a crash), the retry prompt was built from `build_followup_prompt(error_detail, cwd, schema)` alone — which contains only "PREVIOUS ATTEMPT FAILED… here is the schema… write the file" and **not the original goal**. The crash branch already preserved the goal via `options["_original_prompt"]`; this branch didn't.

In practice the agent retried goal-blind and could emit placeholder output (we saw a PM produce a "no goal supplied" PRD) that then poisoned every downstream reasoner — a silent, hard-to-trace failure.

## Change

```python
else:
    error_detail = diagnose_output_failure(output_path, schema)
    _orig = options.get("_original_prompt", "")
    _followup = build_followup_prompt(error_detail, cwd, schema)
    retry_prompt = (_orig + "\n\n" + _followup) if _orig else _followup
```

`_original_prompt` is already populated earlier in `run()`, so this is a minimal, local change.

## Test

Adds `test_schema_retry_preserves_original_goal_in_prompt`: drives a non-crash schema-invalid first attempt and asserts the retry prompt sent to the provider still contains the original goal. Verified **red** before the fix (retry prompt was only `"PREVIOUS ATTEMPT FAILED…"`) and **green** after; the full `test_harness_runner.py` suite (15 tests) passes.

</details>

### Key Findings

**2 issue(s) should be addressed before merge:**

- 🟠 **Three uncoordinated prompt-composition mechanisms combine for multiplicative token amplification on non-crash retries** (`sdk/python/agentfield/harness/_runner.py:414`) — The non-crash schema retry loop assembles its prompt from three independently-designed mechanisms that each carry the original goal text and/or the JSON schema, but none is aware of what the others co…
- 🟠 **Untested crash retry path silently resends the crashing prompt with no error diagnostic, risking deterministic re-crashes** (`sdk/python/agentfield/harness/_runner.py:418`) — The crash retry branch (lines 421-429) sends _original_prompt as the retry prompt when is_crash is True.

**8 suggestion(s) and style note(s):**

- 🔵 Non-crash schema retry prompt inlines the JSON schema twice for small schemas (`sdk/python/agentfield/harness/_runner.py:431`)
- 🔵 No test drives the crash retry path — zero coverage for crash branch prompt content (`sdk/python/agentfield/harness/_runner.py:418`)
- 🔵 Test ignores HarnessResult return value — no assertion on eventual failure mode (`sdk/python/tests/test_harness_runner.py:413`)
- 🔵 Unreachable `else` fallback clauses in both retry branches are misleading (`sdk/python/agentfield/harness/_runner.py:426`)
- 🔵 Claude Code retry sends original goal redundantly when resuming session (`sdk/python/agentfield/harness/_runner.py:430`)
- … and 3 more (see All Findings by Severity)

**Files with findings:** `sdk/python/agentfield/harness/_runner.py`, `sdk/python/tests/test_harness_runner.py`

<details>

#### 🟠 Important (2)

- **Three uncoordinated prompt-composition mechanisms combine for multiplicative token amplification on non-crash retries** `sdk/python/agentfield/harness/_runner.py:414`
- **Untested crash retry path silently resends the crashing prompt with no error diagnostic, risking deterministic re-crashes** `sdk/python/agentfield/harness/_runner.py:418`

#### 🔵 Suggestion (8)

- **Non-crash schema retry prompt inlines the JSON schema twice for small schemas** `sdk/python/agentfield/harness/_runner.py:431`
- **No test drives the crash retry path — zero coverage for crash branch prompt content** `sdk/python/agentfield/harness/_runner.py:418`
- **Test ignores HarnessResult return value — no assertion on eventual failure mode** `sdk/python/tests/test_harness_runner.py:413`
- **Unreachable `else` fallback clauses in both retry branches are misleading** `sdk/python/agentfield/harness/_runner.py:426`
- **Claude Code retry sends original goal redundantly when resuming session** `sdk/python/agentfield/harness/_runner.py:430`
- **Crash retry provides no error diagnostic feedback — model re-runs blind to the crash** `sdk/python/agentfield/harness/_runner.py:421`
- **`resume_session_id` set for all providers but only Claude reads it** `sdk/python/agentfield/harness/_runner.py:448`
- **Test verifies goal preservation but does not verify followup/schema-correction feedback is retained** `sdk/python/tests/test_harness_runner.py:389`

</details>

<details>

**Dimensions Analyzed (3):**

- **Schema content duplication in non-crash retry prompt** — 2 file(s)
- **Session resume presents original goal redundantly in Claude provider** — 2 file(s)
- **Crash and non-crash branches compose `_original_prompt` asymmetrically** — 2 file(s)

- **Semantic** — 3 dimension(s), 80% coverage confidence
- **Mechanical** — 2 dimension(s), 85% coverage confidence
- **Systemic** — 0 dimension(s), 85% coverage confidence

- **2** compound finding(s) synthesized
- **10** finding(s) adversarially tested: 8 confirmed, 2 challenged

</details>

<details>

| Metric | Value |
|--------|-------|
| Duration | 9537.3s |
| Agent invocations | 21 |
| Coverage iterations | 1 |
| Budget exhausted | Yes (timeout: 9537s > 1800s limit) |
| PR type | bugfix |
| Complexity | standard |

</details>

Review ID: `rev_21b18377d6a7`

<br>
<div align="right">
    AgentField PR-AF
  </a>
</div>

santoshkumarradha (approved):
This fix is on the right seam. The non-crash schema retry now preserves the original goal instead of retrying with only the schema-repair text, and the regression test covers that exact failure mode. I also reran the previously red Python 3.12 job and it came back green.

sdk/python/tests/test_harness_runner.py
santoshkumarradha: Fair hardening idea, but not a bug in this patch. The regression we needed here was whether the retry prompt kept the original goal, and the existing suite already covers schema-failure result handling elsewhere.

sdk/python/agentfield/harness/_runner.py
santoshkumarradha: Not blocking on this. Claude resume may duplicate context, but other providers need the explicit goal in the retry prompt, and this change is fixing that correctness issue first. Provider-specific prompt deduping can be a follow-up if needed.

sdk/python/agentfield/harness/_runner.py
santoshkumarradha: Not treating this as a blocker for this fix. The duplication concern is real in the abstract, but this PR is correcting a concrete retry bug and needs to preserve the goal for providers that do not consume `resume_session_id`. We can optimize prompt composition separately if it shows up as a real issue.

sdk/python/tests/test_harness_runner.py
santoshkumarradha: Reasonable extra assertion, but not required to merge this fix. The important regression here was goal loss on non-crash retries, and this test covers that seam directly.

PR #654, opened 2026-06-11
Add AgentField realtime sessions for WebRTC voice ingress

santoshkumarradha:
## Summary

This PR introduces the AgentField session DX for realtime voice and multimodal ingress through the control plane.

The product goal is that browsers, CLIs, and external clients do not connect directly to model providers. They create an AgentField session, and the control plane owns the provider boundary, realtime transport, tool calls, and workflow provenance.

## What Changed

- Adds Python `@app.session(...)` for declaring long-lived realtime/multimodal session handlers.
- Adds Python `RealtimeSession`, `SessionTurn`, `session.input()`, `session.say(...)`, and `session.call(...)` as the handler-facing DX.
- Adds TypeScript `agent.session(...)` and Go `RegisterSession(...)` equivalents so session declarations register consistently across SDKs.
- Adds explicit provider + transport validation in Python, TypeScript, Go, and the control plane.
- Adds control-plane session routes:
  - `POST /api/v1/sessions/:target/start`
  - `POST /api/v1/sessions/:session_id/realtime-offer`
  - `POST /api/v1/sessions/:session_id/tools/:tool`
- Adds OpenAI WebRTC SDP offer proxying through the control plane for `provider=openai`, `transport=webrtc`.
- Adds `af session` CLI commands for start, SDP offer exchange, tool invocation, and workflow lookup.
- Keeps session tool work on AgentField `execute/async`, forwarding `X-Session-ID` so resulting reasoner work belongs to the session and appears in the normal workflow DAG.
- Fixes Python nested `app.call` parent headers by forwarding the current execution as `X-Parent-Execution-ID`.
- Adds a TODO beside control-plane session relationship fields for future explicit session-to-session lifecycle edges.

## DX Shape

```python
@app.session(
    "voice",
    provider="openai",
    model="gpt-realtime-2",
    transport="webrtc",
    modalities=["audio", "text"],
    voice="marin",
    tools=["support.resolve_voice_turn"],
)
async def voice(session):
    turn = await session.input()
    result = await session.call("support.resolve_voice_turn", turn=turn)
    await session.say(result["spoken_response"])
```

Provider and transport are explicit controls:

- `provider` selects who runs the realtime/audio model.
- `transport` selects how the session moves audio/events through the control plane.
- AgentField validates the pair, but does not infer a transport or silently switch providers.

Example validation behavior:

```text
Unsupported session transport 'webrtc' for provider 'openrouter'. Supported transports: audio_turns. AgentField does not infer or switch providers; set provider and transport explicitly.
```

## CLI Shape

```bash
af session start voice-support-af.voice \
  --provider openai \
  --transport webrtc \
  --model gpt-realtime-2 \
  --voice marin

# Pipe-friendly SDP exchange through the control plane.
af session offer sess_123 \
  --provider openai \
  --transport webrtc \
  --sdp @offer.sdp > answer.sdp

# Session tools still route through AgentField execute/async.
af session tool sess_123 launch_support_workflow \
  --target voice-support-af.resolve_voice_turn \
  --in '{"utterance":"I need help with a delayed order"}'

af session workflows sess_123
```

`af session offer` accepts inline SDP, `--sdp @path`, or stdin. The default output is raw SDP so browser automation and WebRTC tooling can pipe it directly; `--output json` wraps the answer as `{ "answer_sdp": "..." }`.

## Product Positioning

Sessions make AgentField the ingress layer for realtime AI applications.

A browser voice call is not a side channel around AgentField. It becomes an AgentField session backed by the control plane. Browser WebRTC setup, provider negotiation, and tool calls all go through AgentField, while every agent action still routes through the existing execution API.

That gives realtime voice apps the same properties as normal AgentField workflows:

- provider boundary owned by the control plane
- explicit provider and transport configuration
- session-scoped execution context
- reasoner-to-reasoner workflow DAGs
- replay/provenance surfaces for agent work
- consistent SDK shape across Python, TypeScript, and Go

## Docs Follow-up

Website docs PR: Agent-Field/website2.0#28.

Merge this AgentField runtime/SDK/control-plane PR first. Merge the website docs PR after this lands so the public docs do not describe session APIs before they are available.

## Validation

- `uv run pytest tests/test_agent_session.py tests/test_session_transport.py tests/test_execution_context_core.py -q`: 13 passed.
- `npm test` from `sdk/typescript`: 619 passed.
- `npm run lint`: passed.
- `go test ./agent -run 'TestValidateSessionTransport|TestAgentRegisterSession|TestAgent'` from `sdk/go`: passed.
- `go test ./pkg/types ./internal/handlers ./internal/cli` from `control-plane`: passed.
- Earlier live voice spike validated browser WebRTC ingress through an AgentField-style session route into OpenAI Realtime, with session tool calls launching a control-plane workflow and preserving a 9-node AgentField DAG.

## Session Tool Semantics

`session.call(...)` and `tools=[...]` serve different paths:

- `session.call(...)` is handler-controlled orchestration. The session function can call AgentField reasoners or skills directly without listing them in `tools`.
- `tools=[...]` is the provider/client-visible allowlist for autonomous realtime tool calls during the live session. For example, a realtime audio model or browser-side tool bridge can choose to call `orders.lookup_order`; AgentField then routes that request through `/api/v1/sessions/:session_id/tools/:tool` into `execute/async` with the session ID attached.

So `tools` is not Python dependency injection and not required for normal handler code. It is the explicit exposure boundary for capabilities the live audio loop may invoke.


santoshkumarradha:
Merge order note: merge this AgentField runtime/SDK/control-plane PR first. The website docs PR Agent-Field/website2.0#28 documents the APIs introduced here and should merge only after this PR lands, so public docs do not get ahead of the shipped session surface.

santoshkumarradha:
Clarification on session `tools`: `session.call(...)` is available to the session handler for normal handler-controlled orchestration and does not require the target to be listed in `tools`.

The `tools=[...]` kwarg is the provider/client-visible allowlist for autonomous realtime tool calls during the live session. Those calls enter through `/api/v1/sessions/:session_id/tools/:tool` and are routed into `execute/async` with `X-Session-ID` so the workflow DAG remains session-scoped.

santoshkumarradha:
Merge order for realtime voice/session work:\n\n1. Merge this PR first: base realtime session SDK/control-plane/WebRTC support.\n2. Merge #655 second: session access tags, explicit session target/instance routes, CLI route updates, and UI surfaces.\n3. Merge/update website docs PR Agent-Field/website2.0#28 after the runtime PRs land, so docs reflect the final route names and access-control behavior.

AbirAbbas:
I did a deep dive on how the realtime-voice space actually wires this up (OpenAI Realtime docs, Azure GPT Realtime, LiteLLM, LiveKit, Twilio ConversationRelay) to sanity-check the design before merge. Short version: the core approach here is the *idiomatic, documented* one, and the security boundary is actually stronger than the common default. A few follow-ups worth tracking, none blocking.

**What's validated**

- The control-plane-proxies-SDP approach is exactly OpenAI's documented "unified interface": the browser POSTs its SDP to a developer server, the server combines it with session config in a multipart form and forwards to `/v1/realtime/calls` with a standard server key. OpenAI describes it as the *simpler* of the two options. `createOpenAIRealtimeCall` matches it almost line for line. The only documented tradeoff is that the app server sits in the critical path of session init — fine here.
- The "client never holds the provider key" boundary is more conservative than OpenAI's default ephemeral-token flow (where the browser does hold a short-lived `ek_...`). The unified interface keeps *all* provider credentials server-side. Azure documents the same backend-mediated variant and explicitly calls it "more secure"; LiteLLM ships the same SDP-proxy + key-boundary pattern in production.
- Keeping media direct (browser↔OpenAI) and only proxying signaling is correct — that's what every OpenAI WebRTC pattern does. Good that this doesn't try to relay audio through the control plane (that would need an SFU like LiveKit).
- Explicit provider/transport with no inference is a good call and consistent with how the mature platforms behave.

**The genuinely novel piece (worth calling out, not changing)**

Routing live tool-calls through `execute/async` for workflow/DAG provenance has no vendor precedent — OpenAI's documented analog is a server-side *sideband WebSocket* to the same session, not a separate execution API. That's fine: the DAG provenance is the whole point of AgentField. But it's the one part no provider validates, so it deserves to be explicit.

**Follow-ups (filed, non-blocking)**

- #662 — tool-calls currently rely on the client bridging the function-call event to `/tools/:tool`. OpenAI's pattern for keeping tool/business logic off the untrusted client is a server-side sideband WebSocket (`wss://.../realtime?call_id=rtc_xxxxx`). Without it the browser is in the tool path; `tools=[...]` as an allowlist mitigates but doesn't remove that.
- #663 — no turn-detection / server-VAD / barge-in config is exposed in `@app.session`; the session config sent to OpenAI is minimal. That's table-stakes for a usable voice UX.
- #664 — `websocket` transport is in the supported matrix but only `webrtc` is implemented, and `RealtimeSession.input()/say()` are still placeholders (input raises, say writes to an outbox). Tracking the end-to-end live handler loop so the scaffolding isn't mistaken for complete.

Net: the SDP-proxy + tool-routing surface is solid and matches what developers expect. I'm comfortable with this as the v1 ingress once it's smoke-tested locally — the rest is iteration.

AbirAbbas:
### Provider support — status + follow-ups

To set expectations for reviewers: the session DX here is provider-agnostic by design (`provider` is an explicit, required knob and the transport matrix is multi-provider), but **OpenAI WebRTC is the only fully-wired realtime path today**. `realtime-offer` hard-gates to `provider=openai`+`transport=webrtc` and posts to `api.openai.com/v1/realtime/calls`; the declared `openrouter: audio_turns` pair currently has no handler, so the matrix advertises a provider that doesn't do anything yet.

That's fine as a first cut — OpenAI is the first provider, not the only intended one — but worth tracking so it isn't mistaken for multi-provider support:

- **#666** — umbrella: grow sessions into a real multi-provider framework (Azure Realtime is a near drop-in; Gemini Live needs its own adapter).
- **#665** — implement real OpenRouter support + voice-capable models (replace the `audio_turns` no-op).
- **#664** — `websocket` transport is in the matrix but only `webrtc` is implemented; `RealtimeSession.input()/say()` are scaffolding.
- **#663** — expose turn-detection / server-VAD / barge-in in `@app.session`.
- **#662** — server-side sideband WebSocket so tool-calls don't depend on the browser bridging them.

### Added: runnable voice example

Pushed `examples/python_agent_nodes/voice_dictation/` (commit 388dab5e) — a Wispr-Flow-style live-dictation page that exercises the real flow end-to-end: browser mic → control-plane SDP proxy → OpenAI Realtime, key server-side, with the `tools=[...]` allowlist routing through `execute/async`.

Verified locally against this branch on WSL: control plane built and ran, all three session routes behave per spec (provider/transport validation, key boundary, agent lookup), `@app.session(...)` registers/serializes/rejects bad pairs, and the SDP proxy reaches OpenAI and gets a live answer. One thing the example surfaces: because session-level transcription/VAD config isn't exposed yet (#663), the page sends a client-side `session.update` — and it needs the **GA shape** (`session.type: "realtime"` + transcription/`turn_detection` nested under `audio.input`); the older flat `input_audio_transcription`/`turn_detection` keys get rejected with `Missing required parameter: 'session.type'`. Worth keeping in mind when #663 wires this server-side.

AbirAbbas (dismissed):

AbirAbbas (approved):

PR #656, opened 2026-06-11
Add Snowflake integration pack

santoshkumarradha:
## Summary

Adds the first Snowflake integration pack with two separate surfaces:

- a built-in control-plane `snowflake` trigger source for Snowflake event-table/custom-query polling
- an optional Docker-hostable Go capability node that registers Snowflake skills/reasoners for cross-node calls

This keeps ingress in the control plane while keeping data/API capabilities modular, deployable, and callable from any AgentField SDK.

### Trigger side

Operators configure Snowflake from **Integrations -> Snowflake -> Connect** or through `POST /api/v1/triggers`.

The target node/reasoner must already be registered. The trigger stores only a secret env-var reference, for example `SNOWFLAKE_PAT`; the PAT value stays in the control-plane environment.

Example config shape:

```json
{
  "mode": "event_table_poll",
  "account_url": "https://<account>.snowflakecomputing.com",
  "database": "AGENTFIELD_TEST",
  "schema": "PUBLIC",
  "table": "AGENTFIELD_EVENTS",
  "warehouse": "AGENTFIELD_WH",
  "role": "AGENTFIELD_TEST_ROLE",
  "interval_seconds": 30,
  "max_batch_size": 100
}
```

The source persists inbound event history, dedupes by `EVENT_ID`, normalizes payload/provenance, and dispatches to the selected reasoner.

### Node side

Users deploy the optional Snowflake node as a Go binary in Docker. It registers as a normal AgentField node, for example `snowflake-prod`.

```bash
docker run --name agentfield-snowflake \
  -p 8012:8012 \
  -e AGENTFIELD_SERVER=http://host.docker.internal:8080 \
  -e AGENTFIELD_NODE_ID=snowflake-prod \
  -e SNOWFLAKE_NODE_LISTEN=:8012 \
  -e SNOWFLAKE_NODE_PUBLIC_URL=http://localhost:8012 \
  -e SNOWFLAKE_ACCOUNT_URL=https://<account>.snowflakecomputing.com \
  -e SNOWFLAKE_PAT=<programmatic-access-token> \
  -e SNOWFLAKE_DATABASE=AGENTFIELD_TEST \
  -e SNOWFLAKE_SCHEMA=PUBLIC \
  -e SNOWFLAKE_WAREHOUSE=AGENTFIELD_WH \
  -e SNOWFLAKE_ROLE=AGENTFIELD_TEST_ROLE \
  agentfield-snowflake-node:latest
```

Other nodes then call it through AgentField:

```go
result, err := a.Call(ctx, "snowflake-prod.query_readonly", map[string]any{
  "sql": "SELECT EVENT_ID, EVENT_TYPE FROM AGENTFIELD_EVENTS LIMIT 10",
})
```

```python
result = await app.call("snowflake-prod.semantic_ask", input={
    "question": "What changed in revenue yesterday?"
})
```

## Included Capabilities

- `query_readonly`
- `describe_table`
- `search_columns`
- `cortex_analyst_message`
- `cortex_search_query`
- `cortex_complete`
- `semantic_ask`
- `explain_result`
- `investigate_metric_change`

Prompt-backed reasoners load prompts from `SNOWFLAKE_PROMPTS_FILE` or packaged defaults, so prompt behavior is not hardcoded into the binary.

## Implementation Notes

- Registers `control-plane/internal/sources/snowflake` through the built-in source registry.
- Adds Snowflake UI metadata, icon, and trigger-dialog defaults using existing UI components/design tokens.
- Adds `integrations/snowflake` contracts, manifest, Dockerfile, prompt defaults, docs, and e2e harnesses.
- SQL API support includes PAT auth header and `202 Accepted` polling.
- Trigger polling normalizes Snowflake timestamp watermarks using `TO_VARCHAR(..., 'YYYY-MM-DD"T"HH24:MI:SS.FF9')` and `TO_TIMESTAMP_NTZ(...)`; this was verified against a real Snowflake account after the first live smoke uncovered Snowflake JSON timestamp serialization behavior.

## Validation

- `go test ./internal/sources/snowflake ./internal/sources/all`
- `go test ./...` in `integrations/snowflake/node`
- `go test ./...` in `integrations/snowflake/e2e/caller-node`
- `go test ./...` in `integrations/snowflake/e2e/fake-snowflake`
- YAML parse for all Snowflake manifests/prompts
- `npm run build` in `control-plane/web/client`
- Docker image build for the Snowflake node
- Multi-node e2e smoke with AgentField control plane + caller node + Docker Snowflake node + fake Snowflake API
- Live Snowflake smoke using a scoped PAT:
  - direct SQL API auth succeeded
  - Docker node `query_readonly` succeeded through AgentField
  - control-plane Snowflake trigger polled and dispatched existing event rows
  - row inserted after trigger startup was picked up and dispatched with its own workflow id

## Safety / Scope

- Default SQL capability is read-only and denies DDL/DML/COPY/CALL/PUT/GET/role switching.
- Secrets are referenced by env var; secret values are not stored in trigger config.
- No real Snowflake token, account id, username, or smoke-test row ids are included in the diff.
- This PR is intentionally draft for product/API review before merge.


santoshkumarradha:
Review result: I would not merge this yet. Required checks are green, and local targeted validation passes, but I found two behavior blockers in the new Snowflake trigger source.

1. `control-plane/internal/sources/snowflake/snowflake.go:271-273` sets the default client to `http.DefaultClient`. `Execute` builds requests against `https://snowflakecomputing.com` and only rewrites to the validated account host when `sqlAPIClient.httpClient` is nil. The production `Run` path makes it non-nil, so polling uses the wrong host and will not hit the configured account. The current test covers `(&sqlAPIClient{}).httpClientFor(...)`, but not the production default path. I would either leave the default client nil so the rewrite transport is used, or construct the request URL directly from the validated account host, plus add a test that exercises the default production path.

2. `control-plane/internal/sources/snowflake/snowflake.go:321-325` and `:525-527` advance the poll watermark by timestamp only and the next query uses `>`. If `max_batch_size` cuts through rows that share the same `OCCURRED_AT`, the remaining rows at that timestamp are skipped permanently. This needs a composite cursor, for example timestamp plus event id with deterministic ordering, or another durable consumed-marker strategy.

Validation I ran:
- `go test ./internal/sources/snowflake ./internal/sources/all` from `control-plane`
- `go test ./...` from `integrations/snowflake/node`
- `go test ./...` from `integrations/snowflake/e2e/caller-node`
- `go test ./...` from `integrations/snowflake/e2e/fake-snowflake`
- `npm ci && npm run build` from `control-plane/web/client`
- YAML parse for the Snowflake package, contracts, and prompts

I am not admin-merging because these affect the new trigger source behavior.

PR #674, opened 2026-06-18
fix(security): restrict query API key auth to streaming routes (#423)

### LINKED ISSUE #423: [Security] API key accepted via ?api_key= query parameter
## Summary

`APIKeyAuth` accepts the API key via a `?api_key=` query parameter, which causes secrets to leak into access logs, proxy logs, browser history, and HTTP `Referer` headers.

## Context

At `auth.go:86`, after checking the `X-API-Key` header and `Authorization: Bearer` token, the middleware falls back to `c.Query("api_key")`. Query-string parameters are logged verbatim by virtually every reverse proxy (nginx, envoy, AWS ALB), appear in browser history and bookmark URLs, and are forwarded in `Referer` headers to any third-party resource loaded on the same page. A leaked API key grants full control-plane access. The query-param path was likely added for SSE client convenience but the tradeoff is not acceptable for general routes.

### In Scope

- Remove the `?api_key=` query-parameter fallback from `APIKeyAuth` for all standard REST routes.
- If SSE or WebSocket routes genuinely cannot set headers (browser `EventSource` API), restrict query-param auth to an explicit allowlist of those streaming paths only.
- Document the allowlist decision with a comment.

### Out of Scope

- Changing the header-based auth logic.
- Rotating or invalidating currently exposed keys — that is an ops action outside this PR.
- Adding OAuth or session-cookie auth — out of scope for this fix.

## Files

- `control-plane/internal/server/middleware/auth.go:86` — remove or restrict the `c.Query("api_key")` fallback
- `control-plane/internal/server/routes.go` — identify streaming routes that may need the restricted allowlist
- `control-plane/internal/server/middleware/auth_test.go` — add test: request with `?api_key=` on a standard REST route is rejected (401)

## Acceptance Criteria

- [ ] Standard REST endpoints reject requests that authenticate only via `?api_key=` query parameter (return 401)
- [ ] If an SSE allowlist is implemented, only allowlisted routes accept query-param auth
- [ ] Existing header-based auth tests continue to pass

## Notes for Contributors

**Severity**: MEDIUM

Check `control-plane/internal/server/routes.go` for any SSE endpoint registrations that use a browser `EventSource` — those are the only legitimate candidates for the allowlist. If none exist, remove the fallback entirely with no replacement.

Luffy2208:
## Summary

This PR removes the global `?api_key=` fallback from API key authentication on normal REST routes, because putting a secret in the URL can leak it into access logs, proxy logs, browser history, and Referer headers.

The change is intentionally scoped:
- `control-plane/internal/server/middleware/auth.go` now only accepts query-string API keys on an explicit allowlist of streaming routes.
- `control-plane/internal/server/routes_middleware.go` defines the allowlist for browser-driven SSE/WebSocket endpoints that cannot reliably send custom headers.
- `control-plane/internal/server/middleware/auth_test.go` now verifies that standard REST routes reject `?api_key=` with `401`, while allowlisted streaming routes still accept it.
- `control-plane/internal/handlers/agentic/helpers.go` and its coverage test were updated so query-string API keys are no longer treated as a general auth signal outside the middleware allowlist.

This keeps header-based auth unchanged, preserves the embedded UI’s streaming auth flow, and closes the security hole on ordinary API routes.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs only
- [ ] Tests only
- [ ] CI / tooling
- [x] Breaking change

## Test plan

- [x] `cd control-plane && go test ./...`
- [x] `make lint` (the target ran; local `golangci-lint` / `ruff` are not installed in this environment, so the wrapper completed but the tools themselves were not available)

## Test coverage

  **85%**.

      this PR only if the removal caused a legitimate regression and I

> the PR summary.

## Checklist

## Related issues / PRs

Fixes #423

### DIFF
--- a/control-plane/internal/handlers/agentic/coverage_additional_test.go
+++ b/control-plane/internal/handlers/agentic/coverage_additional_test.go
@@ -105,7 +105,7 @@ func TestGetAuthLevel_AdditionalFallbacks(t *testing.T) {
 		{
 			name:     "api key query parameter",
 			url:      "/test?api_key=abc",
-			expected: "api_key",
+			expected: "public",
 		},
 		{
 			name:     "non string context falls back to public",
--- a/control-plane/internal/handlers/agentic/helpers.go
+++ b/control-plane/internal/handlers/agentic/helpers.go
@@ -70,7 +70,7 @@ func getAuthLevel(c *gin.Context) string {
 	level, exists := c.Get("auth_level")
 	if !exists {
 		// Check if API key header is present as a fallback
-		if c.GetHeader("X-API-Key") != "" || strings.HasPrefix(c.GetHeader("Authorization"), "Bearer ") || c.Query("api_key") != "" {
+		if c.GetHeader("X-API-Key") != "" || strings.HasPrefix(c.GetHeader("Authorization"), "Bearer ") {
 			return "api_key"
 		}
 		return "public"
--- a/control-plane/internal/server/middleware/auth.go
+++ b/control-plane/internal/server/middleware/auth.go
@@ -10,15 +10,20 @@ import (
 
 // AuthConfig mirrors server configuration for HTTP authentication.
 type AuthConfig struct {
-	APIKey    string
-	SkipPaths []string
+	APIKey                  string
+	SkipPaths               []string
+	QueryAPIKeyAllowedPaths []string
 }
 
-// APIKeyAuth enforces API key authentication via header, bearer token, or query param.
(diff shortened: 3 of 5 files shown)


santoshkumarradha (approved):
Looked through the auth-path changes and the streaming allowlist. The scope matches the security issue, the behavior is covered in tests, and the current CI + CLA are green. I also re-ran the touched control-plane auth and handler suites locally in a clean worktree.

PR #677, opened 2026-06-20
feat(did+cp): DID-encrypted payloads + control-plane knowledge store (discuss/aggregator foundation)

AbirAbbas:
## Summary

Control-plane + SDK foundation for the **discuss/aggregator split**: a DID-based
payload-encryption primitive and a scope-aware knowledge (RAG) store hosted in
the control plane. This is the agentfield half; the hax-sdk half (the two agent
nodes + Discuss wiring) is a separate PR.

## What's here

**1. DID-based encryption (X25519 `keyAgreement`)**
- Agent DIDs gain an X25519 `keyAgreement` key derived from the master seed via
  HKDF (RFC 5869) with a salt distinct from the Ed25519 signing key, so the
  encryption key is cryptographically independent of the signing key.
- Exposed in DID resolution (`key_agreement`) and the W3C DID document; the
  X25519 private key is returned at registration.
- **Key rotation:** the rotation epoch is folded into the HKDF `info`
  (`<path>/enc/<epoch>`); `POST /api/v1/did/key-agreement/rotate` retires the
  prior key entirely (old-epoch payloads no longer decrypt).

**2. Encryption SDK primitive (JWE, cross-language)**
- Python (`agentfield/crypto.py`, joserfc) and TypeScript
  (`sdk/typescript/src/crypto/didEncryption.ts`, jose) encrypt/decrypt over the
  DID's X25519 key. Wire format: **JWE, ECDH-ES + A256GCM** — chosen because
  encrypt happens in TS (hax-sdk) and decrypt in Python (the aggregator), so the
  bytes must interop. A cross-language interop harness (`sdk/interop/`) proves
  both directions.

**3. Control-plane knowledge store (RAG)**
- `internal/embedding`: pinned-dimension embedding provider (OpenAI
  text-embedding-3-small, 1536-dim, with a deterministic fake fallback).
- `internal/knowledge`: scope-aware store with namespaces `ws:` / `proj:` /
  `sender:`; upsert writes under the most-specific namespace, search is
  **additive** (a project query also reads its parent workspace; a sender query
  reads sender + project + workspace).
- HTTP routes: `/api/v1/knowledge/{upsert,search,source/:id}`.
- Reuses the existing vector store — **no Qdrant**.

## Test plan

- [x] `go build ./...` and `go test ./... -short` green
- [x] `golangci-lint run` clean for the changed packages
- [x] Python SDK crypto tests + `ruff` green
- [x] TS SDK builds; cross-language JWE interop harness passes both directions
- [x] X25519 rotation: old-epoch payload rejected after rotate (SDK acceptance gate)

### DIFF
--- a/control-plane/cmd/x25519gen/main.go
+++ b/control-plane/cmd/x25519gen/main.go
@@ -1,0 +1,76 @@
+// Command x25519gen is a standalone interop fixture: it derives an X25519
+// keyAgreement keypair using the same HKDF derivation and JWK encoding the
+// DID service uses, and prints the public/private JWKs as JSON. It exists so the
+// control-plane-derived JWKs can be cross-checked against the SDK crypto layer
+// (encrypt to the public JWK, decrypt with the private JWK).
+package main
+
+import (
+	"crypto/ecdh"
+	"crypto/sha256"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+
+	"golang.org/x/crypto/hkdf"
+)
+
+// deriveX25519PrivateKey mirrors DIDService.deriveX25519PrivateKey: HKDF-SHA256
+// with the keyAgreement-specific salt and the derivation path as info.
+func deriveX25519PrivateKey(masterSeed []byte, derivationPath string) (*ecdh.PrivateKey, error) {
+	salt := []byte("agentfield-did-keyagreement-v1")
+	info := []byte(derivationPath)
+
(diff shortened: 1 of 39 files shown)


santoshkumarradha (approved):
Reviewed the full diff plus the new knowledge and DID rotation paths. I also reran the changed control-plane packages locally in a clean worktree, and the targeted Go suites passed. The cross-language SDK coverage is already green in CI, so this looks good to merge.

PR #679, opened 2026-06-21
fix(go-sdk): surface json.Marshal errors in memory backend requests

### LINKED ISSUE #434: [Go SDK] mustJSONReader silently drops json.Marshal errors
## Summary

`mustJSONReader` discards the `json.Marshal` error and returns an empty-body reader, silently sending a malformed or empty POST to the control plane when a value cannot be serialized.

## Context

`control_plane_memory_backend.go:402` defines `func mustJSONReader(v any) io.Reader { data, _ := json.Marshal(v); return bytes.NewReader(data) }`. When `json.Marshal` fails (unsupported type, cyclic struct, channel, function value), the error is thrown away and an empty `bytes.NewReader([]byte(nil))` is returned. The downstream HTTP POST to the control plane sends an empty body, the store silently records nothing or overwrites with an empty value, and the real marshal error is never surfaced to the caller. This makes debugging data-loss bugs in the memory backend extremely difficult.

### In Scope

- Rename to `jsonReader(v any) (io.Reader, error)` and return the `json.Marshal` error.
- Update all call sites (`Set`, `Get`, `Delete`) in `control_plane_memory_backend.go` to handle the error and propagate it to the caller.
- Remove the `must` prefix convention from any helper in this file that can realistically fail.

### Out of Scope

- Changing the `MemoryBackend` interface signature beyond what is needed to propagate this error (context propagation is tracked in G2).
- Adding JSON schema validation or custom marshal logic.

## Files

- `sdk/go/agent/control_plane_memory_backend.go:402` — replace `mustJSONReader` with `jsonReader(v any) (io.Reader, error)` and propagate the error
- `sdk/go/agent/control_plane_memory_backend.go` — update all callers of the helper to handle and return the error
- `sdk/go/agent/control_plane_memory_backend_test.go` — test: passing an unmarshalable value to `Set` returns an error rather than silently succeeding

## Acceptance Criteria

- [ ] `Set` (and any other caller of the serialization helper) returns a non-nil error when the value cannot be JSON-marshaled
- [ ] No `json.Marshal` errors are silently discarded anywhere in the file
- [ ] The `mustJSONReader` function (or any equivalent silent-drop helper) no longer exists

## Notes for Contributors

**Severity**: HIGH

Check the rest of the file for any other `_, _ = json.Something` or `data, _ :=` patterns and fix those at the same time. The `must` naming convention should only be used for operations that genuinely cannot fail (e.g. parsing a compile-time constant) — not for anything that depends on user-provided data.

ardittirana:
Fixes #434.

### Problem
`mustJSONReader` discarded the `json.Marshal` error:

```go
func mustJSONReader(v any) io.Reader { data, _ := json.Marshal(v); return bytes.NewReader(data) }
```

When a value can't be serialized (unsupported type, cyclic struct, channel, func), `json.Marshal` returns `nil, err`, the error is dropped, and the function returns a reader over a nil slice. The downstream control-plane `POST` is then sent with an **empty body** — silently storing nothing or overwriting with an empty value — and the real error never reaches the caller, making data-loss bugs in the memory backend very hard to debug.

### Fix
Replace it with `jsonReader(v any) (io.Reader, error)` that returns the marshal error, and propagate that error at each of the five call sites (`Set`, `Get`, `Delete`, `SetVector`, `SearchVector`) — all of which already return an `error`. No behavior change on the success path. Stdlib only.

### Test
Updates the existing helper test and adds a case asserting that an unserializable value (a channel) returns an error and a nil reader instead of silently yielding an empty body. `go test ./agent/` passes (all existing backend tests included); `gofmt`/`go vet` clean.

### DIFF
--- a/sdk/go/agent/control_plane_memory_backend.go
+++ b/sdk/go/agent/control_plane_memory_backend.go
@@ -85,15 +85,7 @@ func (b *ControlPlaneMemoryBackend) Get(scope MemoryScope, scopeID, key string) 
 		return nil, false, err
 	}
 
-	body := map[string]any{
-		"key":   key,
-		"scope": b.apiScope(scope),
-	}
-	reader, err := jsonReader(body)
-	if err != nil {
-		return nil, false, err
-	}
-	req, err := http.NewRequest(http.MethodPost, endpoint, reader)
+	req, err := http.NewRequest(http.MethodPost, endpoint, keyScopeReader(key, b.apiScope(scope)))
 	if err != nil {
 		return nil, false, err
 	}
@@ -126,15 +118,7 @@ func (b *ControlPlaneMemoryBackend) Delete(scope MemoryScope, scopeID, key strin
 		return err
 	}
 
-	body := map[string]any{
-		"key":   key,
-		"scope": b.apiScope(scope),
-	}
-	reader, err := jsonReader(body)
-	if err != nil {
-		return err
-	}
-	req, err := http.NewRequest(http.MethodPost, endpoint, reader)
+	req, err := http.NewRequest(http.MethodPost, endpoint, keyScopeReader(key, b.apiScope(scope)))
 	if err != nil {
 		return err
 	}
@@ -425,4 +409,16 @@ func jsonReader(v any) (io.Reader, error) {
 		return nil, err
 	}
 	return bytes.NewReader(data), nil
+}
+
+func keyScopeReader(key, scope string) io.Reader {
+	var body bytes.Buffer
+	_ = json.NewEncoder(&body).Encode(struct {
+		Key   string `json:"key"`
+		Scope string `json:"scope"`
+	}{
+		Key:   key,
+		Scope: scope,
+	})
+	return &body
 }
--- a/sdk/go/agent/control_plane_memory_backend.go
+++ b/sdk/go/agent/control_plane_memory_backend.go
@@ -56,7 +56,11 @@ func (b *ControlPlaneMemoryBackend) Set(scope MemoryScope, scopeID, key string, 
 		"data":  value,
 		"scope": b.apiScope(scope),
 	}
-	req, err := http.NewRequest(http.MethodPost, endpoint, mustJSONReader(body))
+	reader, err := jsonReader(body)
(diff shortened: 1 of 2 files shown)


santoshkumarradha:
Thanks @ardittirana, this looks good.

CLA is still pending. Can you sign it here?

https://cla-assistant.io/Agent-Field/agentfield?pullRequest=679

Once that clears, we can continue.

ardittirana:
Signed 🫡

santoshkumarradha (dismissed):

ardittirana:
Thanks for the review and approval, @santoshkumarradha — CLA's signed now, so it should be good to go. Appreciate the quick turnaround!

santoshkumarradha:
Thanks for the follow-up. CLA is green now, but the branch is still blocked by the required `coverage-summary` check.

The failing part is patch coverage on `sdk/go/agent/control_plane_memory_backend.go` at 67%, so this still needs tests that hit the uncovered error-return branches before it can merge. Once that check goes green, the rest of the review state looks fine.

santoshkumarradha:
I rechecked this in a clean worktree. `go test ./agent/...` still passes locally, and the remaining blocker looks narrower than the earlier summary: the patch-coverage failure is likely coming from the new `jsonReader` error branches in `Get` and `Delete`, where the request body is string-only and cannot actually fail `json.Marshal`. A small follow-up that keeps the error-returning path on the dynamic payload calls (`Set`, `SetVector`, `SearchVector`) and leaves `Get` / `Delete` on the simple reader path should get this over the line.

santoshkumarradha:
I pushed a small maintainer follow-up to this branch that removes the dead marshal-error path on the fixed-string `Get` and `Delete` requests while keeping the real error propagation on the dynamic payload calls.

Local `go test ./agent/... -count=1` still passes from `sdk/go`. I’m waiting on the rerun of the required checks now.

santoshkumarradha (approved):
This follow-up keeps the error-returning path where serialization can actually fail and removes the dead branch on the fixed-string request bodies. Local `go test ./agent/... -count=1` still passes from `sdk/go`.

PR #682, opened 2026-06-23
fix(#424): fail closed on empty API key unless insecure mode is explicitly enabled

### LINKED ISSUE #424: [Security] Empty API key silently disables all HTTP auth
## Summary

When `config.APIKey` is empty, `APIKeyAuth` silently becomes a no-op, so a production deployment that forgets to set `AGENTFIELD_API_KEY` runs with no HTTP authentication and emits no warning.

## Context

The first lines of `APIKeyAuth` at `auth.go:26` are `if config.APIKey == "" { c.Next(); return }`. A mis-injected secret, a misconfigured Kubernetes secret mount, or an accidental environment variable omission silently disables all auth for every API route. There is no startup log, no metric, and no health-check signal indicating auth is off. This has caused production outages in similar systems when ops teams deployed without noticing the missing env var.

### In Scope

- Either: fail server startup when API auth is configured but the key is empty (preferred).
- Or: require an explicit `AGENTFIELD_INSECURE_DISABLE_AUTH=true` (or equivalent config flag) to allow the no-op, and emit a `WARN` log on every request when it is active, plus a loud startup log.
- Whichever path is chosen, make it explicit and auditable.

### Out of Scope

- Changing how keys are validated once present.
- Forcing API keys to meet complexity requirements — that is a separate concern.
- Modifying the admin token middleware (tracked in S7).

## Files

- `control-plane/internal/server/middleware/auth.go:26` — replace silent no-op with explicit startup guard or insecure-mode flag
- `control-plane/internal/config/config.go` — add `InsecureDisableAuth bool` field if taking the explicit-flag route
- `control-plane/cmd/agentfield-server/main.go` — add startup validation that rejects empty API key unless insecure mode is explicitly set
- `control-plane/internal/server/middleware/auth_test.go` — test: empty API key with no insecure flag causes startup error (or middleware panics with clear message)

## Acceptance Criteria

- [ ] Server refuses to start (or middleware refuses to register) when `APIKey` is empty and no explicit insecure-mode flag is set
- [ ] If insecure mode is explicitly enabled, a `WARN`-level log is emitted at startup and optionally on each unauthenticated request
- [ ] Existing tests that configure a non-empty API key are unaffected

## Notes for Contributors

**Severity**: MEDIUM

Look at how `AdminTokenAuth` (S7) is fixed — apply the same pattern here so both middleware pieces are consistent. The startup-validation approach (fail in `main.go` or `server.go` before routes are registered) is generally easier to test than a middleware-level panic.

Luffy2208:
# Fix: Enforce Fail-Closed API Authentication Configuration

## Summary

This PR fixes the critical security vulnerability described in **#424**, where an empty or missing API key silently disabled authentication across all HTTP endpoints.

The fix introduces a **fail-closed startup policy** that prevents the server from launching when API authentication is enabled but no API key is configured, unless an explicit insecure-mode flag is provided.

#### `control-plane/internal/server/middleware/auth.go`

- Removed the silent early-return behavior that effectively disabled authentication when the API key was empty.
- Added an explicit fail-closed guard that rejects requests when the API key is not configured and insecure mode is disabled.
- Prevents empty authentication headers or bearer tokens from matching an empty configuration through `subtle.ConstantTimeCompare`.

#### `control-plane/internal/server/server.go`

- Added centralized configuration validation through `validateAPIAuthConfig`.
- Server startup now fails if authentication is configured insecurely.
- Emits a prominent warning log when authentication is intentionally disabled via the insecure-mode flag.

#### `control-plane/internal/config/config.go` and `cmd/*`

- Added `InsecureDisableAuth` to `AuthConfig`.
- Added support for the following environment variables:
  - `AGENTFIELD_INSECURE_DISABLE_AUTH`
  - `AGENTFIELD_API_AUTH_INSECURE_DISABLE_AUTH`

#### Tests

Updated and added tests covering:

- Authentication middleware behavior when API keys are missing.
- Fail-closed behavior returning `401 Unauthorized`.
- Server startup validation rejecting insecure configurations without the explicit override flag.
- Explicit insecure-mode startup behavior.

Files updated:

- `control-plane/internal/server/middleware/auth_test.go`
- `control-plane/internal/server/server_additional_test.go`

---

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs only
- [ ] Tests only
- [ ] CI / tooling
- [ ] Breaking change

---

### Authentication Middleware Tests

Verifies that missing API keys without the insecure flag fail closed and return `401 Unauthorized`.

```bash
cd control-plane
go test -v ./internal/server/middleware/...
```

### Full Control Plane Test Suite

Ensures all routes, startup paths, and configuration validation logic pass successfully.

```bash
cd control-plane
go test ./...
```

### Static Analysis

```bash
make lint
```

---

## Test Coverage

Before requesting review, confirm the following:

- [x] I ran tests for the surfaces I changed locally.
- [x] If I removed code, I updated `coverage-baseline.json` only if the removal caused a legitimate coverage regression and documented it above.
- [x] The coverage gate check is passing in CI.

---

## Checklist

- [x] I have read `CONTRIBUTING.md` (if present) and `docs/DEVELOPMENT.md`.
- [x] Commits are signed and follow the Conventional Commits specification.

---

## Related Issues

Fixes #424

### DIFF
--- a/control-plane/cmd/af/main.go
+++ b/control-plane/cmd/af/main.go
@@ -216,6 +216,7 @@ func loadConfig(configFile string) (*config.Config, error) {
 	// This is needed because Viper's AutomaticEnv only works for keys that exist in config
 	_ = viper.BindEnv("api.auth.api_key", "AGENTFIELD_API_KEY")
 	_ = viper.BindEnv("api.auth.api_key", "AGENTFIELD_API_AUTH_API_KEY")
+	_ = viper.BindEnv("api.auth.insecure_disable_auth", "AGENTFIELD_INSECURE_DISABLE_AUTH", "AGENTFIELD_API_AUTH_INSECURE_DISABLE_AUTH")
 
 	// Get the directory where the binary is located for UI paths
 	execPath, err := os.Executable()
--- a/control-plane/cmd/agentfield-server/main.go
+++ b/control-plane/cmd/agentfield-server/main.go
@@ -245,6 +245,7 @@ func loadConfig(configFile string) (*config.Config, error) {
 	// This is needed because Viper's AutomaticEnv only works for keys that exist in config
 	_ = viper.BindEnv("api.auth.api_key", "AGENTFIELD_API_KEY")
 	_ = viper.BindEnv("api.auth.api_key", "AGENTFIELD_API_AUTH_API_KEY")
+	_ = viper.BindEnv("api.auth.insecure_disable_auth", "AGENTFIELD_INSECURE_DISABLE_AUTH", "AGENTFIELD_API_AUTH_INSECURE_DISABLE_AUTH")
 	// AutomaticEnv makes viper.IsSet("features.did.enabled") return true once
 	// AGENTFIELD_FEATURES_DID_ENABLED is set, but Unmarshal won't actually
 	// populate the struct field unless the key is bound. Without this, setting
--- a/control-plane/config/agentfield.yaml
+++ b/control-plane/config/agentfield.yaml
@@ -25,6 +25,9 @@ ui:
   dev_port: 5173
 
 api:
+  auth:
+    # Local development only. Production deployments should configure api_key.
+    insecure_disable_auth: true
   cors:
     allowed_origins:
       - "http://localhost:3000"
--- a/control-plane/config/docker-perf.yaml
+++ b/control-plane/config/docker-perf.yaml
@@ -34,6 +34,9 @@ ui:
   backend_url: ""
 
 api:
+  auth:
+    # This benchmark configuration runs in an isolated local environment.
+    insecure_disable_auth: true
   cors:
     allowed_origins:
       - "*"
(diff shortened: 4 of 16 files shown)


santoshkumarradha:
Nice, thanks for taking this on. I think there’s just one deployment path left to update.

`deployments/kubernetes/base/control-plane-deployment.yaml` still runs with `AGENTFIELD_CONFIG_FILE=/dev/null` and no API key. After this change, that manifest won’t start unless it also sets `AGENTFIELD_INSECURE_DISABLE_AUTH=true` or wires an API key.

Could you update that base manifest too?

santoshkumarradha (approved):
Thanks for tightening this up. Failing closed unless unauthenticated mode is explicitly enabled is the right default, and the config and server tests cover the main regressions.

PR #691, opened 2026-06-25
fix(security): Enforce fail-closed for empty admin token (#425)

### LINKED ISSUE #425: [Security] Empty admin token silently disables admin guard
## Summary

When `adminToken` is empty, `AdminTokenAuth` silently becomes a no-op, so admin-only routes fall back to ordinary API-key auth and lose their dedicated protection layer.

## Context

`AdminTokenAuth` at `auth.go:117` starts with `if adminToken == "" { c.Next(); return }`. Admin routes (node management, config reload, debug endpoints) are meant to require a separate, higher-privilege token. When the admin token is not configured, those routes silently accept any valid API key, eliminating the intended separation between regular agent traffic and administrative operations. In a shared deployment this lets a compromised agent API key control admin functionality.

### In Scope

- Refuse to register admin routes at startup if `adminToken` is empty and no explicit insecure-mode flag is set.
- Or: emit a loud startup `WARN` and require `AGENTFIELD_INSECURE_ADMIN_NO_TOKEN=true` to allow operation without an admin token.
- Ensure the fix is consistent with the approach taken for S6 (`APIKeyAuth`).

### Out of Scope

- Merging admin auth and API key auth into a single middleware.
- Adding role-based access within the admin token scope.
- Modifying routes that are not behind `AdminTokenAuth`.

## Files

- `control-plane/internal/server/middleware/auth.go:117` — replace silent no-op with explicit guard or insecure-mode check
- `control-plane/cmd/agentfield-server/main.go` — startup validation: reject empty admin token unless insecure flag set
- `control-plane/internal/server/middleware/auth_test.go` — test: empty admin token with no insecure flag causes startup error; admin route with only API key returns 401

## Acceptance Criteria

- [ ] Server refuses to start (or admin routes refuse to register) when `adminToken` is empty and no insecure override is set
- [ ] Admin routes return 401 when a request presents only a valid API key but no admin token
- [ ] Existing tests with a configured admin token continue to pass

## Notes for Contributors

**Severity**: MEDIUM

Fix S6 (`APIKeyAuth` silent no-op) first — use exactly the same pattern here so both middleware pieces behave consistently. The startup-validation approach in `main.go` or `server.go` is preferable to a middleware-level check so the failure is immediate and obvious before any routes are served.

Luffy2208:
## Summary

Fixes the issue where an empty admin token silently disabled route protection. 

* **`internal/server/middleware/auth.go`**: Added `ValidateAdminTokenAuth` and updated `AdminTokenAuth` middleware to return 401 instead of letting requests through when no token is set.
* **`internal/config/config.go`**: Added `InsecureDisableAdminAuth` field to configuration and mapped the `AGENTFIELD_INSECURE_ADMIN_NO_TOKEN` env var.
* **`internal/server/server.go`**: Added startup validation that stops the server from launching if the admin token is missing, unless insecure mode is explicitly configured.
* **`internal/server/routes_admin.go`**: Updated route registration to pass the new config structure to the middleware.
* **`tests`**: Added and updated unit tests to verify the new startup validations, env override flags, and 401/403 responses.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs only
- [ ] Tests only
- [ ] CI / tooling
- [ ] Breaking change

## Test plan

Run the following test commands locally to verify the new startup validations, the middleware's fail-closed unauthorized/forbidden assertions, and configuration override behavior:

- `cd control-plane && go test -v ./internal/server/...`
- `cd control-plane && go test -v ./internal/server/middleware/...`
- `cd control-plane && go test -v ./internal/config/...`

## Test coverage

## Checklist

## Related issues / PRs

Closes #425

### DIFF
--- a/control-plane/internal/config/config.go
+++ b/control-plane/internal/config/config.go
@@ -307,8 +307,12 @@ type AuthorizationConfig struct {
 	// DefaultApprovalDurationHours is the default duration for permission approvals
 	DefaultApprovalDurationHours int `yaml:"default_approval_duration_hours" mapstructure:"default_approval_duration_hours" default:"720"`
 	// AdminToken is a separate token required for admin operations (tag approval,
-	// policy management). If empty, admin routes fall back to the standard API key.
+	// policy management). When empty and InsecureDisableAdminAuth is false, admin
+	// routes will reject requests at startup or return 401.
 	AdminToken string `yaml:"admin_token" mapstructure:"admin_token"`
+	// InsecureDisableAdminAuth explicitly permits running admin routes without
+	// an admin token. This should only be enabled for trusted local development.
+	InsecureDisableAdminAuth bool `yaml:"insecure_disable_admin_auth" mapstructure:"insecure_disable_admin_auth"`
 	// InternalToken is sent as Authorization: Bearer header when the control plane
 	// forwards execution requests to agents. Agents with RequireOriginAuth enabled
 	// validate this token, preventing direct access to their HTTP ports.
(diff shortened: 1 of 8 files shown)


santoshkumarradha (approved):
Checked the auth/config path locally. The fail-closed behavior and explicit insecure escape hatch both look right, and the focused Go tests passed on my side.

PR #692, opened 2026-06-26
feat(cli): af install/run for agent nodes — encrypted secrets, env prompting, node-to-node deps

AbirAbbas:
## Summary

The `af install` / `af run` agent-node scaffolding has existed since the very first commit but was effectively unusable for real nodes: it required a top-level `main.py` (real nodes start via `python -m pkg.app`), hardcoded `python main.py` at launch, exported `AGENTFIELD_SERVER_URL` while the SDK reads `AGENTFIELD_SERVER`, and stored secrets as plaintext `.env`. This PR makes the flow actually work end-to-end and adds the pieces needed for day-to-day use.

## What's new

- **Real-node install/run.** A node is now defined by its `agentfield-package.yaml` manifest with an `entrypoint.start` (e.g. `python -m pr_af.app`) — no `main.py` required. The runner launches via the manifest entrypoint, honors the manifest `healthcheck` path, and exports `AGENTFIELD_SERVER` (+ legacy `AGENTFIELD_SERVER_URL`).
- **Encrypted secret store.** Secrets live encrypted at rest (AES-256-GCM) under `~/.agentfield/secrets/` with a random 32-byte key in `~/.agentfield/keyring/master.key` (`0600`). They are decrypted only into the child process' environment at start time — never written back to disk in plaintext. Global scope is shared across nodes; node scope overrides it.
- **Env prompting.** On `af run`, required variables resolve in order: process env → node store → global store → manifest default → prompt (hidden for `type: secret`), persisting prompted secrets encrypted. Missing required vars in a non-interactive session produce a clean error instead of hanging.
- **`af secrets` command.** `set` / `ls` (values masked) / `rm`, with `--node` scoping.
- **Node-to-node dependencies.** A manifest may declare `dependencies.nodes` (e.g. `af://registry/swe-planner` → `github.com/Agent-Field/<name>`, or a git URL). `af install` pulls them in recursively (skipping already-installed, which breaks cycles); `af run` starts a node's dependencies first, in order, before allocating its own port — and leaves already-running dependencies untouched.

## Notable fixes uncovered while verifying locally

- The CLI's install/run path runs through `internal/core/services`, a duplicate of the `internal/packages` logic — fixes are now applied there (the two layers share one `ValidatePackage` / `ParsePackageMetadata` / `ShouldSkipCopy`).
- Dependencies were started after the parent allocated its port, causing a port collision; dependency startup now happens before port allocation.

## Verification

Verified end-to-end against a live local control plane with two no-LLM nodes built on the real SDK:
- `af install` an entrypoint-only node (no `main.py`) → registers → `af call node.reasoner` returns a real result.
- A missing required secret errors cleanly; after `af secrets set`, `af run` injects `AGENTFIELD_SERVER` + the stored secret + manifest defaults into the process (confirmed via the node's own env dump; no plaintext on disk, files `0600`).
- Multi-agent: `af run greeter-node` auto-starts its dependency `echo-node` first (distinct ports), both register and execute; an already-running dependency is not restarted.

## Docs

- `docs/installing-agent-nodes.md` — full guide to install/run, the manifest schema, the encrypted secrets model, and `af secrets`.
- `cli-toolkit.md` reference updated (+ embedded skill copy synced).

## Test plan

- [x] `go build ./...` clean
- [x] full `go test ./...` for control-plane green (47 packages)
- [x] new unit tests for the secret store, env resolver, manifest validation, and node-dependency resolution / ordered-start helpers
- [x] manual end-to-end (single node + multi-agent auto-start) against a live control plane

## Follow-ups (not in this PR)

- Author `agentfield-package.yaml` manifests in the public node repos (SWE-AF, pr-af, sec-af, cloudsecurity-af, af-template).
- `af dev` (a third copy of the launch logic) still hardcodes `main.py`; collapsing the duplicated install/run implementations into one is a good follow-up.

### DIFF
--- a/control-plane/internal/cli/coverage_gap_additional_test.go
+++ b/control-plane/internal/cli/coverage_gap_additional_test.go
@@ -219,10 +219,10 @@ func TestCLIExitHelper(t *testing.T) {
 		_ = verifyVC(filepath.Join(t.TempDir(), "missing.json"), VerifyOptions{OutputFormat: "json"})
 	case "output-json-invalid":
 		_ = outputJSON(VCVerificationResult{
-			Valid:      false,
-			Type:       "credential",
+			Valid:       false,
+			Type:        "credential",
 			FormatValid: false,
-			Message:    "bad",
+			Message:     "bad",
 		})
 	case "output-pretty-invalid":
 		_ = outputPretty(VCVerificationResult{
@@ -607,9 +607,9 @@ func TestEnhancedWorkflowVerificationBranches(t *testing.T) {
 	t.Run("validate vc structure covers required fields", func(t *testing.T) {
 		verifier := NewEnhancedVCVerifier(nil, false)
 		cases := []struct {
-			name  string
-			doc   types.VCDocument
-			want  string
+			name string
+			doc  types.VCDocument
+			want string
 		}{
 			{name: "missing context", doc: types.VCDocument{}, want: "missing @context"},
 			{name: "missing type", doc: types.VCDocument{Context: []string{"ctx"}}, want: "missing type"},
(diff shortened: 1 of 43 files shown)


santoshkumarradha:
@AbirAbbas before in, can you test it with our swe/pr-af etc.. ?

santoshkumarradha:
Rechecked this today. The main blocker is still the required `coverage-summary` patch gate: `control-plane` patch coverage is at 67%, with most of the misses in `internal/cli/secrets.go`, `internal/packages/env_resolver.go`, `internal/packages/secrets.go`, and `internal/packages/runner.go`. Once those paths are covered and you have a real `swe` / `pr-af` style run through the install flow, I can take another merge pass.

PR #697, opened 2026-06-29
feat(sdk): ReasonerFailed exception for reporting reasoner failure with result

AbirAbbas:
## Summary

The async execution handler records an execution as `succeeded` whenever the reasoner **returns** a value — it never inspects the result. A reasoner whose own payload says `success: False` (e.g. a SWE-AF build that completed zero issues and merged nothing) therefore surfaces as green, which is easy to act on incorrectly.

This adds `ReasonerFailed`, raised *inside* a reasoner to report that the work ran but failed. The handler maps it to `status="failed"` while still posting the structured `result`, so the control plane — which stores the result payload regardless of terminal status — keeps the rich outcome (debt, DAG state, any PR opened) instead of just a bare error string.

This is the SDK half of Agent-Field/SWE-AF#82 (Gap 2). The SWE-AF side raises `ReasonerFailed` when a build ships nothing.

## Changes Made

- `exceptions.py`: new `ReasonerFailed(AgentFieldError)` carrying `message` + optional `result` + `error_details`; exported.
- `__init__.py`: export `ReasonerFailed` (import + `__all__`).
- `agent.py`: in `_execute_async_with_callback`'s failure path, when the raised exception is a `ReasonerFailed` with a non-`None` result, attach `jsonable_encoder(result)` to the `status="failed"` payload. `error_details` already flows through the existing generic path.

## Test Plan

- [x] New unit tests in `tests/test_async_execution.py`:
  - `ReasonerFailed(result=...)` → status callback is `failed`, `error`/`error_details` set, **and** `result` preserved.
  - Plain `Exception` → `failed` with **no** `result` key (regression guard — result-preservation must not leak into the ordinary failure path).
- [x] `ruff check .` clean (full `sdk/python`).
- [x] Full SDK pytest suite passes.

## Compatibility

Purely additive — existing reasoners that return values are unchanged. Only reasoners that explicitly `raise ReasonerFailed` get the new failed-with-result behavior.

### DIFF
--- a/sdk/python/agentfield/__init__.py
+++ b/sdk/python/agentfield/__init__.py
@@ -64,6 +64,7 @@ from .exceptions import (
     AgentFieldClientError,
     ExecutionTimeoutError,
     MemoryAccessError,
+    ReasonerFailed,
     RegistrationError,
     ValidationError,
 )
@@ -158,6 +159,7 @@ __all__ = [
     "AgentFieldClientError",
     "ExecutionTimeoutError",
     "MemoryAccessError",
+    "ReasonerFailed",
     "RegistrationError",
     "ValidationError",
     # Trigger / webhook plugin system
--- a/sdk/python/agentfield/agent.py
+++ b/sdk/python/agentfield/agent.py
@@ -2573,6 +2573,14 @@ class Agent(FastAPI):
                 "execution_id": execution_id,
                 "reasoner": reasoner_name,
             }
+            # A reasoner that ran, determined its own work failed, and wants its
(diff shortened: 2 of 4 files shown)


santoshkumarradha:
@AbirAbbas do we need to replicate for tsx/go or nay ?

santoshkumarradha (approved):
Read through the SDK-side failure-path change and reran `uv run --extra dev pytest tests/test_async_execution.py -q` plus a quick compile check locally. The new `ReasonerFailed` path keeps the structured result on failed async executions without changing the plain exception path, and the added tests cover both cases well.

PR #717, opened 2026-07-04
fix(control-plane): add timestamp freshness checks to prevent webhook replay attacks

### LINKED ISSUE #65: Inbound webhook secrets stored in plaintext and replayable signatures
## Summary
- Inbound webhook triggers persist signing secrets in plaintext and echo them back in API responses. `CreateTrigger` sets `SecretHash` to the raw secret and returns it, and `RotateSecret` does the same on every rotation.
- Signature verification accepts any timestamp; there is no freshness check or replay protection, so a captured body + signature can be replayed indefinitely.

## Impact
- Anyone with database/backup access (or logs of the creation response) can recover every webhook secret and forge requests. There is no hashing/encryption to limit blast radius.
- Because timestamps are not validated, an intercepted webhook can be replayed forever as long as the body/signature pair is known, bypassing idempotency TTLs.

## Evidence
- `control-plane/internal/handlers/webhook_triggers.go:104-151` stores `SecretHash: secret` and returns `"secret": <raw>` in the creation response; `RotateSecret` returns the raw secret as well.
- `control-plane/internal/webhooks/signature.go:26-37` validates the HMAC but never enforces timestamp age/skew, and `ReceiveWebhook` trusts whatever timestamp is provided (`webhook_triggers.go:479-488`).

## Recommendation
- Store webhook secrets hashed or envelope-encrypted; avoid returning them after creation/rotation (surface only once or offer a short-lived token retrieval endpoint guarded by auth/audit).
- Enforce timestamp freshness (e.g., 5–10 minute skew window) and reject stale or non-numeric timestamps; consider adding nonce-based replay protection.
- Add migrations/tests to cover secret-at-rest handling and replay rejection so regressions are caught.

7vignesh:
## Summary

The `generic_hmac` source and the approval webhook handler accepted any timestamp without checking freshness, allowing captured signatures to be replayed indefinitely. This adds configurable timestamp tolerance to reject stale/replayed webhook requests.

Note: The original issue also mentioned plaintext secret storage, but the codebase has since been refactored to use `SecretEnvVar` (env var names stored, secrets read from environment at request time), so that part is already resolved.

Closes #65

## Type of change

- [x] Bug fix

## Test plan

- [x] `cd control-plane && go test ./internal/sources/generichmac/ -v` (14 tests, 7 new)
- [x] `cd control-plane && go test ./internal/handlers/ -run "TestApprovalWebhook" -v` (17 tests, 2 new)
- [x] `cd control-plane && CGO_ENABLED=0 GOOS=linux go build ./cmd/agentfield-server`
- [x] `golangci-lint run` no new issues

## Test coverage

## Checklist

- [x] I have linked any related issues (Fixes #65).

## Related issues / PRs

Fixes #65


control-plane/internal/sources/generichmac/hmac.go
```diff
@@ -77,6 +88,28 @@ func (s *source) HandleRequest(ctx context.Context, req *sources.RawRequest, cfg
 	}
 	c := parseConfig(cfg)
 
+	// Enforce timestamp freshness if a timestamp header is configured.
+	if c.TimestampHeader != "" {
+		tsStr := strings.TrimSpace(req.Headers.Get(c.TimestampHeader))
```
AbirAbbas: This is the gap: the freshness check reads the timestamp from an unauthenticated header, but the HMAC below only signs the request body (`mac.Write(req.Body)`, ~L125) — the timestamp isn't covered by the signature. A replay attacker resends the captured body+signature untouched and just rewrites this header to `now`: freshness passes, the signature still matches, replay succeeds. Since the attacker controls all headers, this provides essentially no protection as written.

To make it real, bind the timestamp into the signed payload when `timestamp_header` is set — Stripe-style, verify the HMAC over `timestamp + "." + body` instead of the bare body. Then a forged `now` invalidates the signature. Worth adding a test that asserts the attack fails: a valid signed body with a rewritten fresh timestamp must be **rejected**. (The approval-webhook path already does this correctly — it signs `ts + "." + body`.)

AbirAbbas (commented):
Can't approve this yet — the replay protection on the `generic_hmac` source doesn't actually stop a replay. Detail inline. The approval-webhook half is correct (the timestamp is inside the signed payload there), so it's specifically the generic_hmac path that needs fixing.

One note for closing #65: the plaintext-secret half of that issue is already handled separately by the SecretEnvVar refactor, so this PR only needs to land working replay protection to close it — but it does need to actually bind the timestamp to the signature first.

7vignesh:
@AbirAbbas 
latest commit binds the timestamp into the HMAC signature for generic_hmac (mirroring the approval-webhook approach), so replays now fail verification, not just the freshness check. Should close the replay-protection part of #65. Ready for re-review whenever you're free.

AbirAbbas (approved):

PR #723, opened 2026-07-06
Fix/fire and forget

### LINKED ISSUE #622: [Python SDK] fire_and_forget_update isn't F&F
## Describe the bug

`agent_workflow.py` fire_and_forget_update can block completion of a reasoner / skill because it's actually waiting for the completion and not merely a notification. 
`log_execution` can block, since it's a thread method. If any of the calls fail to POST due to a deadlock, it will wait forever.

Related to #620 

## Steps to reproduce

2. Run '...'
3. See error

## Environment

- Control plane version:

DebanKsahu:
## Summary

- Fixes the blocking fire and forget issue by sending them to a background dispatcher `_NotificationDispatcher` (Which itself run as task). Internally it uses a `asyncio.Queue` which help to maintain the order of the notification.
- For `execute_with_tracking()` method in `AgentWorkflow` class, it internally has an agent instance which it uses to delegate notification tasks to dispatcher.
- Attached the `_NotificationDispatcher` class's shutdown method to `Agent` class's `_cleanup_async_resources()` method.
- To use `uvicorn` initialized event loop all queue and task are in `_NotificationDispatcher` class are lazily initialized in lifespan of the agent which is managed in `AgentServer` class.
- Left the sync `log_execution` part as it is as it seems lightweight no, needless to delegate it to a thread.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs only
- [ ] Tests only
- [ ] CI / tooling
- [ ] Breaking change

## Test plan

- [x] `./scripts/test-all.sh`
- [x] Verified the branch rebases cleanly on latest upstream/main.

## Test coverage

I ran the Python SDK coverage test locally because this PR only changes Python-side code.

Command used:

```
cd sdk/python
uv run pytest --cov=agentfield --cov-report=term-missing
```
Result:

1662 passed, 4 skipped, 36 deselected, 27 warnings
TOTAL coverage: 80%

## Checklist

## Related issues / PRs
Fixes #622


sdk/python/agentfield/agent.py
```diff
@@ -2245,13 +2245,17 @@ async def _execute_reasoner_endpoint(
 
         if hasattr(self, "workflow_handler") and self.workflow_handler:
             execution_context.reasoner_name = reasoner_id
-            await self.workflow_handler.notify_call_start(
-                execution_context.execution_id,
-                execution_context,
-                reasoner_id,
-                payload_dict,
-                parent_execution_id=execution_context.parent_execution_id,
+            task = asyncio.create_task(
```
AbirAbbas: Detaching start/complete/error into independent tasks means the events can now hit the endpoint out of order — previously `notify_call_start` was awaited before the reasoner body ran, so "started" was guaranteed to post before "completed". If the control-plane telemetry orders by timestamp/status this is fine, but it's a real weakening of the prior contract — worth confirming the CP tolerates out-of-order start/complete and noting it in the PR.

sdk/python/agentfield/agent.py
```diff
@@ -2245,13 +2245,17 @@ async def _execute_reasoner_endpoint(
...
+                self.workflow_handler.notify_call_start(
+                    execution_context.execution_id,
+                    execution_context,
+                    reasoner_id,
+                    payload_dict,
+                    parent_execution_id=execution_context.parent_execution_id,
+                )
             )
+            self._background_tasks.add(task)
```
AbirAbbas: These background tasks are never drained on shutdown — `stop()` → `_cleanup_async_resources()` calls `self.client.aclose()` while notification POSTs may still be in flight, so the final complete/error event can be dropped if the reasoner returns right before teardown. Previously that terminal event was awaited, i.e. guaranteed delivered.

Suggest draining before the client closes: `await asyncio.gather(*self._background_tasks, return_exceptions=True)` (with a short timeout) in `_cleanup_async_resources`, ahead of `aclose()`.

AbirAbbas (commented):
Good direction — detaching the telemetry POST out of the reasoner's critical path is the right fix for #622, and the GC-safe idiom (`_background_tasks.add` + `add_done_callback(discard)`) is done correctly at all the sites. Two things holding back an approval, both inline.

Also: there are no tests for the new non-blocking behavior (the "covered by tests" box is unchecked), and this only fixes half of #622 — `log_execution` is still called synchronously on the event loop. Fine to scope that out, but say so in the PR/issue rather than leaving it silent. A `ruff format` pass would also clear the trailing-whitespace/indentation noise.

sdk/python/agentfield/agent.py
DebanKsahu: Internally task queue follow the FIFO property so if start pushed first then it will execute first unless and await called will be executed inside the first task which leads it to sleep and task after the first task will be executed. Here the only async code inside the all notify functions is `fire_and_forget_update()` method which might cause issue.

I have 2 approach to this:
1) Convert `fire_and_forget_update()` to a task also so that start, complete and error execute in order then their fire_and_forget_update also queued in order.
2) Instead of converting the entire notify function into a task we should only convert the respective `fire_and_forget_update()` calls into task.

Please review these 2 approaches.

sdk/python/agentfield/agent.py
DebanKsahu: Sorry for my mistake, I will work on it in the next commits.

DebanKsahu:
> Good direction — detaching the telemetry POST out of the reasoner's critical path is the right fix for #622, and the GC-safe idiom (`_background_tasks.add` + `add_done_callback(discard)`) is done correctly at all the sites. Two things holding back an approval, both inline.
> 
> Also: there are no tests for the new non-blocking behavior (the "covered by tests" box is unchecked), and this only fixes half of #622 — `log_execution` is still called synchronously on the event loop. Fine to scope that out, but say so in the PR/issue rather than leaving it silent. A `ruff format` pass would also clear the trailing-whitespace/indentation noise.

Regarding the sync execution of `log_execution`, I left out because If I execute that into a separate thread then there are out of order problems occuring in the existing tests, also `log_execution` is lightweight so it's might don't effect the async execution. That's why I left out like this.

santoshkumarradha (dismissed):
Thanks for iterating on this. I re-checked the latest branch state and the main blocker now is scope: this PR is still far too broad for issue #622. It now spans a few hundred files across Python, Go, TypeScript, desktop, changelog, release config, examples, and docs, which makes it impossible to review or merge safely as a fix for one async fire-and-forget bug.

Please split this down to the issue-scoped change set: the non-blocking notification path, the tests that prove it, and any minimal follow-on needed for that exact behavior. Everything else should move to separate PRs. Once the surface is small enough to validate end to end, I’m happy to take another pass.

DebanKsahu:
Sure, most of these unnecessary changes might be lint changes. I will remove them for now on.

AbirAbbas:
Coming back to your question on the two approaches — what you actually shipped is better than either of them, and it's the right call. Neither original option would have guaranteed ordering: `create_task` FIFO only decides when tasks *start*, and the moment each notify awaits its POST, completion order becomes network order — a start POST can easily resolve after the complete POST. Serializing one consumer over a queue is what actually restores the guarantee, so the dispatcher design is the one I'd have landed on too. For what it's worth, the control plane also has a terminal-state regression guard on the events endpoint, so we're covered from both sides.

I pushed a few commits on top to finish this off. Behavior tests for the actual #622 contract: a slow fake control plane proving `execute_with_tracking` returns without waiting on the telemetry POSTs, ordering under concurrent executions including a failing one, and cleanup draining the queue before the HTTP client closes. A lazy-start in `submit()`, so paths that never run the server lifespan — e.g. mounting the app in uvicorn directly — don't silently drop telemetry (verified live against a running control plane: pre-fix the mounted app returned results but zero events reached the CP; with the fix they all arrive). CLI `call` mode turns out to be unaffected either way — it unwraps to the original function and never had telemetry by design. And small cleanups: typos, an unused helper, a couple of format nits.

Leaving `log_execution` synchronous is fine as scoped in the description — its control-plane forwarding already goes through `create_task`, so what remains is cheap. One tradeoff worth stating explicitly: a single serial consumer means one slow POST delays everything queued behind it, and the queue is unbounded under sustained load. That's acceptable for this fix; a bounded queue or per-execution keying would make a sensible follow-up issue rather than a blocker here.

AbirAbbas (approved):

PR #746, opened 2026-07-09
fix(sdk/go): claude prompt via stdin; note endpoint /api/v1 prefix

AbirAbbas:
## Summary

Two bugs found by live E2E testing of the SWE-AF Go port against a real control plane:

1. **Critical — claude provider prompt swallowed.** The harness appended the prompt as a trailing positional after repeated `--allowedTools` flags; claude CLI 2.1.x treats `--allowedTools` as variadic and consumes the positional, so **every tool-passing call** failed with `Input must be provided either through stdin or as a prompt argument when using --print`. The prompt now flows via stdin (`RunCLIWithStdin`), which `--print` reads natively. codex/opencode/gemini were audited: their arg constructions have no variadic-before-positional hazard. Verified against the real CLI via a build-tagged integration test (`AGENTFIELD_CLAUDE_BIN`-gated) — exit 0 with real structured output on the exact tools+stdin path.
2. **Minor — `Note()` 404.** Posted to `{base}/executions/note` without the `/api/v1` prefix (route is registered under the `/api/v1` group in `routes_core.go:150`), silently dropping every agent progress note. Existing tests masked it by baking `/api/v1` into the base URL contrary to the SDK-wide bare-base contract; tests corrected to the real contract.

## Test plan
- [x] `go mod tidy` no diff; `go build ./...`; `go vet ./...` (incl. `-tags integration`)
- [x] `go test ./...` — all packages green
- [x] New unit tests: prompt arrives on stdin and absent from the arg vector; note URL asserts `/api/v1/executions/note` from a bare base
- [x] Gated integration test PASSES against the real claude CLI

---


AbirAbbas:
Consolidated into #750 — the same commits (7c6f11e1, 18c3f3ea, 5b763d87) are cherry-picked there.

PR #757, opened 2026-07-12
fix(agentfield): add test for issue #435

### LINKED ISSUE #435: [Go SDK] WithDIDAuth silently disables signing on JWK parse failure
## Summary

`WithDIDAuth` logs a warning and silently constructs a client without DID signing when JWK parsing fails, so a caller that explicitly requested DID-authenticated traffic unknowingly sends unsigned requests.

## Context

`client.go:57` in the Go SDK handles JWK parse failure inside `NewDIDAuthenticator` by calling `log.Printf("WARNING: DID auth disabled due to JWK parse error: %v", err)` and returning from the option function — leaving `didAuthenticator == nil`. The `New(...)` call succeeds, the caller gets back a valid client, and there is no error to check. Any subsequent request that should have been DID-signed is sent unsigned. This is a security regression: the caller opted into signing to ensure requests are cryptographically attributed, but got no signing and no error to act on.

### In Scope

- Propagate the JWK parse error out of `WithDIDAuth` so the caller can detect and handle it.
- Two acceptable approaches: (a) change `Option` to `func(*Client) error` and return the error from `New(...)`; or (b) store the error inside the client struct and return it from the first call to any method that requires signing.
- Either way, the client must NOT be usable for DID-signed requests when the authenticator failed to initialize.

### Out of Scope

- Changing the JWK parsing logic itself.
- Retrying JWK loading — the key material should be valid at construction time.
- Changing the behavior for non-DID options that cannot fail.

## Files

- `sdk/go/client/client.go:57` — make `WithDIDAuth` propagate JWK parse errors rather than swallowing them
- `sdk/go/client/client.go` — update `New(...)` signature or add deferred-error pattern as appropriate
- `sdk/go/client/client_test.go` — test: `New(WithDIDAuth(badJWK))` returns an error (or first signed call returns an error); `New(WithDIDAuth(goodJWK))` succeeds

## Acceptance Criteria

- [ ] Passing an invalid JWK to `WithDIDAuth` results in a non-nil error that the caller can inspect
- [ ] A client constructed with a failed `WithDIDAuth` does not silently send unsigned requests
- [ ] Passing a valid JWK continues to work as before

## Notes for Contributors

**Severity**: HIGH

The `Option` → `func(*Client) error` pattern is used by many popular Go client libraries (e.g. `google.golang.org/grpc`). If the codebase already has many callers of `client.New(...)` that ignore errors, prefer the deferred-error approach (store error, return on first use) to minimize the blast radius of the signature change. Either way, remove the `log.Printf` warning — a silent log is not an acceptable substitute for a returned error in a security-critical path.

jaythehardcoder:
## What
Adds test coverage for the behavior described in issue #435.

## Why
Fixes #435

## Testing
- [x] New test added


AbirAbbas:

PR #760, opened 2026-07-13
fix(agentfield): add test for issue #438

### LINKED ISSUE #438: [Go SDK] OpenCode harness has no concurrency guard

jaythehardcoder:
## What
Adds test coverage for the behavior described in issue #438.

## Why
Fixes #438

## Testing
- [x] New test added


AbirAbbas:

PR #775, opened 2026-07-15
fix(agentfield): add test for issue #587

### LINKED ISSUE #587: ImageOutput.save() crashes on data: URLs returned by Gemini image models
## Summary

`ImageOutput.save()` in `sdk/python/agentfield/multimodal_response.py` ~line 99 calls `requests.get(self.url)` unconditionally on the URL branch, which raises `InvalidSchema` for `data:` URLs.

`vision.generate_image_openrouter` populates `ImageOutput.url` with whatever the OpenRouter response carries. For Gemini image models (e.g. `google/gemini-2.5-flash-image`), that value is a `data:image/png;base64,...` URL, not an HTTP URL.

## Repro

```python
from agentfield.multimodal_response import ImageOutput
img = ImageOutput(url="data:image/png;base64,iVBORw0KGgoAAAA...")
img.save("/tmp/out.png")
```

## Error

```
requests.exceptions.InvalidSchema: No connection adapters were found for
'data:image/png;base64,iVBORw0KGgo...'
```

## Suggested fix

In either:
- `generate_image_openrouter` — when the response's `url` is a `data:` URL, base64-decode it once and populate `b64_json` instead, leaving `url=None`. Then the existing `b64_json` branch of `save()` handles it.
- `ImageOutput.save()` / `ImageOutput.get_bytes()` — detect `data:` URLs and decode locally before falling through to `requests.get`.

## Status

This is fixed on `origin/main` — `get_bytes()` now special-cases `data:` URLs and `save()` delegates to `get_bytes()`. The bug is still present in the latest released version `agentfield==0.1.84`, which is what surfaced this. Filing for visibility once an 0.1.85 release cuts so consumers can drop their workarounds.

## Consumer workaround

reel-af ships a `_save_image_output()` helper plus a runtime monkey-patch (`src/reel_af/sdk_patches.py::_patch_image_output_save`) that sniffs `data:` URLs and decodes them itself before delegating back to the SDK for real HTTP URLs.

jaythehardcoder:
## What
Adds test coverage for the behavior described in issue #587.

## Why
Fixes #587

## Testing
- [x] New test added

### DIFF
--- a/tests/test_issue_587.py
+++ b/tests/test_issue_587.py
@@ -1,0 +1,3 @@
+from agentfield.multimodal_response import ImageOutput
+img = ImageOutput(url="data:image/png;base64,iVBORw0KGgoAAAA...")
+img.save("/tmp/out.png")


tests/test_issue_587.py
```diff
@@ -0,0 +1,3 @@
+from agentfield.multimodal_response import ImageOutput
```
santoshkumarradha, line 1: This file is just executable snippet code at module import time, not a pytest test. Please move this into the package’s exercised test suite and add an assertion that reproduces the crash and proves the fix.

santoshkumarradha (changes requested):
Thanks for taking a pass at this. This needs to be turned into a real regression test before it’s mergeable.

jaythehardcoder:
@santoshkumarradha thanks for taking a look — you're right that this needs to be a real pytest regression test rather than module-level executable code. I'll move it into the package's existing test infrastructure as a proper async test. Will push the update shortly.

AbirAbbas:

PR #777, opened 2026-07-15
fix(agentfield): add test for issue #584

### LINKED ISSUE #584: OpenRouterProvider.generate_audio rejects format='wav' under hardcoded stream=true
## Summary

`OpenRouterProvider.generate_audio` in `sdk/python/agentfield/media_providers.py` hardcodes `"stream": True` in the chat-completions audio payload. OpenRouter rejects every format except `pcm16` when `stream=true`, so `format="wav"` (and `mp3`, `flac`, `opus`) all 400 against models in the gpt-audio family.

The SDK signature advertises `format` from `{"wav", "mp3", "flac", "opus", "pcm16"}`, so callers reasonably expect any of those values to work.

## Repro

```python
await OpenRouterProvider().generate_audio(
    text="Hello world",
    model="openrouter/openai/gpt-audio-mini",
    format="wav",
)
```

## Error

```
RuntimeError: OpenRouter audio request failed (400):
{"error":{"message":"Unsupported value: 'audio.format' does not support
'wav' when stream=true. Supported values are: 'pcm16'."}}
```

## Suggested fix

Either:
- auto-switch `stream=false` when `format != "pcm16"`, or
- request `pcm16` over the wire when the caller asks for `wav` and wrap the streamed pcm16 in the requested container client-side before returning.

## Status

This was discovered against `agentfield==0.1.84` while building the `reel-af` example. Looking at `origin/main`, the second strategy appears to already be implemented (see `_wrap_pcm16_as_wav_b64` in `media_providers.py` and the `wire_format = "pcm16" if audio_format == "wav" else audio_format` branch around line 1382). Filing for tracking so a release can cut and consumers can drop their workarounds.

jaythehardcoder:
## What
Adds test coverage for the behavior described in issue #584.

## Why
Fixes #584

## Testing
- [x] New test added

### DIFF
--- a/tests/test_issue_584.py
+++ b/tests/test_issue_584.py
@@ -1,0 +1,6 @@
+from agentfield.media_providers import OpenRouterProvider
+await OpenRouterProvider().generate_audio(
+    text="Hello world",
+    model="openrouter/openai/gpt-audio-mini",
+    format="wav",
+)


tests/test_issue_584.py
```diff
@@ -0,0 +1,6 @@
+from agentfield.media_providers import OpenRouterProvider
+await OpenRouterProvider().generate_audio(
```
santoshkumarradha, line 2: This is another top-level `await` snippet rather than a test the suite can execute and verify. Please move it into the relevant SDK test module, wrap it in an async test, and assert the behavior you want to lock in.

santoshkumarradha (changes requested):
Thanks for the contribution. This needs one more pass to become a real regression test.

jaythehardcoder:
@santoshkumarradha thanks for the review — you're right that the top-level `await` snippet isn't a proper pytest regression test. I'll rewrite it as a proper async test with `pytest.mark.asyncio` and wire it into the existing test suite. Will push the update shortly.

AbirAbbas:

PR #783, opened 2026-07-16
feat(python-sdk): add MiniMax video provider routing

octo-patch:
## Summary

Adds a direct MiniMax video provider to the Python SDK media router. The provider submits asynchronous generation tasks, polls their status, retrieves the generated file URL, and supports configurable global and China API bases while leaving existing video defaults unchanged.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs only
- [ ] Tests only
- [ ] CI / tooling
- [ ] Breaking change

## Test plan

- [x] `sdk/python/.venv/bin/ruff check sdk/python`
- [x] `sdk/python/.venv/bin/pytest -c sdk/python/pyproject.toml sdk/python/tests`

## Test coverage

- [x] I ran tests for the Python SDK locally.
- [x] New request, routing, validation, and error paths are covered by tests.
- [x] No coverage baseline changes were needed.

## Checklist

- [x] I have read `docs/CONTRIBUTING.md` and `docs/DEVELOPMENT.md`.
- [ ] I have linked any related issues.

## Related issues / PRs


sdk/python/tests/test_minimax_video.py
```diff
@@ -0,0 +1,228 @@
...
+    monkeypatch.setattr("aiohttp.ClientSession", lambda **kwargs: error_session)
+    provider = MiniMaxProvider(api_key="unit-value")
+
+    with pytest.raises(RuntimeError, match="authentication failed"):
+        await provider.generate_video(
+            prompt="A landscape",
+            model="minimax/video-model",
+            poll_interval=0,
+        )
+    assert error_session.calls[0][1] == f"{MINIMAX_GLOBAL_BASE_URL}/video_generation"
```
AbirAbbas, line 140: This assertion makes the test environment-sensitive: the provider is constructed at line 131 without an explicit `base_url`, and `MiniMaxProvider.__init__` falls back to `os.environ["MINIMAX_BASE_URL"]` before `MINIMAX_GLOBAL_BASE_URL` (media_providers.py lines 665-667). Since the README tells users to export exactly that variable, the test fails on any machine/CI where it's set.

Suggest isolating it at the top of the test:

```python
monkeypatch.delenv("MINIMAX_BASE_URL", raising=False)
```

(or pass `base_url=MINIMAX_GLOBAL_BASE_URL` explicitly at line 131).

sdk/python/agentfield/media_providers.py
```diff
@@ -646,6 +651,227 @@ async def transcribe_audio(
...
+        if duration is not None:
+            if not float(duration).is_integer():
+                raise ValueError("MiniMax video duration must be a whole number")
+            body["duration"] = int(duration)
+        if resolution:
+            body["resolution"] = resolution.upper()
+        if extra:
+            body.update(extra)
+        if kwargs:
+            body.update(kwargs)
```
AbirAbbas: These two `body.update(...)` calls run after the validated fields are set, so `extra` or stray `kwargs` can silently overwrite `model`, `prompt`, `duration`, or `resolution` — e.g. `extra={"duration": 3.5}` bypasses the whole-number check at lines 767-770, and `extra={"resolution": "1080p"}` skips the `.upper()` normalization.

Consider guarding reserved keys before merging:

```python
reserved = body.keys() & {**(extra or {}), **kwargs}.keys()
if reserved:
    raise ValueError(f"extra/kwargs may not override: {sorted(reserved)}")
```

(or document that `extra` intentionally wins, and merge it before validation so the checks still apply).

sdk/python/agentfield/media_providers.py
```diff
@@ -646,6 +651,227 @@ async def transcribe_audio(
...
+    def _strip_prefix(model: str) -> str:
+        return model[len("minimax/") :] if model.startswith("minimax/") else model
+
+    @staticmethod
+    def _check_api_response(payload: Dict[str, Any], operation: str) -> None:
+        base_resp = payload.get("base_resp") or {}
+        status_code = base_resp.get("status_code")
+        if status_code not in (None, 0):
+            status_msg = base_resp.get("status_msg") or "unknown error"
+            raise RuntimeError(
```
AbirAbbas, line 708: Non-blocking consistency note: MiniMax wraps all API failures in bare `RuntimeError` (here, in `_read_response`, and the missing-`task_id` case), plus `ValueError` for validation and `TimeoutError` for polling — while `FalProvider` lets native client exceptions propagate. Callers of `ai_generate_video` end up needing provider-specific error handling.

Worth considering a shared typed exception for media-provider failures (or matching Fal's propagation behavior) so errors are catchable uniformly across providers.

sdk/python/README.md
```diff
@@ -83,6 +83,45 @@ result = await app.ai(
...
+## MiniMax Video Generation
+
+Set an API key and choose a video model from the MiniMax video API documentation:
+
+```bash
+export MINIMAX_API_KEY="..."
+export MINIMAX_VIDEO_MODEL="..."
+```
+
+The global API base is used by default. Set `MINIMAX_BASE_URL` to select a region:
```
AbirAbbas, line 95: Worth a one-liner on precedence here: `AIConfig(minimax_base_url=...)` silently overrides `MINIMAX_BASE_URL` (agent_ai.py passes the config value as `base_url`, which wins in the provider's `base_url or env or default` chain — same for `minimax_api_key` vs `MINIMAX_API_KEY`). As written, a reader would expect the env var to take effect when both are set.

AbirAbbas (commented):
Thanks for the contribution — the provider implementation and test coverage of the submit/poll/retrieve flow look solid. Ran this through PR-AF (our review agent) and hand-verified its findings against the diff; nothing blocking, but four items worth addressing inline below. Separately: the CLA check is still unsigned, which will block merge regardless.

AbirAbbas:
@octo-patch no movement here since the review on the 16th, so I'm taking this over the finish line — I'll push the fixes for the open review items directly to this branch (maintainer edits are enabled). Your commits and authorship stay intact.

AbirAbbas:
Pushed the review items from the 16th:

- `extra`/`kwargs` can no longer override the validated request fields (`model`, `prompt`, `first_frame_image`, `duration`, `resolution`) — they now raise a `ValueError` instead of silently bypassing the duration/resolution checks. Regression tests cover the `extra`, `kwargs`, and `model` cases; non-reserved passthrough keys like `prompt_optimizer` still merge as before.
- The error-path test no longer inherits `MINIMAX_BASE_URL` from the environment.
- README now notes that `AIConfig(minimax_api_key=..., minimax_base_url=...)` wins over the env vars when both are set.

The error-type consistency point (bare `RuntimeError` vs Fal's propagation) was flagged as non-blocking and is a wider media-provider refactor, so I'm leaving it for a follow-up rather than growing this PR.

Full sdk/python suite is green locally (1728 passed) and CI runs are approved.

AbirAbbas (approved):
Taking this the rest of the way per the takeover note — review items addressed, full suite green locally and in CI.

AbirAbbas (approved):
Re-approving after the branch updates — same content verified: local full suite + CI green.

PR #789, opened 2026-07-17
Add MiniMax music generation provider

octo-patch:
Reason: Add MiniMax music generation to the media provider registry.

Implemented global and China endpoint routing, Bearer authentication, request controls, output formats, and response parsing for MiniMax music generation.

Checks: python3 -m py_compile sdk/python/agentfield/media_providers.py; PYTHONPATH=sdk/python pytest -q sdk/python/tests/test_media_providers.py -q


sdk/python/agentfield/media_providers.py
```diff
@@ -152,6 +152,47 @@ async def generate_music(
...
+        if duration is not None:
+            body["audio_setting"] = {"sample_rate": kwargs.pop("sample_rate", 44100), "bitrate": kwargs.pop("bitrate", 256000), "format": kwargs.pop("format", "mp3"), "duration": duration}
+        body.update({k: v for k, v in kwargs.items() if k in {"lyrics", "lyrics_optimizer", "is_instrumental", "audio_setting", "aigc_watermark", "cover_feature_id"}})
+        async with aiohttp.ClientSession() as session:
+            async with session.post(endpoint, json=body, headers={"Authorization": f"Bearer {key}", "Content-Type": "application/json"}) as response:
+                data = await response.json()
+                if response.status >= 400 or data.get("base_resp", {}).get("status_code") != 0:
+                    raise RuntimeError(f"MiniMax music generation failed ({response.status}): {data}")
+        audio = data.get("data", {}).get("audio")
+        output = AudioOutput(data=audio if audio and body["output_format"] == "hex" else None, format=body.get("audio_setting", {}).get("format", "mp3"), url=audio if body["output_format"] == "url" else None)
```
santoshkumarradha: `AudioOutput.data` is base64-encoded audio bytes everywhere else in the SDK, but this branch stores MiniMax's raw hex string directly. That means helpers like `AudioOutput.get_bytes()` and `.save()` will try to base64-decode hex and produce the wrong bytes or fail.

Can we decode the hex payload to bytes here and then re-encode it as base64 before constructing `AudioOutput`?

santoshkumarradha (dismissed):
Thanks for the contribution. The provider wiring is straightforward, but there’s one SDK contract break that needs fixing before this is safe to merge.

Please also add tests for the new music path, especially the `output_format="url"` and `output_format="hex"` branches. Right now the PR description mentions local checks, but there is no repo test coverage for this provider yet.

Separately, the CLA check is still pending, so that will need to be signed before we can merge.

octo-patch:
Thanks for the review. I now decode MiniMax hex audio bytes and re-encode them as base64 before constructing AudioOutput, preserving the SDK contract. I also added URL and hex regression tests and pushed a new commit; the focused provider tests pass (7 passed).

octo-patch:
Thanks for the review. The requested audio decoding and URL/hex tests are already in the latest commit; the remaining CLA check still requires contributor action.

AbirAbbas:
Re-reviewed the latest push. The contract break santosh flagged is fixed: `AudioOutput.data` is documented as base64, and the new code decodes MiniMax's hex payload and re-encodes it as base64 before constructing the output — verified against `multimodal_response.py`. The url/hex regression tests cover both branches. So the requested changes look addressed.

One structural problem before this can land, though: #783 adds a *video* `MiniMaxProvider` to the same module. The two PRs merge cleanly in git (different regions of the file), but Python will silently rebind the name — whichever class is defined later in the file wins, and the other becomes unreachable through both imports and `get_provider("minimax")`. As written, merging both leaves music generation dead.

Since #783 is older and now green, the plan is: #783 lands first, then this PR rebases to fold `generate_music` into the existing `MiniMaxProvider` (add `"music"` to `supported_modalities`) instead of defining a second class. Happy to do that rebase here if there's no movement once #783 merges.

AbirAbbas:
Took this the rest of the way per the plan above, now that #783 is merged:

- `generate_music` now lives on the unified `MiniMaxProvider` (the video class) — the branch briefly carried two same-named classes after merging main, which Python resolves by silently discarding the earlier one, leaving music unreachable through both imports and `get_provider("minimax")`.
- `supported_modalities` is `["video", "music"]`, and the music endpoint derives from the same `base_url` / `MINIMAX_BASE_URL` configuration as video and `AIConfig` (replacing the never-released `region` kwarg).
- The duplicate `"minimax"` registry entry is gone, the url/hex contract tests are unchanged in behavior, and there's a new endpoint-routing test for a configured base URL.
- Full sdk/python suite green locally (1,828 passed).

@santoshkumarradha your changes-requested is the last gate — the hex→base64 contract fix you asked for is in and verified against `AudioOutput`.

AbirAbbas (approved):

PR #799, opened 2026-07-20
fix(sdk/python): resolve ResultCache cross-loop deadlock (#623)

### LINKED ISSUE #623: [Python SDK] ResultCache can deadlock the process

7vignesh:
## Summary

`ResultCache` mixed a `threading.RLock` with loop-bound asyncio primitives (an `asyncio.Event` for shutdown and an `asyncio.Task` for the cleanup loop). When `start()` and `stop()` ran on different event loops which happens when the client's sync and async execution paths are mixed (#620) `stop()` raised `got Future attached to a different loop` and could wedge the process. This makes the cache lifecycle loop-aware so it never awaits a task from the wrong loop.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs only
- [ ] Tests only
- [ ] CI / tooling
- [ ] Breaking change

## Test plan

- `cd sdk/python && python -m pytest tests/test_result_cache_deadlock.py -v`
- `cd sdk/python && python -m pytest tests/test_result_cache.py tests/test_result_cache_bigfiles_coverage.py`
- Reproduced the deadlock before the fix (`RuntimeError: got Future attached to a different loop`) and confirmed cross-loop `stop()` now completes cleanly with no lingering pending-task warnings.

## Test coverage

`result_cache.py` coverage went from 88% to 95%; the sdk-python aggregate stays at 94% (baseline 93.73%). New `tests/test_result_cache_deadlock.py` covers cross-loop stop, idempotent/rebinding start, concurrent sync access during cleanup, stop-without-start, and the disabled-cache no-op path.

## Checklist

## Related issues / PRs

Fixes #623
Related to #620 (mixing threads and async code)


sdk/python/agentfield/result_cache.py
```diff
@@ -153,26 +161,109 @@ async def __aexit__(self, exc_type, exc_val, exc_tb):
...
+                # don't let teardown wedge the caller.
+                logger.debug(
+                    "Cleanup task await raised RuntimeError during stop",
+                    exc_info=True,
+                )
+        elif task is not None:
+            # Cross-loop stop: cancel on the owning loop, never await here.
+            self._discard_stale_cleanup_task()
+
+        self._cleanup_task = None
```
AbirAbbas: One residual race here: `stop()` snapshots these fields at the top, awaits the task in between, and then unconditionally nulls all three — with no lock on either side. A `start()` racing in from another thread (or even on the same loop during the `await task`) can install a fresh task that gets clobbered here and leaks, still running on its loop. The serialized cross-loop case — the #623 scenario — is handled, and the old code was just as unguarded, so not blocking. But since the instance already has `self._lock`, it'd be cheap to take it around the snapshot and around this mutation block (never across the await).

AbirAbbas (commented):
I took this for a spin rather than just reading the diff. On the merge-base I reproduced the failure with `start()` on a background thread's loop and `stop()` via `asyncio.run()` on the main thread: `RuntimeError: got Future attached to a different loop`, and since `stop()` aborts mid-way the cache is left populated. On this branch the same scenario (and the owning-loop-closed variant) completes cleanly, and I checked that the stale cleanup task really does get cancelled on its owning loop rather than leaked. I also ran the new test file against the merge-base — 3 of the 8 genuinely fail pre-fix, so these are real regression tests rather than mirrors of the implementation. Full suite on the branch is green for me locally (1767 passed, 4 skipped) and ruff is clean. Nice touch that `stop()` now clears the cache even on the cross-loop path — before, the failed stop left the contents behind.

The one thing I'd like sorted before this closes #623: the same foreign-loop `task.cancel(); await task` pattern still lives in `AsyncExecutionManager.stop()` (async_execution_manager.py:337-340) and `ConnectionManager.close()` (connection_manager.py:99-102). I verified on this branch that starting the manager on one loop and stopping it from another still raises the identical RuntimeError at async_execution_manager.py:339 — before `result_cache.stop()` is ever reached. Since `client.aclose()` goes through `manager.stop()`, the end-to-end sync/async mixing case from #620/#623 still fails one level up. Happy with either extending the same loop-aware treatment to those two sites in this PR or a follow-up, but either way I'd change "Fixes #623" to "Partially addresses #623" so the issue doesn't auto-close on merge.

7vignesh:
Good catch you're right that the same foreign-loop `task.cancel(); await task` pattern lived one level up and `client.aclose()` would hit it before `result_cache.stop()` was ever reached. I extended the fix rather than deferring it.

What changed in the latest push:

- New `agentfield/async_lifecycle.py` holds the loop-aware teardown pattern as shared helpers (`current_running_loop`, `cancel_task_cross_loop`, `cancel_and_await_if_same_loop`) so all three sites share one implementation instead of three copies.
- `ResultCache` now uses those helpers (behaviour unchanged).
- `AsyncExecutionManager.stop()` (async_execution_manager.py:337-340) records its owning loop at `start()` and only awaits the four background tasks / sets the shutdown Event when on that loop; cross-loop it cancels without awaiting.
- `http_connection_manager.ConnectionManager.close()` (the site you flagged) records its owning loop at `start()`. On the owning loop it takes the async lock and awaits the session close as before; on a foreign loop it skips the loop-bound lock and schedules the aiohttp session/connector close on the owning loop via `call_soon_threadsafe` without awaiting.

I verified the end-to-end path: starting the manager on a background thread's loop and stopping it from another no longer raises at async_execution_manager.py:339, and `client.aclose()` from a foreign loop now completes cleanly through both managers down to the cache.

New `tests/test_async_lifecycle_deadlock.py` covers cross-loop teardown of both components 3 of its 4 checks genuinely fail on the merge-base, so they're real regression tests. Full result_cache + manager + connection suite is green locally (110 passed) and ruff is clean.

Since the whole `client.aclose()` chain is now loop-safe, I've kept this as "Fixes #623". Happy to revisit if you'd still prefer to scope it as partial.

7vignesh:
@santoshkumarradha pls review

7vignesh:
@AbirAbbas review

sdk/python/agentfield/async_lifecycle.py
```diff
@@ -0,0 +1,94 @@
...
+    if task is None:
+        return
+
+    if owning_loop is current_running_loop():
+        task.cancel()
+        try:
+            await task
+        except asyncio.CancelledError:
+            pass
+        except RuntimeError:
```
AbirAbbas, line 85: This catches more than the loop-association error: a `RuntimeError` raised by the task's own cancellation cleanup (an `__aexit__`, say) is swallowed at debug level, where the old code only suppressed `CancelledError`. Worth matching on the message (`attached to a different loop`) or at least logging at warning so a failed teardown isn't invisible.

sdk/python/agentfield/async_execution_manager.py
```diff
@@ -333,17 +357,13 @@ async def stop(self) -> None:
...
+            await cancel_and_await_if_same_loop(task, owning_loop)
 
         self._polling_task = None
         self._cleanup_task = None
         self._metrics_task = None
         self._event_stream_task = None
+        self._loop = None
 
         # Cancel all active executions
         async with self._execution_lock:
```
AbirAbbas: The task cancellation above is loop-aware now, but this `async with self._execution_lock` still runs on whatever loop called `stop()`. Uncontended it happens to work, but if the owner loop is holding the lock the waiter future binds to the foreign loop and we're back to `got Future attached to a different loop` — one line below the fix. The cross-loop path needs to skip or marshal this section, like the connection manager does.

sdk/python/agentfield/http_connection_manager.py
```diff
@@ -214,8 +236,58 @@ async def close(self) -> None:
                 await self._connector.close()
                 self._connector = None
 
+            self._loop = None
             logger.info("ConnectionManager closed")
 
+    def _close_cross_loop(self, owning_loop: asyncio.AbstractEventLoop) -> None:
```
AbirAbbas, line 242: Two notes on this path, both fine as follow-ups: (1) it mutates `_closed`/`_session`/`_connector` from the foreign thread without the lock, so an in-flight `get_session()` on the owner loop can read `_session` as `None` mid-request and die with `AttributeError` — marshaling the whole teardown to the owning loop would close that window; (2) when the owning loop is already closed (the common sequential `asyncio.run()` case) the session/connector are silently dropped and aiohttp's unclosed-session warnings fire anyway, while this logs `closed (cross-loop)` at info and buries the detail at debug. A warning there would make the leak visible.

sdk/python/agentfield/result_cache.py
```diff
@@ -153,26 +166,79 @@ async def __aexit__(self, exc_type, exc_val, exc_tb):
...
+        owning_loop = self._loop
+
+        # Signal the loop to exit promptly when we're on its own loop; the
+        # shared helper then cancels + awaits it there. Cross-loop, the helper
+        # cancels on the owning loop without awaiting.
+        if shutdown_event is not None and owning_loop is current_loop:
+            shutdown_event.set()
+        await cancel_and_await_if_same_loop(task, owning_loop)
+
+        self._cleanup_task = None
```
AbirAbbas: These clears run unconditionally after an `await`, so a concurrent `start()` on another loop that installs a fresh cleanup task in that window gets its references nulled without the new task being cancelled — orphan cleanup loop, cache looks stopped. Compare-before-clear (`if self._cleanup_task is task: ...`, same for the event and loop) closes it.

AbirAbbas (commented):
Went through the latest revision and hand-verified everything against the diff (review tooling first, then manually). The core loop-aware teardown is right — I couldn't break the main paths, and the earlier repro stays fixed. What's left is edge-case hardening: two items are squarely in scope for this PR and cheap (the execution-lock note and the ResultCache stop/start race, inline below); the other two are fine as follow-ups if you'd rather keep this focused.

AbirAbbas:
Heads up: main just picked up a ruff pin in `.github/workflows/sdk-python.yml` (#831), and this branch touches the same file — it's now showing a merge conflict. Worth resolving when you pick up the review comments; keep the `ruff==0.15.22` pin from main.

7vignesh:
Thanks for the thorough review @AbirAbbas! Just pushed the two in-scope fixes:

1. ResultCache stop/start race: switched to compare-before-clear (`if self._cleanup_task is task: ...`) so a concurrent start() that installs fresh references during the await doesn't get its new task orphaned.

2. AsyncExecutionManager execution lock: the cross-loop path now skips the `async with self._execution_lock` section entirely, same pattern as the connection manager. The executions will be cancelled when the owning loop tears down or the manager restarts.

For the other two (RuntimeError catch breadth in async_lifecycle.py and the connection manager's lockless mutation on cross-loop close), I'll address those in a follow-up once this lands. They're on my radar.

Also already resolved the merge conflict from the ruff pin in the previous push.

7vignesh:
Hey @AbirAbbas, resolved the merge conflict. Rebased the branch cleanly on top of current main so the diff only shows the actual fix commits now, no release-bot noise. The two review fixes you asked for are still there (compare-before-clear + skip execution lock cross-loop). Should be good to go once CI passes.

7vignesh:
@santoshkumarradha @AbirAbbas ready for review when you get a chance. CI is green and all feedback has been addressed.

santoshkumarradha (commented):
Checked the latest revision and I agree the original cross-loop crash is fixed, but I’m not ready to queue this one yet. Two semantics still feel too loose for a teardown path that callers will read as definitive stop behavior: on a foreign-loop stop, `AsyncExecutionManager.stop()` still leaves active executions running until the old loop happens to tear down, and `ConnectionManager.close()` still mutates shared session/connector state without serializing that transition on the owning loop’s lock. If you tighten those two paths so cross-loop stop/close still schedules the real teardown work on the owning loop, I’m happy to take another pass quickly.

santoshkumarradha:
@AbirAbbas

7vignesh:
@santoshkumarradha addressed both points in the latest push:

1. AsyncExecutionManager: cross-loop stop now schedules the execution cancellation (under `_execution_lock`) on the owning loop via `call_soon_threadsafe`, so active executions are actually terminated rather than left dangling.

2. ConnectionManager: `_close_cross_loop` no longer mutates `_session`/`_connector` from the foreign thread. Only `_closed` is set immediately (to gate new requests). The session/connector close and state nulling now run inside a coroutine scheduled on the owning loop under its async lock, so `get_session()` callers on the owning loop never see half-torn-down state.

Verified both paths end-to-end: cross-loop stop completes without error and the owning loop processes the real teardown (session is None after the loop runs). Tests all pass (45 passed).

AbirAbbas (approved):
Both fixes from the last round are in and verified — the compare-before-clear in ResultCache.stop(), and the cross-loop stop() path now rescheduling cancellation onto the owning loop (stronger than the skip I asked for, and the right call: I probed it live and executions get cancelled with capacity released, on the loop that owns the lock).

Beyond reading the diff: the new deadlock tests fail on main with the exact `got Future attached to a different loop` error in all three components, so they genuinely guard the regression; 80+ stress runs of the two test files (including 8-way CPU saturation and 3.12) produced zero flakes; and the full suite + ruff 0.15.22 are green across 3.10/3.11/3.12. The workflow conflict resolution kept main's ruff pin intact.

The two follow-ups we agreed on earlier (ConnectionManager foreign-thread teardown hardening, narrowing the broad RuntimeError suppression in async_lifecycle) remain follow-ups — nothing new blocks. Thanks for sticking with this one through the long review.

sdk/python/agentfield/async_lifecycle.py
AbirAbbas: Resolving — tracked as a follow-up in #901.

sdk/python/agentfield/http_connection_manager.py
AbirAbbas: Resolving — tracked as a follow-up in #901.

PR #817, opened 2026-07-22
feat(control-plane): embedded MCP server at /mcp (zero-setup harness integration)

AbirAbbas:
## What & why

AgentField had no MCP server — the only way for an AI harness to drive installed
agents was the `af` CLI plus the `agentfield-use` skill. This PR ships an
**MCP server embedded in the control plane itself, served on the same port as
the REST API**. Whether the control plane was started by the CLI (`af server`)
or the desktop app, MCP is just there — zero extra processes, zero setup. A
harness connects with one command:

```bash
claude mcp add --transport http agentfield http://localhost:8080/mcp
```

It is **on by default**, with an env kill-switch: `AGENTFIELD_MCP_ENABLED=false`
makes `/mcp` return `404`. The endpoint sits behind the same global API-key auth
and trust domain as `/api/v1`, and the tools call the existing discovery /
execution service layer directly (no loopback HTTP).

- **Transport:** streamable HTTP, JSON-RPC 2.0 over `POST /mcp` (stateless;
  `Mcp-Session-Id` / `MCP-Protocol-Version` headers tolerated and ignored).
  `GET /mcp` → 405, `OPTIONS /mcp` → 204.

## The five tools

| Tool | Purpose |
|---|---|
| `discover_agents` | List agents + reasoners (id, description, entrypoint tags). Defaults to active-only; `health:"all"` includes unhealthy. |
| `get_reasoner_schema` | Input/output JSON Schema for a reasoner, as served by discovery. |
| `execute_reasoner` | Validate `node.reasoner` exists, start an **async** run, return `{run_id, status:"accepted"}`. |
| `get_run` | Run status + result/error + per-execution summaries. |
| `wait_run` | Server-side poll to terminal or timeout (default 60s, hard-capped at 120s); adds a `timed_out` flag. |

## Validation Contract → tests

- `initialize` returns protocolVersion + capabilities.tools + serverInfo →
  `TestMCP_InitializeAndToolsList` (also asserts unknown protocol falls back to
  `2025-06-18` and serverInfo carries the build version).
- `tools/list` returns exactly the 5 tools, each with a valid JSON Schema →
  `TestMCP_InitializeAndToolsList`.
- `discover_agents` default returns only active agents; `health:"all"` includes
  dead ones with `health_status` → `TestMCP_DiscoverAgents`.
- `execute_reasoner` bad/malformed target → tool error naming `discover_agents`;
  good target → `run_id` and a created execution → `TestMCP_ExecuteReasoner`.
- `get_run` aggregates status/result/error + per-execution summaries →
  `TestMCP_GetRun`.
- `wait_run` respects the timeout cap and returns `timed_out` on a never-finishing
  run (and returns immediately for a terminal run) → `TestMCP_WaitRun`.
- Unknown method → `-32601`; unknown tool → `-32602`; batch arrays rejected →
  `TestMCP_ProtocolMiscMethods`.
- `AGENTFIELD_MCP_ENABLED=false` → `/mcp` is `404`; default → enabled;
  `GET /mcp` → 405 → `TestMCPRoutes_DisabledReturns404`,
  `TestMCPRoutes_EnabledByDefault`.
- Config default-on + env override → `TestMCPConfig_IsEnabledDefault`,
  `TestApplyEnvOverrides_MCPEnabled`.

## Test results

- `go build ./...` — clean
- `go vet ./internal/config/... ./internal/handlers/... ./internal/server/...` — clean
- `go test` for the config, handlers, server (MCP) and skillkit packages — all pass
- `gofmt -l` on changed files — empty
- `golangci-lint` — no issues in any changed/new file

One pre-existing, environment-specific test (`TestStartAndStopCoverAdditionalBranches`,
admin gRPC listener bind) fails identically on clean `origin/main` in this sandbox
and is unrelated to this change.


control-plane/internal/handlers/mcp.go
```diff
@@ -0,0 +1,761 @@
...
+}
+
+// ---- shared helpers --------------------------------------------------------
+
+// startAsyncRun starts an asynchronous execution of target ("node.reasoner")
+// with the given input, mirroring the /execute/async handler but driven from
+// the MCP tool. It returns as soon as the job is enqueued.
+func (s *mcpServer) startAsyncRun(ctx context.Context, target string, input map[string]interface{}) (runID, execID string, err error) {
+	controller := newExecutionController(s.store, s.payloads, s.webhooks, s.timeout, s.internalToken)
+	plan, err := controller.prepareExecutionForTarget(ctx, target, ExecuteRequest{Input: input}, executionHeaders{}, "", "")
```
AbirAbbas: > [!CAUTION]
> **Must-fix before merge.** Bypassing DID authorization allows unauthorized execution of targets, a security vulnerability.

Route MCP execution through the same authorization and provenance preparation as REST.

POST /mcp calls toolExecuteReasoner → startAsyncRun without evaluating the access policy, so a caller can execute a target that REST would deny. It also passes empty executionHeaders and DIDs, dropping authenticated provenance.

<sub>🔴 `CRITICAL` · confidence 99%</sub>

<details><summary>Evidence</summary>

> Step 1: `setupRoutes` applies global middleware and then `registerMCPRoutes`; `registerMCPRoutes` registers `s.Router.POST("/mcp", handler)` at routes_mcp.go:49 without permission middleware. Step 2: `registerCoreRoutes` creates `/api/v1/execute` and conditionally applies `PermissionCheckMiddleware` before both REST execute routes at routes_core.go:93-115. Step 3: for REST, `executionController.prepareExecution` passes `readExecutionHeaders(ginCtx)`, `middleware.GetVerifiedCallerDID(ginCtx)`, and `middleware.GetTargetDID(ginCtx)` into `prepareExecutionForTarget` at execute.go:1385-1392; permission middleware evaluates the policy and aborts unmatched requests when `DefaultDeny` is true (permission.go:236-263). Step 4: MCP `toolExecuteReasoner` calls `startAsyncRun` after only discovery/target existence validation (mcp.go:378-395), and `startAsyncRun` calls `prepareExecutionForTarget(..., executionHeaders{}, "", "")` at mcp.go:488, then submits the job at mcp.go:500-506. Step 5: `prepareExecutionForTarget` persists session/actor/parent only from `executionHeaders` (execute.go:1494, 1523-1528), stores caller/target DID on the plan (execute.go:1577-1578), and downstream dispatch forwards those DID headers only when non-empty (execute.go:1799-1803). Thus MCP reaches execution with neither policy enforcement nor provenance. Existing focused MCP tests pass bare MCP routers or an unauthenticated test server; `go test ./internal/handlers -run '^TestMCP_' -count=1` and `go test ./internal/server -run '^TestMCPRoutes_' -count=1` passed, but do not exercise this enabled-authorization path. The existing REST default-deny test passes (`go test ./internal/server/middleware -run '^TestPermissionDefaultDenyOnBlocksUnmatched$' -count=1`).

</details>

control-plane/internal/handlers/mcp.go
```diff
@@ -0,0 +1,761 @@
...
+		version:       strings.TrimSpace(version),
+	}
+	if srv.version == "" {
+		srv.version = "dev"
+	}
+	return srv.handle
+}
+
+// handle is the POST /mcp entry point.
+func (s *mcpServer) handle(c *gin.Context) {
```
AbirAbbas, line 132: > [!NOTE]
> **Advisory — non-blocking.** Safe to merge and address in a follow-up.
>
> _Why non-blocking:_ No demonstrated production impact; the finding describes a protocol compliance gap without evidence of broken clients, security vulnerability, or data loss.

Validate `MCP-Protocol-Version` on every `/mcp` request and reject unsupported versions with HTTP 400.

Currently `/mcp` ignores the header entirely. A client that negotiated a recent version but later sends `tools/list` with `MCP-Protocol-Version: 1999-01-01` receives a normal 200 response, despite Streamable HTTP requiring 400. The version check only runs during `initialize`, so there’s no guard for later requests.

<sub>🟠 `IMPORTANT` · confidence 99%</sub>

<details><summary>Evidence</summary>

> Step 1: `registerMCPRoutes` directly installs `MCPHandler(...).handle` at `POST /mcp` in `routes_mcp.go:47`; no route middleware validates the protocol header.
> Step 2: `handle` begins at line 124 and only reads `c.Request.Body`; it never reads `c.Request.Header.Get("MCP-Protocol-Version")`.
> Step 3: A POST `{"jsonrpc":"2.0","id":2,"method":"tools/list"}` with `MCP-Protocol-Version: 1999-01-01` therefore reaches the `tools/list` switch arm at lines 157-164.
> Step 4: `writeResult` sends HTTP 200 JSON at lines 627-629. `negotiateProtocolVersion` is only called from `handleInitialize` using `params.ProtocolVersion` at lines 171-188, so it cannot reject the invalid later request header.
> Step 5: The existing helper unconditionally sets a supported header (`mcp_test.go:96`) and the protocol tests never override or omit it, so this behavior is untested.

</details>

**💡 Suggested Fix**

Validate `MCP-Protocol-Version` before dispatch: accept supported values, apply the documented compatibility default only when it is absent, and reject unsupported values with HTTP 400. Add real HTTP-style tests for a supported header, an absent header (if compatibility is intentional), and an unsupported header; assert the unsupported case is 400 rather than a JSON-RPC success response.

---
*`MCP transport / JSON-RPC contract` · confidence 99%*

control-plane/internal/handlers/mcp.go
```diff
@@ -0,0 +1,761 @@
...
+	for _, a := range agents {
+		if a == nil || a.ID != node {
+			continue
+		}
+		for _, r := range a.Reasoners {
+			if r.ID == name {
+				return true
+			}
+		}
+		for _, sk := range a.Skills {
```
AbirAbbas, line 650: > [!NOTE]
> **Advisory — non-blocking.** Safe to merge and address in a follow-up.
>
> _Why non-blocking:_ The inconsistency does not cause a security vulnerability, data loss, or break an existing contract real callers depend on; it is a design polish issue.

**`execute_reasoner` accepts hidden skills**
<sub>🟠 `IMPORTANT` · confidence 99%</sub>

Enforce the reasoner-only contract by removing the `a.Skills` loop from `targetExists` in `mcp.go:610-613`.

The `execute_reasoner` tool can start a skill-only node that is hidden from `discover_agents` and `get_reasoner_schema`, breaking the documented discover → schema → execute workflow. This means undocumented skills are silently executable, while the MCP interface only advertises reasoners.

<details><summary>Evidence</summary>

> Step 1: An MCP client calls `tools/call` with `name="execute_reasoner"` and `target="skill-node.only-skill"`.
> Step 2: `toolExecuteReasoner` parses the target and calls `targetExists(agents, parsed.NodeID, parsed.TargetName)` at `mcp.go:382`.
> Step 3: `targetExists` iterates `a.Skills` and returns true when `sk.ID == name` at `mcp.go:610-613`, so `startAsyncRun` is reached.
> Step 4: `startAsyncRun` calls `prepareExecutionForTarget`, whose `determineTargetType` returns `"skill"` for a matching `agent.Skills` entry (`execute.go:2096-2099`); the execution URL resolver then selects `/skills/<target>` (`execute.go:2120-2123`). Thus the skill is executable.
> Step 5: In contrast, `toolDiscoverAgents` serializes only `capp.Reasoners` (`mcp.go:268-280`), while `toolGetReasonerSchema` searches only `capp.Reasoners` (`mcp.go:329-351`). Therefore a node containing only that skill presents no discoverable target or schema, contrary to the reasoner-only tool names/catalog and `docs/mcp-integration.md:37-39`.

</details>

control-plane/internal/handlers/mcp.go
```diff
@@ -0,0 +1,761 @@
...
+	}
+	if srv.version == "" {
+		srv.version = "dev"
+	}
+	return srv.handle
+}
+
+// handle is the POST /mcp entry point.
+func (s *mcpServer) handle(c *gin.Context) {
+	body, err := io.ReadAll(c.Request.Body)
```
AbirAbbas, line 133: > [!NOTE]
> **Advisory — non-blocking.** Safe to merge and address in a follow-up.
>
> _Why non-blocking:_ Memory exhaustion via unlimited body is not listed as blocking; error code mismatch is non-critical.

**Bound `/mcp` request bodies independently of optional DID auth**

<sub>🟠 `IMPORTANT` · confidence 99%</sub>

Add a local request body size limit to the `/mcp` handler. Without DID auth, `POST /mcp` reads the body with `io.ReadAll` (mcpServer.handle, line 125) without a size bound, so a chunked or large request can exhaust server memory. The DID middleware’s `io.LimitReader` only applies when DID is enabled and `X-Caller-DID` is present, leaving ordinary API‑key or unauthenticated requests unprotected.

<details><summary>Evidence</summary>

> Step 1: `AgentFieldServer.setupRoutes` calls `applyGlobalMiddleware()` before `registerMCPRoutes()`, and `registerMCPRoutes` registers `POST /mcp` with `handlers.MCPHandler`.
> Step 2: The always-installed middleware in `routes_middleware.go` is CORS, logging, a context-only one-hour timeout, and API-key authentication; none reads or bounds `c.Request.Body`.
> Step 3: DID middleware is installed only when `Features.DID.Enabled`, `DIDAuthEnabled`, and `didWebService != nil`; even then it calls `c.Next()` without reading the body at `did_auth.go:177-182` when `X-Caller-DID` is absent. Its `io.LimitReader(..., maxBodySize+1)` at `did_auth.go:224-241` therefore does not protect ordinary API-key-only or unauthenticated `/mcp` requests.
> Step 4: Those requests reach `mcpServer.handle`, where line 125 calls unbounded `io.ReadAll(c.Request.Body)`. `io.ReadAll` accumulates every received byte, so a chunked oversized body allocates proportionally without a handler limit or Content-Length rejection.
> Step 5: If that read returns an error, lines 126-128 call `writeError(..., mcpErrParse, ...)`; `writeError` always emits HTTP 200 at lines 631-636, producing a parse-error response for the body-read failure.

</details>

**💡 Suggested Fix**

Add a local MCP maximum body size (for example, wrap `c.Request.Body` with `http.MaxBytesReader` or use `io.LimitReader(max+1)` before `ReadAll`), return 413 when it is exceeded, and return a transport-appropriate 400/5xx response for other read errors rather than `mcpErrParse`. Add handler tests using a body larger than the limit and a custom `io.ReadCloser` that returns a read error; exercise the route without DID headers so the tests prove the local guard is required.

---
*`MCP request body safety` · confidence 99%*

control-plane/internal/server/server.go
```diff
@@ -885,6 +885,7 @@ func (s *AgentFieldServer) setupRoutes() {
 	}
 
 	s.registerKBRoutes()
+	s.registerMCPRoutes()
```
AbirAbbas, line 888: > [!CAUTION]
> **Must-fix before merge.** MCP endpoint lacks permission checks, allowing authenticated users to execute agents and access runs without authorization, a security vulnerability.

**MCP bypasses REST execution permission checks**

<sub>🟠 `IMPORTANT` · confidence 99%</sub>

Add permission checks to MCP tool handlers to match REST execution authorization.

The MCP handler (`handlers/mcp.go`) calls `startAsyncRun` and `buildRunView` directly, bypassing the `PermissionCheckMiddleware` that protects `/api/v1/execute`. A caller authenticated by global middleware can execute reasoners or fetch runs without the target-agent policy evaluation required for REST calls.

<details><summary>Evidence</summary>

> Changed composition: `server.go:864` calls `s.applyGlobalMiddleware()`, then `server.go:888` calls `s.registerMCPRoutes()`. MCP registration is only `s.Router.POST("/mcp", handler)` (`routes_mcp.go:49`). By contrast, REST execution creates `executeGroup := agentAPI.Group("/execute")` and, when authorization is enabled, applies `executeGroup.Use(middleware.PermissionCheckMiddleware(...))` before registering `POST "/:target"` and `POST "/async/:target"` (`routes_core.go:93-115`). MCP explicitly bypasses HTTP routing: its store comment says MCP tools "call the service layer directly rather than looping back through HTTP" (`handlers/mcp.go:21-25`); `toolExecuteReasoner` calls `s.startAsyncRun(ctx, target, input)` (`handlers/mcp.go:359-403`), while `toolGetRun` calls `s.buildRunView(ctx, runID)` (`handlers/mcp.go:406-425`).

</details>

**💡 Suggested Fix**

Apply an MCP-specific authorization layer before dispatching each tool call. It should derive the verified DID established by global middleware and run the same target-agent access-policy evaluation as `PermissionCheckMiddleware` for `execute_reasoner`; add equivalent ownership/visibility authorization for `get_run` and `wait_run`. Alternatively, refactor that authorization into a shared service used by both REST and MCP, since MCP's JSON-RPC target and run ID are not Gin path parameters.

---
*`Consistency Verifier` · confidence 99%*

control-plane/internal/handlers/mcp.go
AbirAbbas: > [!CAUTION]
> **Must-fix before merge.** Bypassing DID authorization allows unauthorized execution of targets, a security vulnerability.

**Route the MCP handler through the DID authorization middleware and pass verified identities to `prepareExecutionForTarget`.**

<sub>🟠 `IMPORTANT` · confidence 99%</sub>

The MCP path at `control-plane/internal/handlers/mcp.go:488` calls `prepareExecutionForTarget` with empty caller and target DIDs, bypassing `PermissionCheckMiddleware` (enforced by REST at `execute.go:1385-1392`). A caller allowed to hit `/mcp` can invoke targets that would be blocked via the REST endpoint; the downstream agent receives no `X-Caller-DID` or `X-Target-DID`.

<details><summary>Evidence</summary>

> Changed MCP path (control-plane/internal/handlers/mcp.go:488): `plan, err := controller.prepareExecutionForTarget(ctx, target, ExecuteRequest{Input: input}, executionHeaders{}, "", "")`. REST path (control-plane/internal/handlers/execute.go:1385-1392): `return c.prepareExecutionForTarget(..., readExecutionHeaders(ginCtx), middleware.GetVerifiedCallerDID(ginCtx), middleware.GetTargetDID(ginCtx))`. The REST async route is in `executeGroup` (routes_core.go:93-115), which installs `PermissionCheckMiddleware` when DID authorization is enabled (lines 95-106). That middleware resolves the verified caller DID and evaluates policies, aborting denied calls with 403 (permission.go:148-169, 228-241). MCP is registered directly as `s.Router.POST("/mcp", handler)` (routes_mcp.go:40-49), outside that group. The shared call forwarding only sets DID headers if the plan contains them (execute.go:1799-1804), while MCP supplies empty strings.

</details>

**💡 Suggested Fix**

Add the `/mcp` route to the `executeGroup` middleware chain (or replicate the policy check in the handler) and supply the verified caller and target DIDs to `prepareExecutionForTarget`, blocking tool execution until authorization succeeds.

---
*`Consistency Verifier` · confidence 99%*

control-plane/internal/handlers/mcp.go
```diff
@@ -0,0 +1,761 @@
...
+		return mcpToolError(fmt.Sprintf(
+			"invalid target %q: %v. Expected \"node.reasoner\". Call discover_agents to see valid targets.",
+			target, err)), nil
+	}
+
+	agents, err := s.store.ListAgents(ctx, types.AgentFilters{})
+	if err != nil {
+		return nil, fmt.Errorf("list agents: %w", err)
+	}
+	if !targetExists(agents, parsed.NodeID, parsed.TargetName) {
```
AbirAbbas, line 391: > [!NOTE]
> **Advisory — non-blocking.** Safe to merge and address in a follow-up.
>
> _Why non-blocking:_ The mismatch between advertised contract and accepted targets is a documentation/consistency issue, not a breaking change or security vulnerability.

MCP accepts skill targets that its discovery contract never advertises

<sub>🟠 `IMPORTANT` · confidence 99%</sub>

Restrict accepted targets to those the MCP discovery contract actually advertises. `targetExists` now accepts both reasoners and skills (`control-plane/internal/handlers/mcp.go:605-614`), but the discovery response (`mcp.go:267-287`) emits only a `reasoners` array, and the `execute_reasoner` tool (`mcp.go:710-721`) describes its target as a reasoner discovered from that array. This mismatch lets a skill target pass validation while being undiscoverable and not fitting the execute tool’s schema, breaking the advertised contract.

<details><summary>Evidence</summary>

> Changed acceptance (control-plane/internal/handlers/mcp.go:605-614): `for _, r := range a.Reasoners { if r.ID == name { return true } }` followed by `for _, sk := range a.Skills { if sk.ID == name { return true } }`.
> 
> Execution resolver (control-plane/internal/handlers/execute.go:2090-2101): `determineTargetType` returns `"reasoner"` for `agent.Reasoners` and `"skill"` for `agent.Skills`.
> 
> Discovery response (control-plane/internal/handlers/mcp.go:267-287): builds `reasoners` only by iterating `capp.Reasoners`, then returns each agent with only `"reasoners": reasoners`; never emits `capp.Skills`.
> 
> Advertised contract (control-plane/internal/handlers/mcp.go:710-721): `execute_reasoner` is described as executing `a reasoner`; its target is `"node.reasoner" (from discover_agents)`, and input is `Input object for the reasoner.`

</details>

control-plane/internal/handlers/mcp.go
```diff
@@ -0,0 +1,761 @@
...
+		})
+	}
+
+	view = map[string]interface{}{
+		"run_id":     runID,
+		"status":     deriveRunStatus(execs),
+		"executions": summaries,
+	}
+
+	if root := rootExecution(execs); root != nil {
```
AbirAbbas, line 584: > [!NOTE]
> **Advisory — non-blocking.** Safe to merge and address in a follow-up.
>
> _Why non-blocking:_ Exposing a child payload as root result is a functional bug, not a security, data loss, or contract-breaking issue.

Stop exposing an arbitrary child execution’s payload as the run-level result.
<sub>🟠 `IMPORTANT` · confidence 98%</sub>
`rootExecution` falls back to `execs[0]` when no execution has an empty `ParentExecutionID`. `QueryExecutionRecords` orders only by `started_at` without a tie-breaker, so the first child is arbitrary. That child’s own `ResultPayload`/`ErrorMessage` is then returned as the run-level result.
<details><summary>Evidence</summary>

> Changed end: `control-plane/internal/handlers/mcp.go:544-550` calls `rootExecution(execs)` and reads `root.ResultPayload` / `root.ErrorMessage`; `mcp.go:566-574` returns the first record if none has an empty parent. Other end: `control-plane/internal/handlers/execute.go:2035-2061` reads `X-Parent-Execution-ID` into `parentExecutionID` whenever it is non-empty, and `execute.go:1491-1494` assigns it directly to a new execution; `execute.go:1883-1885` stores the completed execution's own `ResultPayload`, while `execute.go:1963-1972` stores that execution's own error/result. Storage: `control-plane/internal/storage/execution_records.go:294-317` defaults to `ORDER BY started_at ASC` and supplies no secondary ordering column.

</details>

**💡 Suggested Fix**

Do not use the first record as a result owner when no parentless root exists. Return no run-level result/error (or select an explicitly persisted root execution ID) and add a deterministic ordering tie-breaker if any ordering-dependent fallback remains.

---
*`Consistency Verifier` · confidence 98%*

control-plane/internal/handlers/mcp.go
```diff
@@ -0,0 +1,761 @@
...
+	}
+
+	resp := buildDiscoveryResponse(agents, filters)
+	out := make([]map[string]interface{}, 0, len(resp.Capabilities))
+	for _, capp := range resp.Capabilities {
+		reasoners := make([]map[string]interface{}, 0, len(capp.Reasoners))
+		for _, r := range capp.Reasoners {
+			rm := map[string]interface{}{
+				"id":     r.ID,
+				"target": capp.AgentID + "." + r.ID,
```
AbirAbbas, line 280: > [!NOTE]
> **Advisory — non-blocking.** Safe to merge and address in a follow-up.
>
> _Why non-blocking:_ The finding describes a potential functional failure if IDs contain dots, but no concrete production impact is demonstrated; it is a defensive improvement that can be addressed later.

Advertised MCP targets can be unparsable by execution

<sub>🟠 `IMPORTANT` · confidence 98%</sub>

Reject dots in agent and reasoner IDs or replace the dot-concatenation with an unambiguous encoding.

Agent registration allows any non-empty node ID (`,validate:"required,min=1"`) and reasoner IDs have no validation, so an ID like `team.alpha` yields the target `team.alpha.summarize`. Execution splits targets on dots and rejects any with len != 2, making the advertised target unparseable. Discovery therefore returns targets that the execution path cannot consume.

<details><summary>Evidence</summary>

> Discovery at mcp.go:270-273: `rm := map[string]interface{}{ "id": r.ID, "target": capp.AgentID + "." + r.ID, }`. Schema lookup repeats this at mcp.go:333-337: `"target": node + "." + r.ID`. Execution at execute.go:2076-2087: `parts := strings.Split(value, ".")`; `if len(parts) != 2 { return nil, fmt.Errorf("target must be in format 'node_id.reasoner_name'") }`; then assigns `NodeID: parts[0]` and `TargetName: parts[1]`. AgentNode ID is only `validate:"required,min=1"` at types.go:159, while ReasonerDefinition.ID at types.go:226-237 has no validation constraint.

</details>

**💡 Suggested Fix**

Define and enforce a shared target identifier contract. Either reject dots in registered node/reasoner IDs, or replace dot-concatenation/parsing with an unambiguous encoding shared by discovery, schema lookup, and execution (for example structured node/reasoner fields or an escaped delimiter).

---
*`Consistency Verifier` · confidence 98%*

AbirAbbas (commented):
## 🔴 PR-AF Review — **Multiple Merge-Blockers**

> 🚫 **Merge blocked.** 5 must-fix issues found by automated review.

> **17 findings** · 🚫 5 blocking · 💬 12 advisory · 🔴 1 critical · 🟠 12 important · 🔵 4 suggestions · ⚪ 0 nitpicks

<details>

## What & why

AgentField had no MCP server — the only way for an AI harness to drive installed
agents was the `af` CLI plus the `agentfield-use` skill. This PR ships an
**MCP server embedded in the control plane itself, served on the same port as
the REST API**. Whether the control plane was started by the CLI (`af server`)
or the desktop app, MCP is just there — zero extra processes, zero setup. A
harness connects with one command:

```bash
claude mcp add --transport http agentfield http://localhost:8080/mcp
```

It is **on by default**, with an env kill-switch: `AGENTFIELD_MCP_ENABLED=false`
makes `/mcp` return `404`. The endpoint sits behind the same global API-key auth
and trust domain as `/api/v1`, and the tools call the existing discovery /
execution service layer directly (no loopback HTTP).

- **Transport:** streamable HTTP, JSON-RPC 2.0 over `POST /mcp` (stateless;
  `Mcp-Session-Id` / `MCP-Protocol-Version` headers tolerated and ignored).
  `GET /mcp` → 405, `OPTIONS /mcp` → 204.

## The five tools

| Tool | Purpose |
|---|---|
| `discover_agents` | List agents + reasoners (id, description, entrypoint tags). Defaults to active-only; `health:"all"` includes unhealthy. |
| `get_reasoner_schema` | Input/output JSON Schema for a reasoner, as served by discovery. |
| `execute_reasoner` | Validate `node.reasoner` exists, start an **async** run, return `{run_id, status:"accepted"}`. |
| `get_run` | Run status + result/error + per-execution summaries. |
| `wait_run` | Server-side poll to terminal or timeout (default 60s, hard-capped at 120s); adds a `timed_out` flag. |

## Validation Contract → tests

- `initialize` returns protocolVersion + capabilities.tools + serverInfo →
  `TestMCP_InitializeAndToolsList` (also asserts unknown protocol falls back to
  `2025-06-18` and serverInfo carries the build version).
- `tools/list` returns exactly the 5 tools, each with a valid JSON Schema →
  `TestMCP_InitializeAndToolsList`.
- `discover_agents` default returns only active agents; `health:"all"` includes
  dead ones with `health_status` → `TestMCP_DiscoverAgents`.
- `execute_reasoner` bad/malformed target → tool error naming `discover_agents`;
  good target → `run_id` and a created execution → `TestMCP_ExecuteReasoner`.
- `get_run` aggregates status/result/error + per-execution summaries →
  `TestMCP_GetRun`.
- `wait_run` respects the timeout cap and returns `timed_out` on a never-finishing
  run (and returns immediately for a terminal run) → `TestMCP_WaitRun`.
- Unknown method → `-32601`; unknown tool → `-32602`; batch arrays rejected →
  `TestMCP_ProtocolMiscMethods`.
- `AGENTFIELD_MCP_ENABLED=false` → `/mcp` is `404`; default → enabled;
  `GET /mcp` → 405 → `TestMCPRoutes_DisabledReturns404`,
  `TestMCPRoutes_EnabledByDefault`.
- Config default-on + env override → `TestMCPConfig_IsEnabledDefault`,
  `TestApplyEnvOverrides_MCPEnabled`.

## Test results

- `go build ./...` — clean
- `go vet ./internal/config/... ./internal/handlers/... ./internal/server/...` — clean
- `go test` for the config, handlers, server (MCP) and skillkit packages — all pass
- `gofmt -l` on changed files — empty
- `golangci-lint` — no issues in any changed/new file

One pre-existing, environment-specific test (`TestStartAndStopCoverAdditionalBranches`,
admin gRPC listener bind) fails identically on clean `origin/main` in this sandbox
and is unrelated to this change.

</details>

### Key Findings

**5 issue(s) should be addressed before merge:**

- 🔴 **MCP execution bypasses REST policy enforcement and drops authenticated provenance** (`control-plane/internal/handlers/mcp.go:488`) — `POST /mcp` bypasses the authorization boundary that protects the equivalent REST endpoints. Gate: Authorization bypass allows unprivileged callers to execute operations that REST would deny, violating access control and dropping authenticated provenance.
- 🟠 **MCP bypasses REST execution permission checks** (`control-plane/internal/server/server.go:888`) — `setupRoutes` installs global middleware before registering `/mcp`, so MCP inherits API-key authentication and the optional DID-auth middleware. Gate: MCP endpoint lacks permission checks, allowing authenticated users to execute agents and access runs without authorization, a security vulnerability.
- 🟠 **MCP async execution bypasses REST DID policy enforcement** (`control-plane/internal/handlers/mcp.go:488`) — The MCP execution path bypasses the authorization middleware and discards the verified caller and target identities that REST passes into the shared preparation path. Gate: Bypassing DID authorization allows unauthorized execution of targets, a security vulnerability.
- 🟠 **Authorize run reads against the authenticated principal** (`control-plane/internal/handlers/mcp.go:516`) — `get_run` and `wait_run` turn a run ID into a read capability without binding it to the authenticated caller. Gate: Missing access control allows authenticated users to view other users' run results; a reachable information disclosure vulnerability.
- 🟠 **Honor the MCP toggle from DB-backed YAML configuration** (`control-plane/internal/config/config.go:258`) — `features.mcp.enabled` is ignored when the server is configured to load its YAML from the database. Gate: Failing to honor the MCP toggle in DB-backed config leaves an execution-capable endpoint enabled against operator intent, constituting a security misconfiguration and breaking the configuration contract.

**12 advisory finding(s) surfaced as non-blocking:**

- 🟠 Unsupported MCP protocol-version headers are accepted (`control-plane/internal/handlers/mcp.go:124`)
- 🟠 `execute_reasoner` accepts hidden skills (`control-plane/internal/handlers/mcp.go:610`)
- 🟠 Allow MCP protocol headers in the default CORS policy (`control-plane/internal/server/routes_middleware.go:34`)
- 🟠 Bound `/mcp` request bodies independently of optional DID auth (`control-plane/internal/handlers/mcp.go:125`)
- 🟠 MCP accepts skill targets that its discovery contract never advertises (`control-plane/internal/handlers/mcp.go:382`)
- … and 7 more (see All Findings by Severity)

**Files with findings:** `control-plane/internal/config/config.go`, `control-plane/internal/handlers/mcp.go`, `control-plane/internal/server/routes_mcp.go`, `control-plane/internal/server/routes_middleware.go`, `control-plane/internal/server/server.go`

<details>

#### 🔴 Critical (1)

- **MCP execution bypasses REST policy enforcement and drops authenticated provenance** `control-plane/internal/handlers/mcp.go:488`

#### 🟠 Important (12)

- **Unsupported MCP protocol-version headers are accepted** `control-plane/internal/handlers/mcp.go:124`
- **`execute_reasoner` accepts hidden skills** `control-plane/internal/handlers/mcp.go:610`
- **Allow MCP protocol headers in the default CORS policy** `control-plane/internal/server/routes_middleware.go:34`
- **Bound `/mcp` request bodies independently of optional DID auth** `control-plane/internal/handlers/mcp.go:125`
- **MCP bypasses REST execution permission checks** `control-plane/internal/server/server.go:888`
- **MCP async execution bypasses REST DID policy enforcement** `control-plane/internal/handlers/mcp.go:488`
- **MCP accepts skill targets that its discovery contract never advertises** `control-plane/internal/handlers/mcp.go:382`
- **Do not expose the first child payload as a root result** `control-plane/internal/handlers/mcp.go:544`
- **Advertised MCP targets can be unparsable by execution** `control-plane/internal/handlers/mcp.go:272`
- **Authorize run reads against the authenticated principal** `control-plane/internal/handlers/mcp.go:516`
- **Honor the MCP toggle from DB-backed YAML configuration** `control-plane/internal/config/config.go:258`
- **Do not report request cancellation as a wait_run timeout** `control-plane/internal/handlers/mcp.go:473`

#### 🔵 Suggestion (4)

- **ID-bearing notifications are incorrectly discarded** `control-plane/internal/handlers/mcp.go:150`
- **JSON-RPC envelope and ID validation is missing** `control-plane/internal/handlers/mcp.go:144`
- **Make the advertised OPTIONS probe reachable with API-key auth** `control-plane/internal/server/routes_mcp.go:56`
- **Run aggregation changes when storage record order changes** `control-plane/internal/handlers/mcp.go:515`

</details>

<details>

**Dimensions Analyzed (5):**

- **MCP execution preserves the REST authorization boundary and execution identity** — 3 file(s)
- **Run lookup does not leak results and remains deterministic for multi-execution runs** — 2 file(s)
- **Reasoner-only MCP workflow cannot execute undiscoverable skills** — 3 file(s)
- **JSON-RPC and Streamable HTTP framing compliance** — 3 file(s)
- **MCP request-body bounding and failure mechanics** — 4 file(s)

- **Semantic** — 3 dimension(s), 93% coverage confidence
- **Mechanical** — 2 dimension(s), 90% coverage confidence
- **Systemic** — 3 dimension(s), 93% coverage confidence

- **9** finding(s) adversarially tested: 9 confirmed, 0 challenged

</details>

<details>

| Metric | Value |
|--------|-------|
| Duration | 909.2s |
| Agent invocations | 38 |
| Coverage iterations | 2 |
| Budget exhausted | No |
| PR type | feature |
| Complexity | medium |

</details>

Review ID: `rev_0f887b0ff579`

<br>
<div align="right">
    AgentField PR-AF
  </a>
</div>

control-plane/internal/handlers/mcp.go
```diff
@@ -0,0 +1,761 @@
...
+	}
+
+	return plan.exec.RunID, plan.exec.ExecutionID, nil
+}
+
+// buildRunView aggregates all executions for a run into the shape get_run and
+// wait_run return. terminal reports whether every execution has reached a
+// terminal state; found reports whether the run exists at all.
+func (s *mcpServer) buildRunView(ctx context.Context, runID string) (view map[string]interface{}, terminal, found bool, err error) {
+	execs, err := s.store.QueryExecutionRecords(ctx, types.ExecutionFilter{RunID: &runID})
```
AbirAbbas, line 550: > [!CAUTION]
> **Must-fix before merge.** Missing access control allows authenticated users to view other users' run results; a reachable information disclosure vulnerability.

**Authorize run reads against the authenticated principal**

<sub>🟠 `IMPORTANT` · confidence 94%</sub>

Bind run reads to the authenticated caller: read the DID from the Gin context and filter `QueryExecutionRecords` by it. The handlers `get_run` and `wait_run` ignore the caller’s identity stored in the context, and the execution schema has no DID ownership field to filter on. Because of this, any authenticated user can pass a valid run ID and retrieve another principal’s root result, error text, and execution metadata.

<details><summary>Evidence</summary>

> Step 1: `setupRoutes` applies `DIDAuthMiddleware` globally before registering `/mcp` (`internal/server/server.go:863-889`; `routes_middleware.go:89-101`), and successful authentication stores the caller DID only in Gin context (`middleware/did_auth.go:305-307`). Step 2: an authenticated caller invokes `tools/call` with `name="get_run"` or `"wait_run"`; `handleToolsCall` calls `toolGetRun`/`toolWaitRun` with only `c.Request.Context()` (`handlers/mcp.go:200-215`). Step 3: both tools pass the caller-supplied string to `buildRunView` (`mcp.go:418`, `457`), which calls `QueryExecutionRecords` with only `ExecutionFilter{RunID: &runID}` (`mcp.go:515-516`). Step 4: `QueryExecutionRecords` supports `ActorID` but applies it only when supplied (`storage/execution_records.go:259-262`), while `types.Execution` has no caller-DID field (`pkg/types/execution.go:10-47`); the returned root `ResultPayload` and `ErrorMessage` are copied into the MCP response (`mcp.go:544-559`). Therefore a different verified DID with the run ID receives that principal's result/error.

</details>

**💡 Suggested Fix**

Make run reads identity-aware: retain authenticated caller identity for MCP requests, persist a trustworthy owner/principal on each run/execution, and query/authorize against it before returning a view. Add integration tests that issue requests as two distinct verified DIDs and assert that the non-owner receives no result, error, or execution metadata for the other's run.

---
*`Run visibility and deterministic aggregation` · confidence 94%*

control-plane/internal/config/config.go
```diff
@@ -249,6 +249,22 @@ type FeatureConfig struct {
 	Connector ConnectorConfig `yaml:"connector" mapstructure:"connector"`
 	Tracing   TracingConfig   `yaml:"tracing" mapstructure:"tracing"`
 	Knowledge KnowledgeConfig `yaml:"knowledge" mapstructure:"knowledge"`
+	MCP       MCPConfig       `yaml:"mcp" mapstructure:"mcp"`
+}
+
+// MCPConfig configures the embedded Model Context Protocol server that AI
+// harnesses (Claude Code, etc.) connect to at <server>/mcp. The MCP endpoint is
+// served on the same port and shares the same trust domain as the REST API.
+type MCPConfig struct {
```
AbirAbbas, line 260: > [!CAUTION]
> **Must-fix before merge.** Failing to honor the MCP toggle in DB-backed config leaves an execution-capable endpoint enabled against operator intent, constituting a security misconfiguration and breaking the configuration contract.

Fix `mergeDBConfig` to merge `dbCfg.Features.MCP` so that setting `features.mcp.enabled: false` in the DB-stored YAML actually disables MCP routes.

<sub>🟠 `IMPORTANT` · confidence 94%</sub>

Currently, when `AGENTFIELD_CONFIG_SOURCE=db` is used, `overlayDBConfig` correctly parses the stored `agentfield.yaml`, but `mergeDBConfig` only copies `Features.DID`, ignoring `Features.MCP`. As a result, the nil config defaults to enabled in `MCPConfig.IsEnabled()`, and the execution-capable `POST /mcp` route remains exposed despite the administrator’s explicit intent to disable it.

<details><summary>Evidence</summary>

> Step 1: A production server started with `AGENTFIELD_CONFIG_SOURCE=db` calls `overlayDBConfig(cfg, storageProvider)` from `NewAgentFieldServer` before route setup. Step 2: `overlayDBConfig` unmarshals the stored `agentfield.yaml`, so `features:\n  mcp:\n    enabled: false` populates `dbCfg.Features.MCP.Enabled` with `*false`. Step 3: `mergeDBConfig` only copies `dbCfg.Features.DID` under Features and has no branch that copies `dbCfg.Features.MCP`; the original file/env value therefore remains nil. Step 4: `setupRoutes` calls `registerMCPRoutes`, which evaluates the nil value through `MCPConfig.IsEnabled()` as true and registers `POST /mcp` despite the DB YAML explicitly disabling it.

</details>

**💡 Suggested Fix**

Decide and document the intended precedence for MCP. If DB-backed YAML is meant to be a supported configuration source, merge an explicitly non-nil `dbCfg.Features.MCP.Enabled` into `target.Features.MCP.Enabled` in `mergeDBConfig`, and add an `AGENTFIELD_CONFIG_SOURCE=db` route-registration test. If MCP must be file/env-only as a deployment guardrail, make that restriction explicit in the configuration contract rather than presenting `mcp.enabled` as a general YAML setting.

---
*`Coverage gap review` · confidence 94%*

control-plane/internal/handlers/mcp.go
```diff
@@ -0,0 +1,761 @@
...
+		if terminal {
+			view["timed_out"] = false
+			return mcpToolText(view), nil
+		}
+		if !time.Now().Before(deadline) {
+			view["timed_out"] = true
+			return mcpToolText(view), nil
+		}
+		select {
+		case <-ctx.Done():
```
AbirAbbas, line 500: > [!NOTE]
> **Advisory — non-blocking.** Safe to merge and address in a follow-up.
>
> _Why non-blocking:_ Mislabeling request cancellation as timeout does not cause data loss, security issues, or API breakage; client cannot receive the result anyway.

**Do not report request cancellation as a wait_run timeout**

<sub>🟠 `IMPORTANT` · confidence 94%</sub>

The `case <-ctx.Done()` path in `wait_run` fires on HTTP request cancellation (e.g., client timeout) and incorrectly sets `timed_out: true`, returning a tool result the client can't receive. This conflates cancellation with the tool's own deadline check at lines 468-470.

<details><summary>Evidence</summary>

> Changed location (control-plane/internal/handlers/mcp.go:468-475): `if !time.Now().Before(deadline) { view["timed_out"] = true; return mcpToolText(view), nil }` is followed by `case <-ctx.Done(): view["timed_out"] = true; return mcpToolText(view), nil`. The caller passes `ctx := c.Request.Context()` to `toolWaitRun` at lines 200 and 214-215. Other end (control-plane/internal/server/routes_middleware.go:54-61): `ctx := c.Request.Context()`; `timeoutCtx, cancel := context.WithTimeout(ctx, 3600*time.Second)`; `c.Request = c.Request.WithContext(timeoutCtx)`. Thus the handler's context inherits HTTP request cancellation as well as the global timeout, not only the wait_run poll deadline.

</details>

**💡 Suggested Fix**

Return `ctx.Err()` from the `ctx.Done()` case instead of setting `timed_out: true`. Reserve `timed_out: true` for the local deadline check only.

---
*`Consistency Verifier`* · confidence 94%

control-plane/internal/handlers/mcp.go
```diff
@@ -0,0 +1,761 @@
...
+		return
+	}
+
+	var req jsonrpcRequest
+	if err := json.Unmarshal(trimmed, &req); err != nil {
+		s.writeError(c, nullID(), mcpErrParse, "parse error: "+err.Error(), nil)
+		return
+	}
+
+	// Notifications (no id) and any notifications/* method carry no response
```
AbirAbbas, line 158: > [!NOTE]
> **Advisory — non-blocking.** Safe to merge and address in a follow-up.
>
> _Why non-blocking:_ Protocol edge-case bug (202 vs error) for ID-bearing notifications/unknown, not a security, data-loss, or contract-breaking regression requiring immediate fix.

**ID-bearing notifications are incorrectly discarded**

<sub>🔵 `SUGGESTION` · confidence 99%</sub>

Stop dispatching ID-bearing messages under `notifications/*` as empty `202` notifications; dispatch them as standard JSON-RPC requests.

Currently, `handle` in `mcp_handler.go:152` checks the method prefix and returns `202` with an empty body regardless of `id`. A request like `{"jsonrpc":"2.0","id":7,"method":"notifications/unknown"}` therefore gets a `202` instead of a `-32601` error, causing any client that awaits the correlated response to hang.

<details><summary>Evidence</summary>

> Step 1: `registerMCPRoutes` registers `POST /mcp` with `MCPHandler(...).handle` in `routes_mcp.go:47`.
> Step 2: A client POSTs `{"jsonrpc":"2.0","id":7,"method":"notifications/unknown"}`.
> Step 3: `json.Unmarshal` stores the ID as `json.RawMessage("7")`; therefore this is an ID-bearing request.
> Step 4: `handle` evaluates `strings.HasPrefix(req.Method, "notifications/")` at line 152 and returns `c.Status(202)` at line 153 before the method switch.
> Step 5: No `writeError` call occurs, so the client gets an empty 202 rather than the required correlated method-not-found response.

</details>

**💡 Suggested Fix**

In `handle`, check the presence of `id` before the method prefix. If `id` exists, fall through to the normal method dispatch; if absent, return `202` empty. Add tests for both ID-bearing and ID-less `notifications/*` requests, verifying the correct status and response body (empty for notifications, JSON-RPC error for requests).

---
*`MCP transport / JSON-RPC contract` · confidence 99%*

control-plane/internal/handlers/mcp.go
```diff
@@ -0,0 +1,761 @@
...
+	}
+	// Batch requests (a top-level JSON array) are intentionally unsupported —
+	// harnesses send a single message per POST. Reject clearly rather than
+	// silently processing only the first element.
+	if trimmed[0] == '[' {
+		s.writeError(c, nullID(), mcpErrInvalidRequest, "batch requests are not supported; send a single JSON-RPC message", nil)
+		return
+	}
+
+	var req jsonrpcRequest
```
AbirAbbas, line 152: > [!NOTE]
> **Advisory — non-blocking.** Safe to merge and address in a follow-up.
>
> _Why non-blocking:_ No demonstrated security vulnerability, data loss, or broken contract; invalid JSON-RPC envelopes are not a production-blocking issue.

<sub>🔵 `SUGGESTION` · confidence 99%</sub>

**JSON-RPC envelope and ID validation is missing**

Validate the JSON-RPC envelope and reject invalid requests with error -32600.

The handler accepts requests where `jsonrpc` is not `"2.0"`, the method is missing, or the `id` is not a string or number. For example, `{"jsonrpc":"1.0","id":1,"method":"ping"}` returns a valid 200 response, `{}` or `null` return 202, and a boolean `id` is echoed back instead of returning an error with `id: null`. This breaks JSON-RPC correlation and validation guarantees.

<details><summary>Evidence</summary>

> Step 1: `POST /mcp` reaches `mcpServer.handle` through the route registered in `routes_mcp.go:47`.
> Step 2: At lines 144-148, the body is unmarshaled directly into `jsonrpcRequest`; there is no object-shape check and no subsequent check of `req.JSONRPC`, `req.Method`, or the JSON type held in `req.ID`.
> Step 3: For `{"jsonrpc":"1.0","id":1,"method":"ping"}`, unmarshal succeeds and line 157 dispatches `ping`; `writeResult` serializes a successful JSON-RPC 2.0 response at line 628.
> Step 4: For `{}` or `null`, Go unmarshals to the zero-value struct and line 152 sees an empty ID, so lines 153-154 emit an empty 202.
> Step 5: For boolean/object/array IDs, `json.RawMessage` preserves the invalid token; the nonempty-ID path dispatches and `writeResult`/`writeError` copy that token verbatim via `idOrNull` at lines 628-643, producing an invalid correlation ID instead of an invalid-request error with `id: null`.

</details>

**💡 Suggested Fix**

Parse into a raw top-level value and require a JSON object; validate `jsonrpc == "2.0"`, a nonempty string `method`, valid params shape as applicable, and an ID that is a string or number (with the chosen handling for `null`) before notification dispatch. Emit `-32600` with `id: null` for invalid envelopes. Add table-driven HTTP tests for `{}`, `null`, missing/invalid `jsonrpc`, missing/non-string `method`, and boolean/object/array IDs, asserting 200 JSON error response, `id:null`, and `Content-Type: application/json`.

---
*`MCP transport / JSON-RPC contract` · confidence 99%*

control-plane/internal/server/routes_mcp.go
```diff
@@ -0,0 +1,62 @@
...
+	)
+
+	s.Router.POST("/mcp", handler)
+	s.Router.GET("/mcp", func(c *gin.Context) {
+		c.Header("Allow", "POST, OPTIONS")
+		c.JSON(http.StatusMethodNotAllowed, gin.H{
+			"error": "method not allowed; POST a JSON-RPC 2.0 message to /mcp",
+		})
+	})
+	s.Router.OPTIONS("/mcp", func(c *gin.Context) {
```
AbirAbbas, line 63: > [!NOTE]
> **Advisory — non-blocking.** Safe to merge and address in a follow-up.
>
> _Why non-blocking:_ The OPTIONS probe returning 401 is a configuration/documentation issue, not a runtime break of an existing required contract without credentials.

Make the advertised OPTIONS probe reachable with API-key auth

<sub>🔵 `SUGGESTION` · confidence 90%</sub>

`OPTIONS /mcp` returns 401 instead of the documented 204 when API-key auth is enabled because the global API-key middleware intercepts the request before the route handler. This breaks the unauthenticated transport probe on secured deployments.

<details><summary>Evidence</summary>

> Step 1: `setupRoutes` calls `applyGlobalMiddleware()` before `registerMCPRoutes()` in `server.go:864-888`.
> Step 2: `applyGlobalMiddleware` installs `middleware.APIKeyAuth` globally in `routes_middleware.go:79-84`; `/mcp` is not in its skip paths.
> Step 3: With `API.Auth.APIKey` configured, a client sends plain `OPTIONS /mcp` without `X-API-Key` or `Authorization`; `APIKeyAuth` reaches its constant-time comparison and calls `AbortWithStatusJSON(401, ...)` in `middleware/auth.go:124-140`.
> Step 4: Gin therefore never reaches the route handler registered here at line 56, so the route's stated `204` probe behavior is unreachable for that request.

</details>

**💡 Suggested Fix**

Either exempt unauthenticated `OPTIONS /mcp` from API-key enforcement (keep auth for POST), or document that the probe requires credentials. Add a route test with `API.Auth.APIKey` configured and a headerless plain OPTIONS request to pin the chosen contract.

---
*`Coverage gap review — MCP server wiring` · confidence 90%*

control-plane/internal/handlers/mcp.go
```diff
@@ -0,0 +1,761 @@
...
+		return "", "", queueErr
+	}
+
+	return plan.exec.RunID, plan.exec.ExecutionID, nil
+}
+
+// buildRunView aggregates all executions for a run into the shape get_run and
+// wait_run return. terminal reports whether every execution has reached a
+// terminal state; found reports whether the run exists at all.
+func (s *mcpServer) buildRunView(ctx context.Context, runID string) (view map[string]interface{}, terminal, found bool, err error) {
```
AbirAbbas, line 542: > [!NOTE]
> **Advisory — non-blocking.** Safe to merge and address in a follow-up.
>
> _Why non-blocking:_ Non-deterministic aggregation is a quality improvement, not a breaking bug, security issue, or data loss; can be fixed later.

Make run aggregation deterministic regardless of storage record order.

The run view derives status, error, result, and execution order directly from the records in the order they come from storage, and that order varies because the current default `ORDER BY started_at ASC` lacks a tie-breaker and other `MCPStore` implementations may return a different sequence. The status becomes whichever terminal failure was encountered first, the error field picks the first non-empty child error, the root-result falls back to the first record, and `executions` preserves that same unspecified order – so `get_run` and `wait_run` can report different outcomes for identical inputs.

<details><summary>Evidence</summary>

> Step 1: `buildRunView` requests records without `SortBy` or a documented ordering requirement (`handlers/mcp.go:515-516`). Step 2: the storage implementation defaults to `ORDER BY started_at ASC` (`storage/execution_records.go:294-317`), but supplies no secondary key, while the MCPStore interface permits alternate implementations. Step 3: summaries are appended in returned order (`mcp.go:525-536`); `rootExecution` returns `execs[0]` if every record has a parent (`mcp.go:564-575`); the first non-empty error wins (`mcp.go:552-558`); and `deriveRunStatus` immediately returns the first failed/timeout/cancelled status (`mcp.go:581-596`). Step 4: for one run containing failed and timeout records with different errors, reversing the returned slice changes the reported status and error; if no root row exists, reversing two records with different result payloads also changes `result` and `executions`.

</details>

**💡 Suggested Fix**

Sort records deterministically (e.g., root-first, `started_at`, then `execution_id`), document error precedence, and pick a deterministic terminal status order. Add tests that return the same multi-record run in opposite orders and assert identical result, error, status, and execution ordering for both `get_run` and `wait_run`.

---
*`Run visibility and deterministic aggregation` · confidence 88%*

PR #820, opened 2026-07-22
feat(cli): health-aware list/stop, startup failure diagnostics, port retry, doctor --probe

AbirAbbas:
## Problem

A live field test of the sub-harness flow hit four operational failures in one session, all of the same class: **the platform knew something and didn't say it.**

1. **`af stop` silently killed in-flight work.** A graceful stop terminated a node that had a long-running execution in progress, with no warning — even though running executions are queryable from the control plane.
2. **`af run` startup failures were opaque and non-recoverable.** When a node was assigned a port that looked free but lost the bind race (a just-stopped node's port lingering under mirrored networking), the SDK exited with the strict-port "assigned port unavailable" error and nothing retried. The CLI only printed "did not become ready within 30s"; the real traceback lived in a separate `af logs`.
3. **`af list` trusted the local registry blindly.** Registry status is a claim, not a fact — a node the registry calls "running" can be dead on the control plane (or vice versa) and nothing reconciled the two views.
4. **`af doctor` reported providers as healthy from a PATH check alone.** A provider CLI whose binary exists can still return instant empty completions (broken auth, model outage); doctor never actually probed provider health.

## What changed

Four focused commits, one per failure class:

- **`feat(cli): warn before af stop interrupts running executions`** — before signalling the process, `af stop <node>` queries the control plane for executions still running on the node. Found + TTY → list count/ids/age and prompt to confirm; found + non-TTY → warn and proceed (prior behavior); no executions or `--force` → stop immediately; control plane unreachable → note it and proceed (best-effort). Warnings go to stderr so `--json` stdout stays a clean envelope.
- **`feat(cli): surface af run startup logs and retry once on strict-port bind conflict`** — on any startup failure the last ~15 lines of the node's log are printed inline plus a `Full logs: af logs <node>` pointer. When the failure is the SDK's strict-port bind conflict (detected from the node log), the run path retries **exactly once** on a fresh port (the failed port is reserved first so it is never reused), logging `Port <p> unavailable, retrying on a fresh port`. The port-alloc/start/wait section is refactored into `attemptStart` + `startWithPortRetry` so the retry decision and port-change logic are unit-testable without the real health poll.
- **`feat(cli): reconcile af list registry status with control-plane health`** — `af list` gains a HEALTH column that fetches `GET /api/v1/nodes?show_all=true` (so inactive nodes are included) and reconciles it with each node's registry status. Agreement shows plain health; disagreement is marked `(mismatch)` with an explanatory footer; a node absent from the control plane while the registry calls it running reads `not on control plane (mismatch)`; an unreachable control plane yields `unknown (control plane unreachable)` and never errors. The same `health` / `health_discrepancy` fields are added to `af list --json`.
- **`feat(cli): add af doctor --probe to smoke-test detected provider CLIs`** — new opt-in `--probe` flag runs a minimal one-shot prompt against each **detected** provider CLI (`claude -p`, `codex exec`, `gemini -p`, `opencode run`) with a per-provider 60s timeout, classifying the result as `ok` (non-empty), `empty` (exit 0 but no completion — the silently-broken case a PATH check misses), `error` (non-zero exit, stderr head captured), or `timeout`. Probes run only for providers doctor already detects; without `--probe` the command is unchanged. Help text and output note that a probe consumes a trivial amount of provider quota.

## Validation Contract

Tests are derived from behavior, not implementation. Each contract item maps to named tests:

| Contract item | Tests |
|---|---|
| Stop with running executions: non-TTY warns with count; `--force` does not warn (and never queries); no executions is silent; unreachable proceeds; interactive decline aborts | `TestConfirmStop_NonInteractive_WarnsWithCount`, `TestConfirmStop_Force_NoWarnNoQuery`, `TestConfirmStop_NoRunningExecutions_Silent`, `TestConfirmStop_ControlPlaneUnreachable_Proceeds`, `TestConfirmStop_Interactive_DeclineAborts`, `TestConfirmStop_Interactive_AcceptProceeds`, `TestQueryRunningExecutions_ParsesEnvelope`, `TestQueryRunningExecutions_EmptyServerURL`, `TestReadAffirmative`, `TestFormatExecutionAge` |
| Startup-failure path prints the tail of the node's log and the `af logs` pointer | `TestPrintStartupFailureDiagnostics`, `TestReadLogTailLines` |
| Strict-port failure triggers exactly one retry on a different port; non-conflict failures do not retry; the failed port is excluded | `TestStartWithPortRetry_RetriesOnceOnPortConflict`, `TestStartWithPortRetry_NoRetryOnNonConflictFailure`, `TestStartWithPortRetry_SuccessRunsOnce`, `TestStartWithPortRetry_NoDistinctPortDoesNotRetry`, `TestFreshRetryPort_ExcludesFailedPort`, `TestLogIndicatesPortConflict` |
| `af list` merges the registry with a mocked `/api/v1/nodes`: agreement renders plainly, discrepancy is flagged, unreachable yields `unknown` without erroring | `TestReconcileHealth`, `TestResolveNodeHealth_Merge`, `TestResolveNodeHealth_Unreachable`, `TestFetchControlPlaneNodes_Reachable`, `TestFetchControlPlaneNodes_Unreachable` |
| Doctor probe classification of ok/empty/error/timeout from (exit code, stdout, timed-out) tuples; probes only detected providers; `af doctor` without `--probe` performs no probe | `TestClassifyProbe`, `TestProbeHarnessProvider_RealProcesses`, `TestRunHarnessProbes_SkipsUndetected`, `TestDoctorCommand_NoProbeByDefault` |

## Test results

- `npm ci && npm run build` (control-plane UI): pass
- `GOFLAGS=-buildvcs=false go build ./...`: pass
- `go vet ./...`: pass
- `go test ./internal/cli/... ./internal/core/services/...`: pass (all new spec tests green)
- `gofmt -l` on changed files: empty
- `golangci-lint run`: changed files introduce zero new findings (errcheck counts unchanged vs. base for `doctor.go` and `agent_service.go`)

One unrelated test, `internal/server/TestStartAndStopCoverAdditionalBranches`, fails locally in this sandbox because the admin gRPC listener cannot bind here; it fails identically on the base commit and this branch touches no code in `internal/server`, so it is a pre-existing environment artifact, not a regression.


control-plane/internal/cli/doctor.go
```diff
@@ -108,13 +129,25 @@ Coding agents and skills (e.g. agentfield) should call this once at the
...
 Examples:
   af doctor                  # Pretty human-readable output
   af doctor --json           # Machine-readable JSON for tooling/skills
+  af doctor --probe          # Also smoke-test each detected provider CLI
   af doctor --json | jq      # Pipe to jq for filtering`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			report := buildDoctorReport(controlPlaneURL)
 
+			if probe {
+				report.HarnessProbes = runHarnessProbes(report)
```
AbirAbbas, line 148: > [!NOTE]
> **Advisory — non-blocking.** Safe to merge and address in a follow-up.
>
> _Why non-blocking:_ Bug in new --probe feature does not break existing contracts or cause security/data issues; can be fixed in a follow-up.

Make `--probe` recompute `Recommendation.HarnessUsable`, `HarnessProviders`, and the harness note from providers whose probe status is `ok`.

<sub>🟠 `IMPORTANT` · confidence 99%</sub>

A CLI that returns `empty`, `error`, or `timeout` during a probe does not update `recommendation.harness_providers`, `recommendation.harness_usable`, or the harness note—so zero successful probes still recommend a harness, and mixed results include failed providers. Only `harness_probes` surfaces the actual status, breaking consumers that rely on the recommendation fields.

<details><summary>Evidence</summary>

> Step 1: `NewDoctorCommand` calls `buildDoctorReport` at line 145, which builds `Recommendation` from PATH detection: `HarnessUsable: len(availableHarness) > 0` and `HarnessProviders: availableHarness` (lines 328-333). Step 2: for `af doctor --probe`, line 148 only assigns `report.HarnessProbes = runHarnessProbes(report)`. Step 3: `runHarnessProbes` records statuses such as `empty`, `error`, and `timeout` (lines 172-180), but no code recomputes `Recommendation` or its notes. Step 4: JSON encoding at lines 151-154 and human recommendation rendering at lines 458-466 therefore expose the stale PATH-based recommendation even when every probe failed.

</details>

**💡 Suggested Fix**

When `--probe` is requested, derive `Recommendation.HarnessUsable`, `Recommendation.HarnessProviders`, and the harness-related note from providers whose probe status is `ok` (while retaining `DoctorReport.HarnessProviders` as the raw detection map). Preserve the PATH-based behavior when `--probe` is absent, and add zero-success and mixed-success JSON assertions.

---
*`Machine-consumer semantics` · confidence 99%*

control-plane/internal/cli/stop.go
```diff
@@ -221,6 +286,140 @@ func (as *AgentNodeStopper) markStopped(registry *packages.InstallationRegistry,
...
+// queryRunningExecutions asks the control plane for executions still running on
+// nodeID. An unreachable control plane or an unusable response returns an error;
+// callers treat that as "proceed, best-effort".
+func queryRunningExecutions(serverURL, nodeID string) ([]runningExecution, error) {
+	serverURL = strings.TrimRight(strings.TrimSpace(serverURL), "/")
+	if serverURL == "" {
+		return nil, fmt.Errorf("no control plane URL configured")
+	}
+
+	payload := map[string]interface{}{
```
AbirAbbas, line 316: > [!NOTE]
> **Advisory — non-blocking.** Safe to merge and address in a follow-up.
>
> _Why non-blocking:_ The warning undercounts but does not cause data loss, security issues, or broken contracts; it is a UX improvement opportunity.

Return the decoded total alongside the results and make the warning state the real total, including that only the first 100 IDs are shown.

The guard requests at most 100 executions but discards the server-provided `total`. The query handler computes `total` before applying the limit, so a node with 101+ running executions returns 100 records and `total:101`. The prompt then says only 100 will be interrupted, hiding the remaining execution and undermining the acknowledgement.

<details><summary>Evidence</summary>

> Step 1: `queryRunningExecutions` sends `limit: 100` (stop.go:316-320). Step 2: the control-plane query implementation calculates `total := len(execs)` before truncating `execs` to `req.Limit`, then returns both values (internal/handlers/agentic/query.go:134-149). Step 3: the CLI response type decodes `Data.Total` but returns only `parsed.Data.Results` (stop.go:297-305, 345-349). Step 4: `confirmStopWithRunningExecutions` uses `len(execs)` in both its warning and TTY confirmation (stop.go:375, 386), so 101+ actual executions are represented as 100 and the stop then proceeds after that incomplete confirmation.

</details>

**💡 Suggested Fix**

Return the total alongside results and update the warning to show the real total, noting that only the first 100 IDs are displayed. Alternatively, page until all executions are fetched with a bounded timeout; add a test for `total > len(results)`.

---
*`Stop execution warning correctness` · confidence 98%*

control-plane/internal/cli/doctor.go
```diff
@@ -128,10 +161,104 @@ Examples:
...
+		return outBuf.String(), err.Error(), 1, false
+	}
+	return outBuf.String(), errBuf.String(), 0, false
+}
+
+// classifyProbe maps a probe outcome to a status. Order matters: a timeout is
+// checked before the exit code (a killed process also exits non-zero), and an
+// empty completion on a clean exit is the real-world "silently broken provider"
+// case that a mere PATH check misses.
+func classifyProbe(exitCode int, stdout string, timedOut bool) string {
```
AbirAbbas, line 238: > [!NOTE]
> **Advisory — non-blocking.** Safe to merge and address in a follow-up.
>
> _Why non-blocking:_ Diagnostic tool misreporting does not break build, security, data, or API contracts.

**Classify OpenCode's exit-zero stderr failures as errors**

<sub>🟠 `IMPORTANT` · confidence 99%</sub>

Pass stderr into classification and treat OpenCode's exit-zero stderr errors as errors.

`classifyProbe` ignores stderr, so a zero exit with blank stdout returns `empty` even when OpenCode's stderr contains known error patterns (e.g., “Model not found”). This causes `af doctor --probe` to misreport a real provider failure as `empty` instead of `error`.

<details><summary>Evidence</summary>

> Changed code: `control-plane/internal/cli/doctor.go:187-188` calls `classifyProbe(exitCode, stdout, timedOut)` and drops `stderr`; `doctor.go:238-247` returns `"empty"` for `exitCode == 0` and blank stdout. Other end: `sdk/python/agentfield/harness/providers/opencode.py:325-336` checks `result_text is None`, nonempty `clean_stderr`, and `_OPENCODE_STDERR_ERROR_PATTERNS`, with the comment `opencode sometimes exits 0 even on hard failures like "Model not found"`, then sets `is_error = True` and `failure_type = FailureType.CRASH`.

</details>

**💡 Suggested Fix**

Pass stderr into classification and, before returning `empty` for a zero exit with blank stdout, recognize OpenCode's known stderr error patterns (or use a provider-specific probe/result parser) and classify that outcome as `error`.

---
*`Consistency Verifier` · confidence 99%*

control-plane/internal/core/services/agent_service.go
```diff
@@ -126,37 +170,95 @@ func (as *DefaultAgentService) runAgentGuarded(name string, options domain.RunOp
 			expectedNodeID = metadata.AgentNode.NodeID
 		}
 	}
-	if err := as.waitForAgentNode(port, healthPath, expectedNodeID, nodeReadyTimeout()); err != nil {
-		// Kill the process if it failed to start properly
+
+	if waitErr := as.waitForAgentNode(port, healthPath, expectedNodeID, nodeReadyTimeout()); waitErr != nil {
```
AbirAbbas, line 175: > [!NOTE]
> **Advisory — non-blocking.** Safe to merge and address in a follow-up.
>
> _Why non-blocking:_ The described race from stale logs can cause a spurious retry but does not concretely break builds, security, data integrity, or a public contract; safe to fix later.

Record the log offset before `Start` and classify only output written by the current child. Also match the attempted port in the strict-port message.

The conflict classifier reads the last 40 lines of an append-only log without an attempt boundary, so a stale `"assigned port ... unavailable"` line from a previous run can trigger a retry for an unrelated failure (e.g., import error, health-check timeout). This burns the sole retry and may launch the node on a different port despite the current child never reporting a bind conflict.

<details><summary>Evidence</summary>

> Step 1: The real process manager opens `config.LogFile` with `os.O_CREATE|os.O_WRONLY|os.O_APPEND` in `internal/infrastructure/process/manager.go:46-61`, preserving output from prior runs.
> Step 2: A prior failed run can therefore leave `"assigned port ... unavailable"` in the final 40 lines.
> Step 3: On any current readiness failure, `attemptStart` reads those final 40 lines at `agent_service.go:174-176`; it does not record a pre-start offset or otherwise limit the read to current-process output.
> Step 4: `logIndicatesPortConflict` returns true for any line containing the two substrings (`:255-261`), even when it names another port.
> Step 5: `startWithPortRetry` receives `conflict=true` and executes a fresh-port retry (`:190-203`) even if the current failure was, for example, an import error or a non-port health-check failure.

</details>

**💡 Suggested Fix**

Capture the log file size immediately before `Start` and inspect only bytes appended after that offset (or use a fresh per-attempt log). Require the detected strict-port message to identify the attempted port. Add the requested stale-line-plus-non-port-readiness-failure test.

---
*`Startup retry behavioral contract` · confidence 98%*

control-plane/internal/cli/stop.go
```diff
@@ -221,6 +286,140 @@ func (as *AgentNodeStopper) markStopped(registry *packages.InstallationRegistry,
...
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("control plane returned status %d", resp.StatusCode)
+	}
+
+	var parsed executionsQueryResponse
+	if err := json.NewDecoder(io.LimitReader(resp.Body, 1<<20)).Decode(&parsed); err != nil {
+		return nil, err
+	}
+	return parsed.Data.Results, nil
```
AbirAbbas, line 349: > [!NOTE]
> **Advisory — non-blocking.** Safe to merge and address in a follow-up.
>
> _Why non-blocking:_ Incomplete warning is a UX issue, not a security vulnerability, data loss, or contract break.

**Fix the stop guard to check `total` and handle pagination so operators see all running executions before confirming.**

<sub>🟠 `IMPORTANT` · confidence 98%</sub>

The CLI requests only 100 executions and ignores `total`. If there are more than 100, the warning/prompt is incomplete, and an operator may approve the stop without realizing all running executions will be interrupted.

<details><summary>Evidence</summary>

> Changed CLI (`control-plane/internal/cli/stop.go:316-320,345-349`): `"limit": 100`, then `json.NewDecoder(io.LimitReader(resp.Body, 1<<20)).Decode(&parsed)` and `return parsed.Data.Results, nil`; it does not inspect `parsed.OK` or `parsed.Data.Total`. Serializer (`control-plane/internal/handlers/agentic/helpers.go:37-48`) sends successful replies as `AgenticResponse{OK: true, Data: data}` with HTTP 200, so the success envelope matches. But executions query (`control-plane/internal/handlers/agentic/query.go:134-150`) sets `total := len(execs)`, then truncates with `if len(execs) > req.Limit { execs = execs[:req.Limit] }`, and returns both `"results": execs` and `"total": total`. Thus `total` can exceed 100 while the CLI only returns the first 100 results.

</details>

**💡 Suggested Fix**

After decoding, reject `!parsed.OK` and either page through executions until all `total` results are collected or treat `parsed.Data.Total > len(parsed.Data.Results)` as an incomplete check and warn/proceed according to the existing best-effort failure path.

---
*`Consistency Verifier` · confidence 98%*

control-plane/internal/cli/doctor.go
```diff
@@ -78,19 +81,37 @@ var providerEnvVars = []struct {
...
-	Binary string // executable name to look up on PATH
+	Name      string   // value passed to provider= in app.harness()
+	Binary    string   // executable name to look up on PATH
+	ProbeArgs []string // minimal one-shot invocation used by `--probe`
 }{
-	{Name: "claude-code", Binary: "claude"},
-	{Name: "codex", Binary: "codex"},
-	{Name: "gemini", Binary: "gemini"},
-	{Name: "opencode", Binary: "opencode"},
+	{Name: "claude-code", Binary: "claude", ProbeArgs: []string{"-p", "Say OK"}},
```
AbirAbbas, line 88: > [!NOTE]
> **Advisory — non-blocking.** Safe to merge and address in a follow-up.
>
> _Why non-blocking:_ This is a false negative in a diagnostic tool, not a production-breaking issue.

<sub>🟠 `IMPORTANT` · confidence 98%</sub>

Remove the PATH-based `claude` binary check from `claude-code` harness health in `af doctor`.

`af doctor` treats `claude-code` as usable only when a global `claude` executable is on PATH, but the actual Python `app.harness(provider="claude-code")` implementation uses the `claude_agent_sdk` package and explicitly does not require that binary. Thus a valid Claude Code harness installation can be reported unavailable (and cannot be probed) solely because `claude` is absent from PATH.

<details><summary>Evidence</summary>

> Changed location, `control-plane/internal/cli/doctor.go:88`: `{Name: "claude-code", Binary: "claude", ProbeArgs: []string{"-p", "Say OK"}},`; the same file executes `checkTool(h.Binary, "--version")` at line 282. Dispatch/preflight source, `control-plane/internal/cli/harness_doctor.go:36-39`: `// claude-code has no Binary: the Python provider runs on the` / `// claude_agent_sdk pip package (which bundles its own CLI), not on a` / `// globally installed \`claude\` binary. See claudeCodeHealth.` / `{Name: "claude-code", InstallCommand: "pip install 'agentfield[harness-claude]'", AuthEnvVars: []string{"ANTHROPIC_API_KEY"}},`. The Python factory also accepts `claude-code` and constructs `ClaudeCodeProvider()` without a binary argument in `sdk/python/agentfield/harness/providers/_factory.py`.

</details>

control-plane/internal/core/services/agent_service.go
```diff
@@ -104,20 +104,64 @@ func (as *DefaultAgentService) runAgentGuarded(name string, options domain.RunOp
 		}
 	}
 
+	// 4-5. Start the process and wait for readiness. Retry exactly once on a
+	// fresh port if the node exits with a strict-port bind conflict: the
+	// assigned port looked free but the bind lost a race (commonly a
+	// just-stopped node's port lingering on mirrored-network setups), and the
+	// SDK exits fast so the control plane can reallocate — which is what we do.
+	pid, port, startErr := as.startWithPortRetry(port, func(p int) (int, error, bool) {
```
AbirAbbas: > [!CAUTION]
> **Must-fix before merge.** Silent reassignment of an explicit --port violates the CLI contract and can break caller firewall rules or service discovery.

Prevent `--port` values from being silently reassigned on bind failure.

**Explicit `--port` is silently reassigned on bind failure**
<sub>🟠 `IMPORTANT` · confidence 96%</sub>

`startWithPortRetry` unconditionally replaces the port with a free port from `FindFreePort(8001)` on any bind error, even when the port was explicitly requested via `--port`. This causes the agent to run on an unexpected port and breaks callers that rely on the specified port for firewall rules, reverse proxies, or service discovery.

<details><summary>Evidence</summary>

> Step 1: `internal/cli/commands/run.go:80` constructs `domain.RunOptions{Port: port, ...}` from the user's `--port` value and calls `RunAgent`.
> Step 2: `runAgentGuarded` assigns `port := options.Port` at `agent_service.go:99`, so an explicit value is used for the first `attemptStart`.
> Step 3: On a strict-port failure, `agent_service.go:112` calls `startWithPortRetry` without passing whether the port was user-supplied.
> Step 4: `startWithPortRetry` calls `freshRetryPort`, which unconditionally calls `FindFreePort(8001)` (`agent_service.go:195-210`), then starts the node on that result (`:202`).
> Step 5: The command can therefore return success and persist/report a port different from the one explicitly requested.

</details>

**💡 Suggested Fix**

Pass retry eligibility from `runAgentGuarded` and disable automatic reassignment when `options.Port != 0` (return the strict-port startup error after cleanup). Add an end-to-end `RunAgent` test asserting that an explicit port is never replaced.

---
*`Startup retry behavioral contract` · confidence 96%*

control-plane/internal/core/services/agent_service.go
```diff
@@ -126,37 +170,95 @@ func (as *DefaultAgentService) runAgentGuarded(name string, options domain.RunOp
 			expectedNodeID = metadata.AgentNode.NodeID
 		}
 	}
-	if err := as.waitForAgentNode(port, healthPath, expectedNodeID, nodeReadyTimeout()); err != nil {
-		// Kill the process if it failed to start properly
+
+	if waitErr := as.waitForAgentNode(port, healthPath, expectedNodeID, nodeReadyTimeout()); waitErr != nil {
+		// Read the log before killing so a strict-port exit is still visible.
+		conflict := logIndicatesPortConflict(readLogTailLines(agentNode.Runtime.LogFile, 40))
 		if stopErr := as.processManager.Stop(pid); stopErr != nil {
```
AbirAbbas, line 178: > [!NOTE]
> **Advisory — non-blocking.** Safe to merge and address in a follow-up.
>
> _Why non-blocking:_ The issue describes a potential resource leak under specific failure conditions, but does not demonstrate data loss, security vulnerability, or API breakage.

**Do not retry until the first child is terminated; treat a failed `Stop` as fatal.**  
<sub>🟠 `IMPORTANT` · confidence 96%</sub>

If `processManager.Stop` fails during cleanup, `attemptStart` returns a startup error but preserves the `conflict` flag, so `startWithPortRetry` launches a second child. The first may still be live, and only the second PID is persisted—leaving an untracked process and potentially two concurrent node processes.

**💡 Suggested Fix**  
Make a failed `Stop` terminal for the retry path; confirm the child is no longer running before allowing another attempt. Add a test with a strict-conflict failure plus a failing `Stop`, asserting no second `Start` occurs and no runtime state is recorded.

<details><summary>Evidence</summary>

> Step 1: `attemptStart` starts a child via `processManager.Start` (`agent_service.go:160-163`).  
> Step 2: If readiness fails, it calls `processManager.Stop(pid)` (`:174-177`).  
> Step 3: If `Stop` returns an error, `attemptStart` returns a startup error but preserves the previously computed `conflict` flag (`:177-179`), so there is no proof that the child is gone.  
> Step 4: `startWithPortRetry` checks only `err` and `conflict` (`:190-195`), then invokes a second attempt whenever `conflict` is true (`:202`).  
> Step 5: If that second attempt succeeds, `runAgentGuarded` writes runtime state only for its PID and retry port (`:124-126`), leaving the first potentially live child unrecorded. A stale strict-port line makes this scenario reachable even when the first child itself did not exit for a port conflict.

</details>

---
*Startup retry behavioral contract* · confidence 96%

control-plane/internal/cli/list.go
```diff
@@ -84,6 +95,135 @@ func runListCommandJSON() error {
...
+		}
+		if n, ok := cpNodes[cand]; ok {
+			return n, true
+		}
+	}
+	for _, cand := range candidates {
+		if cand == "" {
+			continue
+		}
+		for id, n := range cpNodes {
```
AbirAbbas, line 176: > [!NOTE]
> **Advisory — non-blocking.** Safe to merge and address in a follow-up.
>
> _Why non-blocking:_ Incorrect node association in a CLI list command does not cause security, data loss, or API breakage; it is a correctness bug that can be fixed later.

Make `lookupControlPlaneNode` reject ambiguous equivalence matches instead of returning the first map entry.

<sub>🟠 `IMPORTANT` · confidence 96%</sub>

Registration allows distinct node IDs that differ only by case, hyphens, or underscores (e.g., `foo-bar`, `foo_bar`, `FOO_BAR`). The fallback lookup treats them as equivalent via `NodeIDsEquivalent`, so a candidate without an exact match can be associated with an arbitrary registered node, depending on map iteration order.

<details><summary>Evidence</summary>

> Changed lookup: `for id, n := range cpNodes { if packages.NodeIDsEquivalent(id, cand) { return n, true } }` (control-plane/internal/cli/list.go:176-179). Equivalence folds case and replaces every `-` with `_`: `return strings.ToLower(strings.ReplaceAll(s, "-", "_"))` (control-plane/internal/packages/node_identity.go:24-28). Registration validates only that ID is nonempty (`validate:"required,min=1"`, control-plane/pkg/types/types.go:159) and persists with `ON CONFLICT(id, version)`, which only conflates exact IDs (control-plane/internal/storage/local.go:4464-4469).

</details>

**💡 Suggested Fix**

Either enforce a canonical ID form at registration and reject collisions, or make `lookupControlPlaneNode` accept an equivalence match only when it is unique; otherwise report an ambiguous match error.

---
*`Consistency Verifier` · confidence 96%*

control-plane/internal/cli/list.go
```diff
@@ -63,14 +69,19 @@ func runListCommandJSON() error {
...
 		node := map[string]interface{}{
-			"name":        name,
-			"version":     pkg.Version,
-			"status":      pkg.Status,
-			"description": pkg.Description,
+			"name":               name,
+			"version":            pkg.Version,
+			"status":             pkg.Status,
+			"description":        pkg.Description,
+			"health":             h.Display,
```
AbirAbbas, line 83: > [!NOTE]
> **Advisory — non-blocking.** Safe to merge and address in a follow-up.
>
> _Why non-blocking:_ The documentation mismatch in the README does not break any existing functionality; the Desktop integration is not yet implemented.

Switch the Desktop integration from `af list -o json` to `af list --json`.

<sub>🟠 `IMPORTANT` · confidence 96%</sub>

The `-o` short flag is rejected by Cobra, so the Desktop integration will not receive the JSON envelope or the new `health`/`health_discrepancy` fields. Without them, the integration cannot distinguish registry `status` from control-plane health states like unreachable or mismatch.

<details><summary>Evidence</summary>

> Changed end, control-plane/internal/cli/list.go:83-84: `"health": h.Display,` and `"health_discrepancy": h.Discrepancy,`; control-plane/internal/cli/list.go:45: `cmd.Flags().BoolVar(&listJSON, "json", false, "Emit a machine-readable JSON envelope instead of the table")`. Other end, desktop/README.md:226-228: `once af list -o json lands, the app should shell out to the CLI instead`.

</details>

**💡 Suggested Fix**

Update the Desktop integration and documentation to invoke `af list --json`, mapping `status` separately from `health` and `health_discrepancy`, including `unknown (control plane unreachable)` and `(mismatch)` values.

---
*`Consistency Verifier` · confidence 96%*

control-plane/internal/cli/stop.go
```diff
@@ -130,6 +179,22 @@ func (as *AgentNodeStopper) StopAgentNode(agentNodeName string) error {
 		}
 	}
 
+	// The process is genuinely ours and running — a graceful stop from here on
+	// will interrupt whatever it is doing. Warn about (or, on a terminal,
+	// confirm) any in-flight executions before we signal it, so a long build or
+	// analysis is not silently killed. Resolve the manifest node_id for the
+	// query since the control plane keys executions by node_id, which may differ
+	// from the registry install name.
+	queryID := agentNodeName
```
AbirAbbas, line 188: > [!NOTE]
> **Advisory — non-blocking.** Safe to merge and address in a follow-up.
>
> _Why non-blocking:_ The finding describes a missing warning in a specific configuration scenario, not a security vulnerability, data loss, build break, or API contract break.

**Resolve registered node ID in stop guard when manifest omits node_id**

<sub>🟠 `IMPORTANT` · confidence 94%</sub>

When a manifest omits `agent_node.node_id`, `stop.go` queries executions by install name, but executions are stored under the ID the agent supplies at registration. If the agent registers with a different ID, `confirmStopWithRunningExecutions` finds no executions and allows the stop, missing in-flight work.

<details><summary>Evidence</summary>

> Changed end, control-plane/internal/cli/stop.go:188-192: `queryID := agentNodeName`; `if md, err := packages.ParsePackageMetadata(agentNode.Path); err == nil && md.AgentNode.NodeID != "" { queryID = md.AgentNode.NodeID }`; `if !as.confirmStopWithRunningExecutions(queryID) {`. Other ends: docs/installing-agent-nodes.md:37-39 says only `name`, `version`, and a start method are required and "everything else is optional"; `AgentNodeConfig` defines `NodeID string `yaml:"node_id"`` (control-plane/internal/packages/installer.go:146-150), and `ParsePackageMetadata` only YAML-unmarshals it (lines 664-667), so an omitted field remains empty. Registration unmarshals the agent-supplied node into `newNode` (control-plane/internal/handlers/nodes_register.go:353,362) and persists it with `storageProvider.RegisterAgent(ctx, &newNode)` (line 654). Execution creation then uses the persisted registered ID: `AgentNodeID: agent.ID` (control-plane/internal/handlers/execute.go:1491-1497), which is inserted into the `agent_node_id` column (control-plane/internal/storage/execution_records.go:33-43).

</details>

**💡 Suggested Fix**

In `stop.go`, when `md.AgentNode.NodeID` is empty, look up the registered node ID from the agent's registration identity before calling `confirmStopWithRunningExecutions`, or enforce at registration that the install name is used as the node ID.

---
*`Consistency Verifier` · confidence 94%*

control-plane/internal/cli/doctor.go
```diff
@@ -128,10 +161,104 @@ Examples:
...
 	return cmd
 }
 
+// runHarnessProbes smoke-tests every provider CLI that doctor detected on PATH.
+// Providers that were not detected are skipped entirely — probing runs only for
+// what doctor already reports as available.
+func runHarnessProbes(report DoctorReport) map[string]HarnessProbeResult {
+	results := map[string]HarnessProbeResult{}
+	for _, h := range harnessProviders {
+		if !report.HarnessProviders[h.Name].Available {
```
AbirAbbas, line 175: > [!NOTE]
> **Advisory — non-blocking.** Safe to merge and address in a follow-up.
>
> _Why non-blocking:_ The issue is about defensive validation in a diagnostic tool, not a proven production impact.

Only mark a harness provider as available for probing after the version command succeeds.

`buildDoctorReport` sets `Available` from `exec.LookPath` alone; a failed or timed-out `--version` invocation leaves the bit `true`, so `runHarnessProbes` executes the binary and consumes quota even for a mismatched or broken CLI. The required meaning of `Available` does not hold.

<details><summary>Evidence</summary>

> Changed end, `control-plane/internal/cli/doctor.go:174-178`:
> `for _, h := range harnessProviders {`
> `    if !report.HarnessProviders[h.Name].Available {`
> `        continue`
> `    }`
> `    results[h.Name] = probeHarnessProvider(h.Name, h.Binary, h.ProbeArgs, harnessProbeTimeout)`
> `}`
> 
> Other end, `control-plane/internal/cli/doctor.go:281-285`:
> `for _, h := range harnessProviders {`
> `    status := checkTool(h.Binary, "--version")`
> `    report.HarnessProviders[h.Name] = status`
> `    if status.Available { ... }`
> `}`
> 
> And `checkTool`, `control-plane/internal/cli/doctor.go:341-356`, sets availability solely after path lookup:
> `path, err := exec.LookPath(bin)`
> `if err != nil { return ToolStatus{Available: false} }`
> `status := ToolStatus{Available: true, Path: path}`
> ...
> `out, err := cmd.CombinedOutput()`
> `if err == nil { status.Version = first }`
> A failing or timed-out `--version` invocation leaves `Available: true`.

</details>

---
*`Consistency Verifier`* · confidence 93%

control-plane/internal/cli/list.go
```diff
@@ -84,6 +95,135 @@ func runListCommandJSON() error {
...
+		if registryRunning {
+			return "not on control plane (mismatch)", true
+		}
+		return "—", false
+	}
+	if cpHealth == "" {
+		cpHealth = string(types.HealthStatusUnknown)
+	}
+	// Active and degraded both mean the node is registered and heartbeating.
+	cpAlive := cpHealth == string(types.HealthStatusActive) || cpHealth == string(types.HealthStatusDegraded)
```
AbirAbbas, line 153: > [!NOTE]
> **Advisory — non-blocking.** Safe to merge and address in a follow-up.
>
> _Why non-blocking:_ The CLI misclassification does not break the build, security, data integrity, or public API contracts; it is an operational improvement opportunity, not an immediate production blocker.

<sub>🟠 `IMPORTANT` · confidence 90%</sub>

Only treat `HealthStatusActive` as alive; do not treat `HealthStatusDegraded` as a heartbeat signal.

The server never sets `health_status` to `degraded` for a live node; it maps lifecycle degraded to `AgentStateActive` and then to `HealthStatusActive`. Treating `degraded` as alive masks a mismatch when a stale or external `health_status: degraded` is present.

<details><summary>Evidence</summary>

> Changed code, control-plane/internal/cli/list.go:152-154: `// Active and degraded both mean the node is registered and heartbeating.` followed by `cpAlive := cpHealth == string(types.HealthStatusActive) || cpHealth == string(types.HealthStatusDegraded)`. Server heartbeat logic, control-plane/internal/services/status_manager.go:480-497 maps `case types.AgentStatusDegraded: derivedState = types.AgentStateActive` and persists via `UpdateAgentStatus`; that method sets `newStatus.HealthStatus = newStatus.ToLegacyHealthStatus()` (lines 397-400). `ToLegacyHealthStatus` in control-plane/pkg/types/types.go:416-424 maps only `AgentStateActive` to `HealthStatusActive`; it maps inactive/stopping to inactive and all other states to unknown. The reconciliation logic likewise selects `HealthStatusActive` for a fresh heartbeat (control-plane/internal/services/status_manager.go:797-802).

</details>

**💡 Suggested Fix**

Treat only `HealthStatusActive` as the live value for `health_status`. If lifecycle degradation needs to be shown as live, fetch and evaluate `lifecycle_status` separately rather than interpreting the legacy health field's `degraded` value as a heartbeat signal.

control-plane/internal/cli/stop.go
```diff
@@ -221,6 +286,140 @@ func (as *AgentNodeStopper) markStopped(registry *packages.InstallationRegistry,
...
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("control plane returned status %d", resp.StatusCode)
+	}
+
+	var parsed executionsQueryResponse
```
AbirAbbas, line 345: > [!NOTE]
> **Advisory — non-blocking.** Safe to merge and address in a follow-up.
>
> _Why non-blocking:_ Missing validation of the API response envelope reduces the safety warning but does not cause data loss, security breach, or contract breakage; it can be addressed in a follow-up.

**Check `parsed.OK` and validate `data.results` before returning the results slice.**

`queryRunningExecutions` decodes the HTTP 200 body but never verifies the envelope’s `ok` field or that `data.results` is present. A 200 response with `{"ok":false,"error":{...}}` or `{"ok":true,"data":{}}` yields a nil `Data.Results`, which `confirmStopWithRunningExecutions` treats as zero running executions and silently stops the node, bypassing the unusable-response warning.

<details><summary>Evidence</summary>

> Step 1: A verified live-node `AgentNodeStopper.StopAgentNode` calls `confirmStopWithRunningExecutions(queryID)` at `stop.go:192` before any shutdown/signal action. Step 2: `confirmStopWithRunningExecutions` calls `queryRunningExecutions` at line 366 and only emits the unusable-response warning when that function returns a non-nil error (lines 367-370). Step 3: `queryRunningExecutions` accepts status 200 (lines 340-342), decodes into `executionsQueryResponse`, and returns `parsed.Data.Results` directly at line 349 without checking `parsed.OK` (declared at line 300) or required data fields. Step 4: `respondOK` defines success as `{"ok":true,"data":...}` while `respondError` defines error envelopes as `{"ok":false,"error":...}` (`handlers/agentic/helpers.go:12-16, 37-38, 52-57`); absent JSON fields decode to zero values, so either an `ok:false` 200 error envelope or incomplete 200 response yields a nil results slice. Step 5: line 371 treats that nil slice as no running executions and proceeds silently. The focused tests cover successful envelopes, an empty URL, and transport failure, but contain no `ok:false`, error-envelope, missing-data, or missing-results case.

</details>

---
*`Stop guard control-plane protocol boundary` · confidence 97%*

control-plane/internal/cli/stop.go
```diff
@@ -9,14 +10,18 @@ import (
...
+	"strings"
 	"time"
 
 	"github.com/Agent-Field/agentfield/control-plane/internal/packages"
 	"github.com/spf13/cobra"
 	"gopkg.in/yaml.v3"
 )
 
-var stopJSON bool
+var (
```
AbirAbbas, line 21: > [!NOTE]
> **Advisory — non-blocking.** Safe to merge and address in a follow-up.
>
> _Why non-blocking:_ The issue requires an unusual embedding pattern and does not demonstrate a concrete production impact; it is a code quality improvement.

Fix the `--force` flag to not persist across command instances so ordinary stop prompts are never bypassed.

Because every command shares the same package-level `stopForce` variable, executing a forced stop sets it true, and a later ordinary stop reads that value instead of its default, skipping the execution guard at stop.go:361-364.

<sub>🔵 `SUGGESTION` · confidence 95%</sub>

<details><summary>Evidence</summary>

> Step 1: `NewStopCommand` binds every command instance's `--force` flag to the single package variable `stopForce` (stop.go:21-24, 51). Step 2: an embedding can construct `forced := NewStopCommand()` and `ordinary := NewStopCommand()`, then execute `forced` with `--force`; Cobra/pflag writes `true` through the shared pointer. Step 3: executing `ordinary` without `--force` does not parse a value that resets that shared pointer, and `runStopCommand` reads it into `AgentNodeStopper.Force` (stop.go:55-66). Step 4: `confirmStopWithRunningExecutions` returns immediately when `Force` is true (stop.go:361-364), so the ordinary invocation neither queries nor warns before the verified node is stopped. `resetLifecycleFlags` also resets `stopJSON` but omits `stopForce` (node_json_test.go:18-26), leaving this lifecycle behavior untested.

</details>

**💡 Suggested Fix**

Make `force` and `json` command-instance state (for example, local variables captured by the command's `RunE`), or reset/read flags from `cmd.Flags()` per execution. Add a lifecycle test that constructs both commands before executing the forced one, then verifies the ordinary command queries/prompts; also reset `stopForce` in the shared test helper if globals remain.

---
*`Stop execution warning correctness` · confidence 95%*

AbirAbbas (commented):
## 🔴 PR-AF Review — **Merge Blocked — Must-Fix Found**

> 🚫 **Merge blocked.** 1 must-fix issue found by automated review.

> **19 findings** · 🚫 1 blocking · 💬 18 advisory · 🔴 0 critical · 🟠 14 important · 🔵 4 suggestions · ⚪ 1 nitpicks

<details>

## Problem

A live field test of the sub-harness flow hit four operational failures in one session, all of the same class: **the platform knew something and didn't say it.**

1. **`af stop` silently killed in-flight work.** A graceful stop terminated a node that had a long-running execution in progress, with no warning — even though running executions are queryable from the control plane.
2. **`af run` startup failures were opaque and non-recoverable.** When a node was assigned a port that looked free but lost the bind race (a just-stopped node's port lingering under mirrored networking), the SDK exited with the strict-port "assigned port unavailable" error and nothing retried. The CLI only printed "did not become ready within 30s"; the real traceback lived in a separate `af logs`.
3. **`af list` trusted the local registry blindly.** Registry status is a claim, not a fact — a node the registry calls "running" can be dead on the control plane (or vice versa) and nothing reconciled the two views.
4. **`af doctor` reported providers as healthy from a PATH check alone.** A provider CLI whose binary exists can still return instant empty completions (broken auth, model outage); doctor never actually probed provider health.

## What changed

Four focused commits, one per failure class:

- **`feat(cli): warn before af stop interrupts running executions`** — before signalling the process, `af stop <node>` queries the control plane for executions still running on the node. Found + TTY → list count/ids/age and prompt to confirm; found + non-TTY → warn and proceed (prior behavior); no executions or `--force` → stop immediately; control plane unreachable → note it and proceed (best-effort). Warnings go to stderr so `--json` stdout stays a clean envelope.
- **`feat(cli): surface af run startup logs and retry once on strict-port bind conflict`** — on any startup failure the last ~15 lines of the node's log are printed inline plus a `Full logs: af logs <node>` pointer. When the failure is the SDK's strict-port bind conflict (detected from the node log), the run path retries **exactly once** on a fresh port (the failed port is reserved first so it is never reused), logging `Port <p> unavailable, retrying on a fresh port`. The port-alloc/start/wait section is refactored into `attemptStart` + `startWithPortRetry` so the retry decision and port-change logic are unit-testable without the real health poll.
- **`feat(cli): reconcile af list registry status with control-plane health`** — `af list` gains a HEALTH column that fetches `GET /api/v1/nodes?show_all=true` (so inactive nodes are included) and reconciles it with each node's registry status. Agreement shows plain health; disagreement is marked `(mismatch)` with an explanatory footer; a node absent from the control plane while the registry calls it running reads `not on control plane (mismatch)`; an unreachable control plane yields `unknown (control plane unreachable)` and never errors. The same `health` / `health_discrepancy` fields are added to `af list --json`.
- **`feat(cli): add af doctor --probe to smoke-test detected provider CLIs`** — new opt-in `--probe` flag runs a minimal one-shot prompt against each **detected** provider CLI (`claude -p`, `codex exec`, `gemini -p`, `opencode run`) with a per-provider 60s timeout, classifying the result as `ok` (non-empty), `empty` (exit 0 but no completion — the silently-broken case a PATH check misses), `error` (non-zero exit, stderr head captured), or `timeout`. Probes run only for providers doctor already detects; without `--probe` the command is unchanged. Help text and output note that a probe consumes a trivial amount of provider quota.

## Validation Contract

Tests are derived from behavior, not implementation. Each contract item maps to named tests:

| Contract item | Tests |
|---|---|
| Stop with running executions: non-TTY warns with count; `--force` does not warn (and never queries); no executions is silent; unreachable proceeds; interactive decline aborts | `TestConfirmStop_NonInteractive_WarnsWithCount`, `TestConfirmStop_Force_NoWarnNoQuery`, `TestConfirmStop_NoRunningExecutions_Silent`, `TestConfirmStop_ControlPlaneUnreachable_Proceeds`, `TestConfirmStop_Interactive_DeclineAborts`, `TestConfirmStop_Interactive_AcceptProceeds`, `TestQueryRunningExecutions_ParsesEnvelope`, `TestQueryRunningExecutions_EmptyServerURL`, `TestReadAffirmative`, `TestFormatExecutionAge` |
| Startup-failure path prints the tail of the node's log and the `af logs` pointer | `TestPrintStartupFailureDiagnostics`, `TestReadLogTailLines` |
| Strict-port failure triggers exactly one retry on a different port; non-conflict failures do not retry; the failed port is excluded | `TestStartWithPortRetry_RetriesOnceOnPortConflict`, `TestStartWithPortRetry_NoRetryOnNonConflictFailure`, `TestStartWithPortRetry_SuccessRunsOnce`, `TestStartWithPortRetry_NoDistinctPortDoesNotRetry`, `TestFreshRetryPort_ExcludesFailedPort`, `TestLogIndicatesPortConflict` |
| `af list` merges the registry with a mocked `/api/v1/nodes`: agreement renders plainly, discrepancy is flagged, unreachable yields `unknown` without erroring | `TestReconcileHealth`, `TestResolveNodeHealth_Merge`, `TestResolveNodeHealth_Unreachable`, `TestFetchControlPlaneNodes_Reachable`, `TestFetchControlPlaneNodes_Unreachable` |
| Doctor probe classification of ok/empty/error/timeout from (exit code, stdout, timed-out) tuples; probes only detected providers; `af doctor` without `--probe` performs no probe | `TestClassifyProbe`, `TestProbeHarnessProvider_RealProcesses`, `TestRunHarnessProbes_SkipsUndetected`, `TestDoctorCommand_NoProbeByDefault` |

## Test results

- `npm ci && npm run build` (control-plane UI): pass
- `GOFLAGS=-buildvcs=false go build ./...`: pass
- `go vet ./...`: pass
- `go test ./internal/cli/... ./internal/core/services/...`: pass (all new spec tests green)
- `gofmt -l` on changed files: empty
- `golangci-lint run`: changed files introduce zero new findings (errcheck counts unchanged vs. base for `doctor.go` and `agent_service.go`)

One unrelated test, `internal/server/TestStartAndStopCoverAdditionalBranches`, fails locally in this sandbox because the admin gRPC listener cannot bind here; it fails identically on the base commit and this branch touches no code in `internal/server`, so it is a pre-existing environment artifact, not a regression.

</details>

### Key Findings

**1 issue(s) should be addressed before merge:**

- 🟠 **Explicit `--port` is silently reassigned on bind failure** (`control-plane/internal/core/services/agent_service.go:112`) — A caller that explicitly passes `--port P` can receive a successful run on an unrelated port. Gate: Silent reassignment of an explicit --port violates the CLI contract and can break caller firewall rules or service discovery.

**18 advisory finding(s) surfaced as non-blocking:**

- 🟠 Failed probes do not affect harness recommendation (`control-plane/internal/cli/doctor.go:148`)
- 🟠 Execution warning silently undercounts results beyond the fixed limit (`control-plane/internal/cli/stop.go:316`)
- 🟠 Classify OpenCode's exit-zero stderr failures as errors (`control-plane/internal/cli/doctor.go:238`)
- 🟠 Stale append-only logs can trigger a retry for the wrong attempt (`control-plane/internal/core/services/agent_service.go:174`)
- 🟠 Stop warning silently omits running executions beyond the first page (`control-plane/internal/cli/stop.go:349`)
- … and 13 more (see All Findings by Severity)

**Files with findings:** `control-plane/internal/cli/doctor.go`, `control-plane/internal/cli/list.go`, `control-plane/internal/cli/stop.go`, `control-plane/internal/core/services/agent_service.go`

<details>

#### 🟠 Important (14)

- **Failed probes do not affect harness recommendation** `control-plane/internal/cli/doctor.go:148`
- **Execution warning silently undercounts results beyond the fixed limit** `control-plane/internal/cli/stop.go:316`
- **Classify OpenCode's exit-zero stderr failures as errors** `control-plane/internal/cli/doctor.go:238`
- **Stale append-only logs can trigger a retry for the wrong attempt** `control-plane/internal/core/services/agent_service.go:174`
- **Stop warning silently omits running executions beyond the first page** `control-plane/internal/cli/stop.go:349`
- **Do not require a global `claude` binary for the SDK-backed `claude-code` harness** `control-plane/internal/cli/doctor.go:88`
- **Explicit `--port` is silently reassigned on bind failure** `control-plane/internal/core/services/agent_service.go:112`
- **Retry can start a second child after first-child cleanup fails** `control-plane/internal/core/services/agent_service.go:177`
- **Equivalence lookup can conflate distinct registered nodes** `control-plane/internal/cli/list.go:176`
- **Desktop list integration names an unsupported JSON flag** `control-plane/internal/cli/list.go:83`
- **Stop guard can miss executions for manifests without node_id** `control-plane/internal/cli/stop.go:188`
- **Do not use PATH-only availability as authorization to run provider probes** `control-plane/internal/cli/doctor.go:175`
- **Do not classify legacy health_status "degraded" as alive** `control-plane/internal/cli/list.go:153`
- **Validate the agentic query envelope before treating it as empty** `control-plane/internal/cli/stop.go:345`

#### 🔵 Suggestion (4)

- **A previous forced command can make an ordinary stop skip the guard** `control-plane/internal/cli/stop.go:21`
- **Unforced stop can inherit `--force` from an earlier execution** `control-plane/internal/cli/stop.go:51`
- **A failed reservation can suppress the only fresh-port retry** `control-plane/internal/core/services/agent_service.go:209`
- **Control-plane authentication and response failures are mislabeled as unreachable** `control-plane/internal/cli/list.go:204`

#### ⚪ Nitpick (1)

- **Failed-port reservation leaks into future runs** `control-plane/internal/core/services/agent_service.go:208`

</details>

<details>

**Dimensions Analyzed (6):**

- **Stop guard preserves its safety decision under complete and failed control-plane responses** — 3 file(s)
- **Strict-port retry must exclude the failed port under the real PortManager contract** — 3 file(s)
- **Strict-port retry neither violates requested-port expectations nor retries on stale evidence** — 2 file(s)
- **Stop guard must honor the agentic query response envelope** — 2 file(s)
- **New health/probe output does not report a usable provider or reachable control plane when its own probe disproves it** — 4 file(s)
- **Cobra command construction must not retain --force across executions** — 3 file(s)

- **Semantic** — 3 dimension(s), 87% coverage confidence
- **Mechanical** — 3 dimension(s), 90% coverage confidence
- **Systemic** — 2 dimension(s), 88% coverage confidence

- **15** finding(s) adversarially tested: 6 confirmed, 9 challenged

</details>

<details>

| Metric | Value |
|--------|-------|
| Duration | 796.1s |
| Agent invocations | 32 |
| Coverage iterations | 1 |
| Budget exhausted | No |
| PR type | feature |
| Complexity | medium |

</details>

Review ID: `rev_cf4d09b58367`

<br>
<div align="right">
    AgentField PR-AF
  </a>
</div>

control-plane/internal/cli/stop.go
```diff
@@ -28,15 +33,22 @@ func NewStopCommand() *cobra.Command {
...
   af stop data-analyzer
-  af stop email-helper --json`,
+  af stop email-helper --json
+  af stop email-helper --force`,
 		Args: cobra.ExactArgs(1),
 		RunE: runStopCommand,
 	}
 
 	cmd.Flags().BoolVar(&stopJSON, "json", false, "Emit a machine-readable JSON envelope instead of progress output")
+	cmd.Flags().BoolVar(&stopForce, "force", false, "Stop immediately without warning about or confirming in-flight executions")
```
AbirAbbas, line 51: > [!NOTE]
> **Advisory — non-blocking.** Safe to merge and address in a follow-up.
>
> _Why non-blocking:_ The flag leak requires programmatic reuse of command instances, not standard CLI usage; no concrete production impact.

Make the `force` flag local to each command instance to prevent state leaking between invocations.

<sub>🔵 `SUGGESTION` · confidence 95%</sub>

When `NewStopCommand` binds the `force` flag to a package-level variable, reusing the command or constructing multiple command trees before execution causes a later unforced stop to inherit `--force` from a previous invocation. This silently bypasses the running-execution query and confirmation.

<details><summary>Evidence</summary>

> Step 1: An embedding can call `cmd := NewStopCommand()`, execute it with `cmd.SetArgs([]string{"node-a", "--force"})`, then call `cmd.SetArgs([]string{"node-b"})` and execute the same command again (or create two `NewRootCommand`/`NewStopCommand` values before executing the first).
> Step 2: `NewStopCommand` binds every command's `force` flag to the single package variable via `cmd.Flags().BoolVar(&stopForce, "force", false, ...)`.
> Step 3: The first parse sets `stopForce` to `true`; pflag's `Parse` only calls `Set` for flags present in the new argument list, so the second invocation's absent `--force` leaves that value true. With two preconstructed command trees, the second tree's flag also points at the same variable.
> Step 4: `runStopCommand` copies `stopForce` into `AgentNodeStopper.Force`, and `confirmStopWithRunningExecutions` immediately returns when `Force` is true. Therefore the unforced second stop skips both the control-plane query and its safety warning/prompt.

</details>

control-plane/internal/core/services/agent_service.go
```diff
@@ -126,37 +170,95 @@ func (as *DefaultAgentService) runAgentGuarded(name string, options domain.RunOp
...
 
-	// Convert to domain model and return
-	runningAgent := as.convertToRunningAgent(agentNode)
-	runningAgent.PID = pid
-	runningAgent.Port = port
-	runningAgent.StartedAt = time.Now()
+// freshRetryPort excludes the port that just failed to bind, then asks the port
+// manager for a new one, so the retry never reuses the conflicting port.
+func (as *DefaultAgentService) freshRetryPort(failedPort int) (int, error) {
+	_ = as.portManager.ReservePort(failedPort)
```
AbirAbbas, line 211: > [!NOTE]
> **Advisory — non-blocking.** Safe to merge and address in a follow-up.
>
> _Why non-blocking:_ The finding describes a race condition that may prevent retry on a different port, but does not demonstrate a concrete production impact such as data loss, security vulnerability, or broken API contract.

**`freshRetryPort` doesn’t exclude the failed port when reservation fails, so the retry can pick the same port and bail out**

<sub>🔵 `SUGGESTION` · confidence 94%</sub>

`freshRetryPort` calls `ReservePort(failedPort)` but ignores the error and immediately scans from 8001 via `FindFreePort`. If the port was system-busy at reservation time but frees up before the scan, `FindFreePort` returns it—there’s no reservation and no listener. `startWithPortRetry` then sees `retryPort == port` and returns the original error without trying any other port, even when one is available.

The startup tests only cover the case where `ReservePort` succeeds; they inject the same return value but never exercise the production path where reservation fails and the port later becomes free during `FindFreePort`.

<details><summary>Evidence</summary>

> Step 1: `RunAgent` calls `runAgentGuarded`, which invokes `startWithPortRetry(port, ...)` at `agent_service.go:112` after `attemptStart` reports a strict-port conflict. Step 2: `startWithPortRetry` calls `freshRetryPort(failedPort)` at line 195. Step 3: `freshRetryPort` ignores `ReservePort(failedPort)` at line 209 and immediately calls `FindFreePort(8001)`. Step 4: the `PortManager` contract permits `ReservePort` to return an error when a port is in use, and `DefaultPortManager.ReservePort` does exactly that before adding nothing to `reservedPorts` (`port_manager.go:69-76`). Step 5: if that system-busy port becomes available before `FindFreePort` scans it, `FindFreePort` sees neither a reservation nor a busy listener and returns it (`port_manager.go:34-36`). Step 6: `startWithPortRetry` detects `retryPort == port` and returns the original error at lines 196-199, without searching for any distinct available port. The only production constructors use `process.NewPortManager`, and no other production caller invokes `freshRetryPort`, so this concrete behavior applies to the runtime path.

</details>

**💡 Suggested Fix**

Make exclusion independent of a successful reservation—for example, add an exclusion-aware port-manager operation that skips `failedPort` while scanning, and use it here. Add a startup test where `ReservePort(failedPort)` returns the concrete system-busy error and the subsequent scan would otherwise return `failedPort`, while a distinct port is available.

---
*`Runtime contract of strict-port retry` · confidence 94%*

control-plane/internal/cli/list.go
```diff
@@ -84,6 +95,135 @@ func runListCommandJSON() error {
...
+	client := &http.Client{Timeout: 3 * time.Second}
+	req, err := http.NewRequest(http.MethodGet, serverURL+"/api/v1/nodes?show_all=true", nil)
+	if err != nil {
+		return nil, false
+	}
+	if key := GetAPIKey(); key != "" {
+		req.Header.Set("Authorization", "Bearer "+key)
+	}
+
+	resp, err := client.Do(req)
```
AbirAbbas, line 204: > [!NOTE]
> **Advisory — non-blocking.** Safe to merge and address in a follow-up.
>
> _Why non-blocking:_ The finding describes a UX/observability improvement for automation consumers, not a security vulnerability, data loss, API break, or regression.

**Control-plane authentication and response failures are mislabeled as unreachable**

<sub>🔵 `SUGGESTION` · confidence 99%</sub>

Return a typed outcome from `fetchControlPlaneNodes` so `af list` exposes authentication and response failures separately from unreachable errors.

Currently, `resolveNodeHealth` receives only a boolean reachability flag, which bundles 401/403, 5xx responses, malformed JSON, and empty/invalid URLs into `unknown (control plane unreachable)`. This prevents automation from distinguishing configuration errors (credentials, URL) from a transient network outage, despite the control plane having responded in those cases.

<details><summary>Evidence</summary>

> Step 1: both `runListCommandJSON` (line 72) and `runListCommand` (line 261) call `resolveNodeHealth`. Step 2: `resolveNodeHealth` receives only the boolean returned by `fetchControlPlaneNodes` (line 118) and passes it to `reconcileHealth` (line 128). Step 3: `fetchControlPlaneNodes` returns `reachable=false` for a non-200 HTTP response (lines 204-211) and JSON decoding failure (lines 213-217), exactly as it does for transport errors (lines 204-206) and an empty URL (lines 190-193). Step 4: `reconcileHealth` maps every `cpReachable=false` case to the literal `unknown (control plane unreachable)` (lines 138-141), which is serialized as JSON `health` at lines 78-85 without any separate outcome field.

</details>

**💡 Suggested Fix**

Return a typed reconciliation/fetch outcome (for example `unreachable`, `unauthorized`, `http_error`, and `invalid_response`) in addition to the nodes. Keep `af list` non-failing and health unknown, but render/serialize a distinct reason (ideally a stable JSON status/reason field) so consumers can act on authentication and response failures correctly. Add 401/403, 5xx, and malformed-body cases to `list_health_test.go`.

---
*`Machine-consumer semantics` · confidence 99%*

control-plane/internal/core/services/agent_service.go
```diff
@@ -126,37 +170,95 @@ func (as *DefaultAgentService) runAgentGuarded(name string, options domain.RunOp
...
+}
 
-	// Convert to domain model and return
-	runningAgent := as.convertToRunningAgent(agentNode)
-	runningAgent.PID = pid
-	runningAgent.Port = port
-	runningAgent.StartedAt = time.Now()
+// freshRetryPort excludes the port that just failed to bind, then asks the port
+// manager for a new one, so the retry never reuses the conflicting port.
+func (as *DefaultAgentService) freshRetryPort(failedPort int) (int, error) {
```
AbirAbbas, line 210: > [!NOTE]
> **Advisory — non-blocking.** Safe to merge and address in a follow-up.
>
> _Why non-blocking:_ The leaked reservation only affects in-memory state for the lifetime of a single control-plane process and does not cause data loss, security issues, or break a public contract.

**Failed-port reservation leaks into future runs**

<sub>⚪ `NITPICK` · confidence 82%</sub>

Release the reserved failed port after `freshRetryPort` selects a retry port.

`freshRetryPort` reserves the failed port at `agent_service.go:209` but never calls `ReleasePort`. A successful `ReservePort` marks the port in the in-memory map (`port_manager.go:67-78`), which `FindFreePort` rejects (`port_manager.go:32-37`). This makes the port permanently unavailable for future allocations until the process restarts, even after it becomes system-free. Release the reservation (including on error paths) once `FindFreePort` returns, and add a test that a later independent run can allocate the previously failed port.

<details><summary>Evidence</summary>

> Step 1: A strict-port failure reaches `startWithPortRetry`, which calls `freshRetryPort` (`agent_service.go:195`).
> Step 2: `freshRetryPort` calls `ReservePort(failedPort)` at `agent_service.go:209` and has no corresponding `ReleasePort` call.
> Step 3: Under the real implementation, a successful `ReservePort` sets `pm.reservedPorts[port] = true` (`internal/infrastructure/process/port_manager.go:67-78`).
> Step 4: Every future `FindFreePort` rejects any port present in that map (`port_manager.go:32-37`).
> Step 5: Thus, if the failed port becomes system-free, subsequent `af run` calls still cannot allocate it until process restart, despite no node owning it.

</details>

---
*`Startup retry behavioral contract` · confidence 82%*

PR #822, opened 2026-07-23
feat: user-built subharness agents — personal-agent skill path, installable scaffolds, skill routing + cleanup

AbirAbbas:
## What this does

Lets users build **personal agents** — agents that live on their machine, installed through `af`, visible in AgentField Desktop — and gives coding agents a clean path to discover and delegate to them. Three pieces:

### 1. New `agentfield-personal` skill (0.1.0)

A standalone skill for the machine-installed deliverable: author real source in `~/agentfield-agents/<name>`, package it with a `config_version: v1` manifest, declare scoped secrets (`global` vs `node`) that surface as a keys form in Desktop, `af install` + `af run`, then verify actual control-plane registration and a live reasoner call before claiming success. Secrets are only ever set through the `af` CLI — never echoed, committed, or written into the manifest.

Skill selection happens at routing level via the skill description. The existing `agentfield` (repo builder) skill is **unchanged in behavior**: its body is byte-identical to main, with only a frontmatter version and a one-sentence description pointer to the new skill. An earlier revision of this PR added an in-skill coverage/routing gate to the builder; that was dropped as a regression for existing users in favor of this split.

`agentfield-use` gains a bounded fallback: after discovery conclusively finds no capable installed agent, it may *offer* to build one (naming both builder skills) — list/inspect/diagnose requests never authorize a build.

### 2. `af init` now emits `agentfield-package.yaml`

All three language templates (Python, Go, TypeScript) scaffold a v1 manifest, so a generated project is installable with `af install ./dir` out of the box. The TypeScript path also gains real dependency installation (`npm install`) in the package installer.

### 3. Skillkit alias-orphan reconciliation

`agentfield-multi-reasoner-builder` was folded into `agentfield` as an alias earlier; installs that predate that kept a stale standalone entry. `af skill install`/`update` now reconciles those orphans: removes recorded targets (symlinks / marker blocks, name-matched), the canonical store dir, and the state entry. It only touches state entries whose name is a catalog alias, so it cannot affect user skills.

## Testing

- Contract tests pin the split: the builder skill must keep its original workflow (routing-gate text is explicitly banned), the personal skill must carry the full lifecycle/secret-safety contract, and the use-skill offer must stay authorization-bounded.
- Mirror tests keep `skills/` and the embedded `skill_data/` copies aligned (name/version/content).
- Reconciliation has ~700 lines of table tests over recorded-state permutations, including failure injection via fs seams.
- Rendered-manifest and scaffold tests verify every emitted `agentfield-package.yaml` parses with `packages.ParsePackageMetadata`.
- Full control-plane suite green locally and in CI.

---


habibgurdov77-star:
53b6b757985a2a209d9c63e6a4eb9fb18921ab7b

habibgurdov77-star:
https://github.com/Agent-Field/SWE-AF/blob/main/docs%2FCONTRIBUTING.md

PR #843, opened 2026-07-30
fix(control-plane): package sync reconciliation + install_status on the packages listing

AbirAbbas:
## Summary

Found via live end-to-end testing of the desktop convergence branch (#842): `GET /api/ui/v1/agents/packages` cannot tell an installed package from a pre-seeded catalog entry, for two stacked reasons:

1. **`SyncPackagesFromRegistry` was insert-only.** A package whose row already existed (catalog seeding) never got upgraded to `installed` when actually installed, and an uninstalled package kept `status: installed` in the DB forever — the sync never downgraded or pruned anything.
2. **The wire format had no install-state field at all** — `PackageInfo.status` is a derived configuration summary (`not_configured`/`configured`), and the raw registry-backed state wasn't exposed.

Concrete symptom: a client treating the listing as "installed agents" shows the entire marketplace catalog as installed, and uninstalled packages never disappear.

## Changes

- **`package_sync.go`** — the sync is now a true reconcile: registry entries are upserted (catalog rows upgrade to `installed`, preserving `installed_at`/configuration status), and rows claiming installed/running/stopped that are absent from `installed.yaml` are downgraded to `uninstalled`. An absent registry file is treated as an empty registry (nothing installed) rather than skipping reconciliation. The fsnotify watcher already re-runs this on every registry change, so API uninstalls now propagate to the DB live.
- **`packages.go`** — additive wire fields: `install_status` (raw `installed|running|stopped|uninstalled|...`) and `installed_at` (RFC3339, omitted when never installed). The existing derived `status` field is untouched, so current web-UI consumers are unaffected.

## Validation contract → tests

Upgrade of a pre-seeded catalog row (status/schema/installed_at), downgrade of removed packages, catalog rows untouched by the downgrade pass, absent-registry-file downgrade, plus the pre-existing insert/skip tests unchanged.

## Test plan

- [x] `go build ./...`, `go test -tags sqlite_fts5 ./internal/server/ ./internal/handlers/ui/`, lint clean on changed files
- [x] Live-verified against a real server + real install/uninstall cycle (the failing desktop E2E scenario now has correct server-side state)

---

### DIFF
--- a/control-plane/internal/handlers/ui/packages.go
+++ b/control-plane/internal/handlers/ui/packages.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"net/http"
 	"strings"
+	"time"
 
 	"github.com/Agent-Field/agentfield/control-plane/internal/storage"
 	"github.com/Agent-Field/agentfield/control-plane/pkg/types"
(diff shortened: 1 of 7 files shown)


AbirAbbas:
Added a third commit (211827c5): the job manager now triggers the registry→DB sync synchronously after successful API-driven install/update/uninstall, so API clients read their own writes instead of racing the fsnotify watcher. Found by the same live E2E — listing packages immediately after an API uninstall still showed the package until the async watcher fired.

PR #845, opened 2026-07-30
feat(desktop): cloud mode — connect the desktop to a remote control plane

AbirAbbas:
## Summary

The final piece of the one-management-path arc (#837 → #841 → #842 → #843 → #844): AgentField Desktop can now connect to a **remote control plane** with a saved `{server_url, api_key}` profile, managing agents on that box exactly the way it manages a local one — same client, same flows, only the base URL changes.

## What's new

**Cloud tab** (`agentfield://cloud`, ⌘/Ctrl+5): connection status, server URL + API key form, live **Test connection** verdict (reachable / auth / install API / server version), **Save & switch**, and **Switch back to local** (keeps the saved profile for one-click re-enable). A **Deploy on Railway** section opens the control-plane template with paste-back instructions, via a dedicated IPC channel — no generic URL-opener hole.

**Connection core**: a new cycle-free connection-state module supplies base URL + API key to `cpClient` *and* the five raw dashboard/tray fetches that previously bypassed auth entirely. URL normalization refuses plaintext `http://` to public hosts, so keys can't leak over cleartext. The cloud profile is applied at startup and immediately on settings change, and survives restarts (`normalizeSettings` extended — it silently drops unknown keys otherwise).

**Guardrails in cloud mode**: no local port probing, no `af server` spawn, no agent autostart (one remote health check instead), local-server-start IPC refuses politely, skills sync skipped. Switching back restores previous local behavior exactly.

Also fixes a latent bug: `fetchUsageStats` ignored the active base URL.

## Validation

- **Live E2E (5/5)** against a real CP running with `AGENTFIELD_API_KEY` enforcement: full management lifecycle (install → secrets → start → stop → uninstall) driven through the cloud profile; test-connection matrix verified live (valid key / rejected key / unreachable); autostart gating and local-mode restore confirmed. Notable: with a key configured, the CP properly 401s missing/wrong keys — fail-closed as required.
- 339 desktop tests green (up from 323), `typecheck` clean, `dist:dir` builds.
- API key is stored in the desktop settings JSON for now (same trust level as existing `~/.agentfield` config material); moving it to Electron `safeStorage` is a noted follow-up.

## Follow-ups (not in this PR)

- Railway template v2 (point at the `control-plane-cloud` image + volume + generated key default)
- Automated Railway deploy via loopback+PKCE OAuth (needs an OAuth client registration)
- Electron `safeStorage` for the API key at rest

---

### DIFF
--- a/desktop/src/main/agentfield.test.ts
+++ b/desktop/src/main/agentfield.test.ts
@@ -290,6 +290,11 @@ describe('fetchControlPlaneNodes', () => {
     )
   })
 
+  it('treats a null nodes slice as an empty control-plane view', async () => {
+    const fetchImpl: FetchLike = async () => jsonResponse({ nodes: null, count: 0 })
+    expect(await fetchControlPlaneNodes('http://localhost:8080', fetchImpl)).toEqual(new Map())
+  })
+
   it('returns null on a non-200 response', async () => {
     const fetchImpl: FetchLike = async () => jsonResponse({ error: 'nope' }, 500)
     expect(await fetchControlPlaneNodes('http://localhost:8080', fetchImpl)).toBeNull()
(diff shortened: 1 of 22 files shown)


AbirAbbas:
Folded into #847, which carries these commits plus the one-click Railway deploy — one PR with the full cloud-mode context.

PR #851, opened 2026-08-01
fix(control-plane): reconcile must honor granted node status leases

AbirAbbas:
## Bug

Idle Go-SDK nodes deterministically flap offline ~30s out of every 2-minute lease cycle. Measured on a live cloud deploy with a 10s poll: the node reads active for ~90s, disappears from `/api/v1/nodes` for ~30s, then returns on the next lease renewal — repeating indefinitely. Reproduced locally with an idle agent: **33/42 polls active** over 7 minutes on current main.

## Root cause

Three constants that never met each other:

- `NodeStatusLeaseHandler` grants nodes a **5-minute** lease (`DefaultLeaseTTL`) and tells them so (`lease_seconds`, `next_lease_renewal`).
- The Go SDK renews every **2 minutes** — comfortably inside its granted lease.
- The status manager's reconcile sweep marks any active node inactive once its heartbeat is older than `HeartbeatStaleThreshold` (**60s** default).

The server promises "you're good for 300s" and then declares the node dead at 60s. Busy nodes are masked (execution traffic keeps status fresh); idle nodes flap every cycle. This is the remaining half of the lease-liveness work from #847 (which fixed the frozen `last_seen` snapshot).

## Fix

`StatusManager` now records each granted lease expiry (`RecordLease`, called from both lease-granting paths in the handler), and the heartbeat-staleness reconcile rule skips nodes whose recorded lease (+30s grace) is still running. Entries are overwritten on renewal and lazily dropped after expiry.

Deliberately unchanged:
- **Dead-node detection stays fast.** The HTTP health monitor's consecutive-failure path is untouched; its heartbeat gate stops protecting a node once the heartbeat goes stale, so a dead lease-holder is still marked inactive after N failed probes — it does not get to hide behind an unexpired lease.
- **Python-SDK heartbeaters are unaffected.** Nodes that never receive a lease grant have no registry entry; the 60s staleness rule applies to them exactly as before.
- All other reconcile rules (stuck-starting, unknown-with-fresh-heartbeat, inconsistent status).

## Verification

- Idle-agent A/B, same harness (local CP + Go-SDK agent, 10s poll, ≥3 lease cycles): main **33/42** active → this branch **42/42**.
- New unit tests derive from the contract: unexpired lease suppresses staleness reconcile; expired lease (+grace) restores it; no-lease nodes unchanged; renewal extends the horizon; handler records leases on both the normal and `pending_approval` paths.
- `gofmt` clean on touched files; `go build ./...` clean; `go test ./...`: `internal/services` and `internal/handlers` (home of all changed code) pass. `internal/server`'s `TestStartAndStopCoverAdditionalBranches` fails identically on unmodified main in my environment (a local process holds the admin gRPC port) — pre-existing and environmental, expecting CI to confirm.

### DIFF
--- a/control-plane/internal/handlers/coverage_additional_test.go
+++ b/control-plane/internal/handlers/coverage_additional_test.go
@@ -159,6 +159,7 @@ func (s *cleanupStorageStub) CleanupWorkflow(ctx context.Context, workflowID str
 
 type nodeRESTStorageStub struct {
 	storage.StorageProvider
+	mu               sync.Mutex
 	agent            *types.AgentNode
 	versionedAgent   *types.AgentNode
 	listAgents       []*types.AgentNode
@@ -170,6 +171,7 @@ type nodeRESTStorageStub struct {
 	lastVersion      string
 	updatedLifecycle *types.AgentLifecycleStatus
 	registeredAgent  *types.AgentNode
+	healthUpdates    []types.HealthStatus
 }
 
 func (s *nodeRESTStorageStub) GetAgent(ctx context.Context, id string) (*types.AgentNode, error) {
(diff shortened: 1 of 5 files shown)


santoshkumarradha (approved):
Read the lease-vs-heartbeat fix closely and spot-checked it locally in a clean worktree with `go test ./internal/services ./internal/handlers`. The lease recording and reconciliation coverage look good to me.

PR #852, opened 2026-08-01
Add MiniMax text-to-speech support

octo-patch:
Reason: Add native MiniMax text-to-speech support through the existing media provider route.

- Send regional `/t2a_v2` requests with the current default model, speech options, and hex or URL response handling.
- Route `minimax/` audio models to the media provider while preserving provider validation errors.
- Add focused coverage for request construction, regional endpoints, response parsing, validation, and high-level routing.

Checks:
- `uvx --from ruff==0.15.22 ruff check .`
- `uvx --from ruff==0.15.22 ruff format --check agentfield/media_providers.py agentfield/agent_ai.py tests/test_minimax_audio.py tests/test_minimax_video.py`
- `uv run --frozen --extra dev ./scripts/run_pytest.sh tests/test_minimax_audio.py tests/test_minimax_video.py tests/test_media_providers.py tests/test_agent_ai.py --no-cov`

### DIFF
--- a/sdk/python/agentfield/agent_ai.py
+++ b/sdk/python/agentfield/agent_ai.py
@@ -1443,16 +1443,21 @@ class AgentAI:
         """
         AI method optimized for audio output generation.
 
-        Automatically detects the model type and uses the appropriate LiteLLM function:
+        Automatically detects the model type and uses the appropriate provider:
+        - For MiniMax TTS models (minimax/speech-2.8-hd, etc.): Uses MiniMax T2A
         - For TTS models (tts-1, tts-1-hd, gpt-4o-mini-tts): Uses litellm.speech()
         - For audio-capable chat models (gpt-4o-audio-preview): Uses litellm.completion() with audio modalities
 
         Args:
             *args: Input arguments (text prompts, etc.)
-            voice: Voice to use for audio generation (alloy, echo, fable, onyx, nova, shimmer)
-            format: Audio format (wav, mp3, etc.)
+            voice: Voice identifier. OpenAI voices include alloy, echo, fable,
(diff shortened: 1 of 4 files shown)


AbirAbbas (approved):
Took this over along with your other two MiniMax PRs since they'd gone quiet.

This was the cleanest of the three. I ran TTS end-to-end against a mocked transport and confirmed the `/t2a_v2` request shape for both the global and China endpoints, both the hex and URL response paths, and every validation branch — all behaved correctly. I also checked the `agent_ai.py` routing change specifically, since that touches the public high-level API: `minimax/` audio models reach the media provider and nothing else changed route.

On the `test_minimax_video.py` edit — I checked it isn't weakening anything. It only removes the now-obsolete assertion that `generate_audio()` raises `NotImplementedError`, which this PR makes untrue. Fine as-is.

Two small things I pushed a commit for:

The `ai_generate_audio` docstring still listed only LiteLLM and Fal.ai and documented only OpenAI voices, so there was no way for a caller to discover MiniMax support or realize the `alloy` default doesn't apply to it. Updated that and `ai_with_audio` with a `minimax/speech-2.8-hd` example.

The validation test only covered four of the branches this PR adds. I verified the uncovered ones behave correctly by hand, but a regression in any of them would have passed the suite — so I parameterized it to cover invalid `output_format`, speed-without-voice, missing `voice_id`, empty model, HTTP and API errors, malformed JSON, incomplete status, absent audio, and invalid hex. No production change was needed for that.

Heads up that #853 conflicts with this one in `supported_modalities` and the shared video test. This one goes first; I'll reconcile #853 on top.

PR #853, opened 2026-08-01
Add MiniMax image generation support

octo-patch:
Reason: Add native MiniMax image generation support to the existing media provider pipeline.

Changes:
- Implement image generation requests with global and China regional endpoint support.
- Support URL and base64 image responses, optional request fields, and model-prefix routing.
- Add focused tests for request construction, response parsing, validation, and provider registration.

Checks:
- `git diff --check origin/main...HEAD`

### DIFF
--- a/sdk/python/agentfield/media_providers.py
+++ b/sdk/python/agentfield/media_providers.py
@@ -745,7 +745,7 @@ class MiniMaxProvider(MediaProvider):
 
     @property
     def supported_modalities(self) -> List[str]:
-        return ["video", "music", "audio"]
+        return ["video", "music", "audio", "image"]
 
     @property
     def video_model_metadata(self) -> Dict[str, Dict[str, Any]]:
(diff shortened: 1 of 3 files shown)


AbirAbbas (approved):
Last of the three MiniMax PRs — took these over since they'd been quiet for a few days.

I found one real bug here. Response parsing read `data.image_urls` unconditionally, but MiniMax documents base64 responses under `data.image_base64`. So `response_format="base64"` — one of the two modes this PR advertises — would have failed against the real API every time with "returned no image_urls". The reason it looked fine is that the base64 test built its fixture with an `image_urls` payload, so it asserted the implementation's assumption rather than the provider's actual contract, and passed while masking the defect.

I've pushed a fix that selects the response key from the requested format and names the actual mode in the error, plus corrected the base64 fixture to the documented `image_base64` shape. I confirmed it fails before the change and passes after.

Two smaller things in the same commit: `base_resp` and `data` were only type-checked at the top level, so a non-dict nested value leaked `AttributeError` instead of the normalized `RuntimeError` (#852 already did this properly, so I matched its idiom), and I added regression coverage for the branches that had none — empty model, HTTP status errors, non-dict JSON, missing and empty image collections.

On the `test_minimax_video.py` edit: I checked it isn't weakening video coverage. It only drops the obsolete `generate_image` `NotImplementedError` expectation, which this PR makes untrue.

Since this landed last of the three, I rebased it and reconciled the overlap with #852 and #854 by hand: `supported_modalities` is now the union `['video', 'music', 'audio', 'image']`, the H3 constants and `video_model_metadata` from #854 are preserved, and both stale `NotImplementedError` assertions are gone. Full Python SDK suite is green on the reconciled result.

PR #854, opened 2026-08-01
Add MiniMax H3 video v2 support

octo-patch:
Reason: Refresh MiniMax video support for the H3 v2 API and current pricing metadata.

- add v2 video task create, query, list, and delete operations
- validate H3 multimodal content, roles, 2K resolution, duration, ratio, and regional fields
- route H3 generation through v2 while preserving existing v1 video and music behavior
- expose current H3 model and regional pricing metadata

Checks:
- `uvx --from ruff==0.15.22 ruff check sdk/python/agentfield/media_providers.py sdk/python/tests/test_minimax_video.py`
- `uv run --frozen --extra dev python -m pytest tests/test_minimax_video.py tests/test_media_providers.py tests/test_media_integration.py -q --no-cov`
- `python -m py_compile sdk/python/agentfield/media_providers.py sdk/python/tests/test_minimax_video.py`
- `git diff --check`

### DIFF
--- a/sdk/python/agentfield/media_providers.py
+++ b/sdk/python/agentfield/media_providers.py
@@ -44,29 +44,35 @@ MINIMAX_H3_VIDEO_METADATA: Dict[str, Any] = {
     ],
     "input_modalities": ["text", "image", "video", "audio"],
     "output_modalities": ["video", "audio"],
-    "resolutions": ["2K"],
+    "resolutions": ["768P", "2K"],
     "duration_seconds": {"min": 4, "max": 15, "integer": True},
     "pricing": {
         "global_en": {
             "currency": "USD",
-            "output_video": {"unit": "second", "rates": {"2K": 0.13}},
+            "output_video": {
+                "unit": "second",
+                "rates": {"768P": 0.08, "2K": 0.13},
+            },
             "input_reference_video": {
                 "unit": "second",
-                "rates": {"2K": 0.02},
+                "rates": {"768P": 0.08, "2K": 0.13},
             },
             "input_reference_audio": {"free": True},
             "input_reference_images": {
                 "unit": "image",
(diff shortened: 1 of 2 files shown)


AbirAbbas (approved):
Picked this up since it had been quiet for a few days. I tested it rather than just reading the diff — ran the V2 lifecycle against a local mock server, and replayed the legacy V1 video and music calls against both this branch and main to diff the outgoing requests. The preservation claim holds: legacy request construction is byte-identical to main.

Two things needed fixing before this could land, and I've pushed a commit for both:

**Pricing was wrong in a way that matters.** Global rates were `$0.02/sec` for reference video and `$0.03` per additional image; MiniMax publishes `$0.13/sec` and `$0.04`. Because `cost_usd` flows into AgentField's usage tracking, that under-reports real spend rather than just being a cosmetic metadata error. The fixture asserted `$0.72` where the correct math gives `$0.95`. I corrected the constants for both regions and made `_estimate_h3_cost_usd` pick its rate from the task's actual resolution instead of assuming 2K.

**768P was rejected.** The official V2 schema's resolution enum is `["768P", "2K"]`, but `_build_h3_request` hard-required 2K, so a valid 768P request failed locally. Now both are accepted and both are advertised in the metadata, with 768P pricing filled in for both regions.

I also dropped `expired` from the status enum (official statuses end at `cancelled`), added validation for `task_type` — it was previously unvalidated, so `filter.task_type=bogus` sailed through to the server — and added a regression test locking the legacy music request shape, since that preservation guarantee was only proven by my throwaway probe and needed to be protected in the suite while the three MiniMax PRs get reconciled.

Worth flagging for the other two: #852 and #853 conflict with each other in `supported_modalities` and the shared video test. This one is textually clean against both, so it goes first.

PR #861, opened 2026-08-02
feat(harness): add Grok Build CLI provider

storyicon:
## Summary

This PR adds a first-class **Grok Build CLI** harness provider so AgentField can orchestrate local `grok` headless sessions with the same provider surface as Codex, OpenCode, Gemini, and Claude Code.

The goal is to make Grok a peer runtime for multi-turn coding-agent work (read/write/edit/bash tool loops) without requiring callers to shell out ad hoc or invent a parallel execution path.

## Motivation

AgentField already standardizes coding-agent execution behind `HarnessProvider` + `build_provider()`. Operators who use Grok Build CLI locally currently cannot plug that binary into the harness stack:

- No `provider="grok"` registration
- No binary resolution / doctor metadata
- No metrics parsing that tolerates Grok’s incomplete usage payloads

Without this, multi-agent frameworks built on AgentField (for example SWE-AF) cannot select Grok as a first-class runtime the way they select Codex or Claude Code.

### 1. Provider shape matches existing CLI providers

`GrokProvider` follows the same contract as `CodexProvider` / `OpenCodeProvider`:

| Concern | Approach |
| --- | --- |
| Discovery | `ensure_cli_available("grok", bin_path)` |
| Invocation | async `run_cli` with optional env / cwd |
| Errors | `RawResult(is_error=True, failure_type=…)` |
| Success payload | final agent text in `RawResult.result` |

This keeps the runner, schema pipeline, and retry logic provider-agnostic.

### 2. PTY wrapping is required for Grok today

Plain pipe attachment causes Grok to fail with `Device not configured (os error 6)`. When `script(1)` is available, the provider runs:

```text
script -q /dev/null grok … --output-format json --prompt-file …
```

`script` allocates a PTY while still delivering agent JSON on stdout. If `script` is missing, the provider falls back to a direct `grok` invocation (same flags) so environments without `script` are not hard-broken, but PTY-capable hosts remain the recommended path.

### 3. Prompt delivery via `--prompt-file`

Long AgentField prompts exceed practical argv limits. The provider always writes the prompt to a temporary file and passes `--prompt-file`, then unlinks the file in `finally`.

### 4. Permission mode mapping for unattended runs

Harness executions are unattended. The provider maps common AgentField permission modes onto Grok’s `--permission-mode` values and defaults to non-interactive approvals (`bypassPermissions` / `acceptEdits` / `--always-approve`) so coding loops do not block on TTY prompts.

### 5. Metrics robustness (token fields may be absent)

Grok’s JSON envelope frequently omits `usage.input_tokens` / `output_tokens`. The previous pattern of storing `None` into `Metrics` integer fields breaks aggregation in the shared runner:

```text
tokens["input_tokens"] += raw.metrics.input_tokens  # int += None → TypeError
```

This PR:

- Normalizes missing token counts to **`0`** inside `_usage_metrics`
- Also coerces `None → 0` in `_accumulate_metrics` as a defensive invariant for all providers

This is intentional fail-soft for **accounting only**. It does not invent business outcomes or hide failed CLI runs.

### 6. Localhost / proxy hygiene

The provider sets default `NO_PROXY` / `no_proxy` for `127.0.0.1,localhost,::1` so control-plane traffic is not accidentally routed through local HTTP proxies during harness sessions.

### 7. What this PR deliberately does **not** do

- Does **not** implement Grok `--json-schema` constrained decoding (can be a follow-up)
- Does **not** change the universal Write-to-`.agentfield_output.json` schema strategy in the runner
- Does **not** add network/API-based Grok chat completions (this is CLI-only, consistent with Codex/OpenCode)

## API / config surface

- `SUPPORTED_PROVIDERS` includes `"grok"`
- `HarnessConfig.provider` documents `"grok"`
- `HarnessConfig.grok_bin` (default `"grok"`)
- Doctor / availability metadata for binary install + optional `XAI_API_KEY`

Example:

```python
from agentfield.types import HarnessConfig
from agentfield.harness.providers._factory import build_provider

provider = build_provider(HarnessConfig(provider="grok", model="grok-4.5"))
```

## Files

| Path | Change |
| --- | --- |
| `sdk/python/agentfield/harness/providers/grok.py` | New Grok CLI provider |
| `sdk/python/agentfield/harness/providers/_factory.py` | Register `grok` |
| `sdk/python/agentfield/harness/_availability.py` | ProviderSpec for `grok` |
| `sdk/python/agentfield/harness/_runner.py` | Safe metrics accumulation |
| `sdk/python/agentfield/types.py` | `grok_bin` + provider docs |
| `sdk/python/tests/test_harness_factory.py` | Factory + protocol coverage |
| `sdk/python/tests/test_harness_grok_provider.py` | Unit tests for provider helpers |

## Test plan

- [x] `build_provider(HarnessConfig(provider="grok"))` returns `GrokProvider`
- [x] Custom `grok_bin` is honored
- [x] Missing usage tokens become `0` (not `None`)
- [x] JSON payload extraction tolerates `script` control-character noise
- [x] Factory supported-provider list includes `grok`
- [ ] CI: `sdk/python` harness unit tests on this branch
- [ ] Manual (optional): with local `grok` installed and logged in, run a short harness task with `provider="grok"` and confirm non-error `RawResult`

## Compatibility

- **Additive** for existing providers (Claude / Codex / Gemini / OpenCode paths unchanged aside from defensive metrics coercion)
- Requires a local Grok Build CLI on `PATH` (or `grok_bin`) for live runs
- Consumers that pin older AgentField versions are unaffected until they upgrade

## Follow-ups (out of scope)

1. Optional native `--json-schema` structured decoding for Grok
2. Streaming / idle-watchdog tuning for long non-streaming JSON sessions
3. Version-pin guidance once Grok CLI flags stabilize across releases


sdk/python/agentfield/harness/providers/grok.py
```diff
@@ -0,0 +1,302 @@
...
+            if variant_value:
+                grok_cmd.extend(["--reasoning-effort", variant_value])
+
+            system_prompt = options.get("system_prompt")
+            if isinstance(system_prompt, str) and system_prompt.strip():
+                grok_cmd.extend(["--system-prompt-override", system_prompt])
+
+            if shutil.which("script") is not None:
+                # -q quiet, /dev/null typescript file; stdout still carries agent JSON.
+                cmd = ["script", "-q", "/dev/null", *grok_cmd]
```
santoshkumarradha: This `script` invocation works on BSD/macOS, but util-linux `script` on Linux does not accept `script -q /dev/null grok ...` as the command form. Its supported forms are `script [options] [file] [-- command ...]` or `-c`, so on our Linux CI runners this ends up invoking the default shell instead of Grok. Since this provider is specifically trying to fix the non-TTY startup path, we need to build the wrapper in a Linux-safe form before merging, for example `script -q /dev/null -- grok ...` or `script -q -c '...' /dev/null` with correct argv escaping.

santoshkumarradha (dismissed):
Thanks for pushing this through. The provider shape looks reasonable overall, but I’m not comfortable merging it yet because the PTY wrapper is not portable to the Linux runners we support. I left the concrete spot inline.

sdk/python/agentfield/harness/providers/grok.py
AbirAbbas: Fixed in 4acd4c2 (author's been quiet since your review, so I took it over). One finding from testing on util-linux 2.37.2 (the ubuntu-22.04 runner version): the `script -q /dev/null -- grok ...` form doesn't work either — same silent fallback to an interactive shell. The only working Linux form is `-c`, and `-e` turns out to be mandatory too, otherwise script masks grok's exit code and the provider's returncode handling is dead code. The fix dispatches per-platform (`-qec` + shlex.join on Linux, your validated trailing form untouched on macOS, passthrough when script is absent), with tests pinning the argv shape on all three paths plus a shell round-trip test for the quoting. Details in the helper's docstring.

AbirAbbas: Re-verified the whole matrix on util-linux 2.37.2 with a grok-shaped argv before approving, and it's worth recording *why* the old form fails: it's not a shell fallback in the simple case — GNU getopt permutes, so `script` parses the wrapped command's own flags as its *own* options (`script: unrecognized option '--cwd'`, grok never runs). The `-- grok ...` variant hangs on an interactive shell. `-qec` runs the real argv and preserves exit codes (probed rc=7 through; without `-e` it collapses to 0). Injection round-trip through a real shell is byte-exact with the shlex quoting. Resolving this thread on that basis — the Linux-safety concern is addressed.

AbirAbbas (approved):
Approving on the strength of a full local re-verification of the takeover fix at 5586e97: CI-exact Python gates on 3.10 (ruff 0.15.22 clean, full suite 1929 passed via scripts/run_pytest.sh), the 38 grok provider/factory tests, and a live util-linux 2.37.2 probe of all three script(1) forms — pre-fix form never runs grok (script eats its flags), the `--` variant hangs on an interactive shell, the `-qec` form runs the exact argv and preserves exit codes. Quoting round-trips byte-exact through a real shell. @santoshkumarradha the Linux PTY concern from your review is fixed and thread-resolved with evidence — needs your re-review or a dismissal of the stale changes-requested to unblock the queue.

PR #863, opened 2026-08-02
feat: add a Parallel search option to the deep research example

georgeatparallel:
## Summary

The Deep Research example currently requires a Tavily API key for web search. This adds an explicit `SEARCH_PROVIDER=parallel` option that uses Parallel's hosted Search MCP endpoint without changing the default Tavily path.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs only
- [ ] Tests only
- [ ] CI / tooling
- [ ] Breaking change

## Test plan

- [x] `PYTHONDONTWRITEBYTECODE=1 python3 -m unittest discover -s examples/python_agent_nodes/deep_research/tests -p 'test_*.py' -v`
- [x] `git diff --check` and `python3 -m compileall -q examples/python_agent_nodes/deep_research`
- [x] Ruff lint and format checks for the changed Python files
- [x] Bounded anonymous request to `https://search.parallel.ai/mcp`, parsed through the new adapter

## Test coverage

  **85%**.

- [x] I did not lower or otherwise change `coverage-baseline.json`.

> the PR summary.

## Checklist

- [x] Commits are signed off and follow conventional-commits style.
- [ ] I have linked any related issues (none for this focused example change).

## Related issues / PRs

None.

The repository-wide CI and coverage checks will run after this draft is opened. The hosted CLA will also require completion through the repository's CLA Assistant check.

Disclosure: I work at Parallel.ai, which operates this MCP endpoint.


AbirAbbas (approved):
Merging this — and sorry it sat for three days. The delay was on us: the required checks never appeared because the workflow runs were parked behind our first-time-contributor approval gate, and the advice you got to rebase onto main wouldn't have helped. You merged main twice chasing that. Apologies.

On the review itself, two concerns came up and I want to record how they resolved, because both landed in your favour.

The first was that the adapter isn't a conforming Streamable HTTP MCP client — no `initialize`/`initialized` handshake, no `Mcp-Session-Id`, and `response.json()` despite an `Accept` that permits `text/event-stream`. I ran your `_execute_parallel_search` unmodified against the live endpoint rather than guessing. It returns real results on a bare `tools/call`, answers `application/json` with a 200, and mints a fresh session id per request — it's stateless, so there's no handshake to perform and no session to carry. And listing both media types in `Accept` is a MUST for every Streamable HTTP client, not a contradiction; `json(content_type=None)` already covers a non-JSON content-type header. Concern withdrawn.

The second was that per-query failures get dropped and synthesis proceeds on an empty result set, so a 503 or a timeout reads as "the web had nothing." That behaviour is real, but it predates you — the filter in `execute_search` is untouched context in your diff and applies identically to the Tavily path. Not yours to fix in a PR scoped to adding a backend.

I also confirmed the default path is genuinely unchanged: `_execute_tavily_search` is the old body moved verbatim, same client calls and same options, and `aiohttp` is already a hard dependency of the SDK so the lazy import costs nothing.

A couple of things I may do as a follow-up, all optional and none of them yours: an env override for the MCP URL so it can be repointed without a patch, threading Parallel's own `session_id` argument across the per-query loop for cross-query coherence, and slicing results to match Tavily's `max_results=5` so switching providers doesn't quietly double prompt size.

One thing worth knowing: these example tests don't run in any CI workflow — Python CI is scoped to `sdk/python` — so they're documentation rather than a gate. Not a blocker, just so the coverage isn't assumed to be load-bearing.

PR #872, opened 2026-08-04
Let a manifest declare itself superseded by another package

AbirAbbas:
## Why

A node author who renames or replaces their own node has no way to carry
existing users across. `af install <their url>` keeps installing the old
package forever, because the manifest at that source is the only thing the
installer looks at — and the update path just replays the recorded source.

This adds a generic mechanism for that. **No node is named anywhere in the
control plane**: the redirect lives in the package's own manifest, so any
author gets it and this repo stays agnostic about what people install.

## What

An optional manifest key:

```yaml
name: my-node
superseded_by: https://github.com/me/my-repo//v2
```

Installing a superseded package installs the successor instead, and replaces
the old one when it is already present. The value is anything `af install`
accepts, including a `//subdir` selector and an `@ref`.

Ordering and safety, all covered by tests:

- The successor is installed **first**; only then is the old package retired,
  so a failed install leaves the existing node exactly as it was.
- The redirect is taken before the force check and before anything is copied,
  so it never half-installs the package it is redirecting away from.
- **Node-scoped secrets move to the successor** before the old package is
  uninstalled, which would otherwise delete that scope outright. A value
  already set on the successor wins. Global secrets are shared and untouched.
- Retiring never fails the install — the successor is already working, so a
  leftover is a cleanup chore, not a failure.
- A chain is bounded at 3 hops, so two manifests pointing at each other fail
  loudly instead of cloning forever.

The user is warned before the swap, naming exactly what will be replaced.

## Also: a subdir bug this depends on

A subdirectory reaches the installer two ways — the `//subdir` selector on the
URL, or the `--path` flag. The install API takes the second route, so
`info.URL` arrives bare and the registry recorded the **repo root** as the
source. The next update then resolved that bare source and installed whatever
manifest lives at the root — a *different package* than the one installed.

For a repo shipping a Python root and a Go port side by side, updating the Go
node silently replaced it with the Python one. Fixed in its own commit.

## Validation

Seven tests, one per behaviour, driven through the real `InstallFromGit`
against the existing fake-git harness rather than against internals: redirect,
replace-existing, secret migration (including that the successor's own value
wins and globals survive), the no-prior-install case, the bounded cycle, and
the `--path` source round-tripping back through `ParseGitURL`.

Verified end to end against a real repo with a root manifest and a `go/`
subdirectory: installing the bare repo URL installs the subdirectory's node,
records `//go` in the source, migrates a node-scoped secret, removes the old
package and its directory, and leaves global secrets intact.

`go build`, `go vet`, `gofmt` and the full `go test ./...` are clean (51
packages, 0 failures). `golangci-lint` reports no new issues in the changed
files.

---

### DIFF
--- a/control-plane/internal/packages/git.go
+++ b/control-plane/internal/packages/git.go
@@ -392,6 +392,24 @@ func (gi *GitInstaller) parsePackageMetadata(packagePath string) (*PackageMetada
 }
 
 // updateRegistryWithGit updates the installation registry with Git source info
+// appendSubdirSelector rewrites "https://host/owner/repo[@ref]" into
+// "https://host/owner/repo//subdir[@ref]" so a recorded source round-trips
+// through ParseGitURL. It is needed whenever the subdirectory arrived by the
+// --path flag (or the install API, which splits the selector off before
+// calling): without it the registry records the REPO ROOT, and the next update
+// installs whatever lives there instead of the package that is installed.
+func appendSubdirSelector(url, subdir string) string {
+	subdir = strings.Trim(strings.TrimSpace(subdir), "/")
+	if subdir == "" {
+		return url
+	}
+	base, ref := url, ""
+	if at := strings.LastIndex(url, "@"); at > strings.LastIndex(url, "/") {
+		base, ref = url[:at], url[at:]
+	}
+	return base + "//" + subdir + ref
+}
+
 func (gi *GitInstaller) updateRegistryWithGit(metadata *PackageMetadata, info *GitPackageInfo, sourcePath, destPath string) error {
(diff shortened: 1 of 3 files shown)


AbirAbbas:
Folded into #864 — this was the answer to the migration question raised there, so it belongs in the same PR rather than a stacked one. All three commits (`d26a02f2` subdir source fix, `a18d8561` the `superseded_by` mechanism, `7c43de96` tests) are on `feat/swe-catalog-single-go-entry` unchanged, rebased onto current main.

Superseded by https://github.com/Agent-Field/agentfield/pull/864

PR #888, opened 2026-08-06
feat(control-plane): expose furrow_public_addr in health

AbirAbbas:
## Summary
The desktop's workspace-sync probe (merged in #885, `furrowAddressFrom` in `desktop/src/main/cloud.ts`) reads a `furrow_public_addr` field from the control plane's health response — but no control-plane version emits that field, so `furrowAvailable` can never be true, even on a deployment where workspace sync is fully provisioned and reachable. Found while e2e-testing #885 + Agent-Field/SWE-AF#130 against a live cloud deployment.

## Changes Made
- The shared health handler (`GET /health` and `GET /api/v1/health`) now includes top-level `furrow_public_addr` when the `FURROW_PUBLIC_ADDR` env var is non-empty — the var the deploy engine already sets on provisioned services. When unset, the key is absent (not empty), matching what the desktop probe expects.
- Handler test covering both cases (`TestHealthCheckHandlerFurrowPublicAddr`).

## Test Plan
- [x] `go test ./internal/server` (new subtests `set` / `empty` pass)
- [x] `go build ./...`, `go vet ./...`, gofmt clean
- [x] Desktop-side consumer verified separately against a live cloud deployment (probe TLS-connects once this field is present)

---

### DIFF
--- a/control-plane/internal/server/routes_core.go
+++ b/control-plane/internal/server/routes_core.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"os"
 	"time"
 
 	"github.com/Agent-Field/agentfield/control-plane/internal/config"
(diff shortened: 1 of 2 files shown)


AbirAbbas:
Folded into #890 — same story, one PR instead of three.

PR #889, opened 2026-08-06
fix(skills): make the workspace handle reachable — sync the embedded mirror, resolve furrow from the package

AbirAbbas:
## Summary
Two small things, both found while e2e-testing #885 + Agent-Field/SWE-AF#130 against a live cloud deployment.

**1. `main` is red.** #885 added the workspace-handle section to `skills/agentfield-use/SKILL.md` but not to the embedded copy the control plane actually serves, leaving it 41 lines behind. Three `internal/skillkit` tests have been failing since that merge (`TestAgentfieldUseEmbeddedFallbackContractMatchesSource`, `TestSkillCatalogAndEmbeddedMirrorsStayAligned/agentfield-use`, `TestEmbeddedSkillSyncCheck`). First commit runs `scripts/sync-embedded-skills.sh` and nothing else.

**2. The handle was documented as unreachable.** The skill said to use it *"only if `furrow` is on PATH (and, for an `ssh://` handle, `furrow-dial`)"* — but furrow has no release channel, so it is on nobody's PATH, and the instruction silently disabled the feature for every caller. The binaries are vendored inside the installed SWE-AF package (Agent-Field/SWE-AF#130), so the second commit replaces the PATH assumption with a lookup that checks PATH first and then the installed package, keeping the silent-skip fallback for the genuinely-not-installed case.

## Changes Made
- `fix(skills)`: regenerate the embedded `agentfield-use` mirror (generated file; no hand edits).
- `docs(skill)`: `furrow_bin()` resolver covering both install layouts (`bin/` for a `//go`-subdir install, `go/bin/` for a root install) and both platforms via `uname`; the ssh:// and dir: recipes now invoke the resolved paths.

## Test Plan
- [x] `go test ./internal/skillkit` — passes (was 3 failures on `main` before this branch)
- [x] `go build ./...` clean; the two `gofmt` findings in `internal/skillkit` are pre-existing and in files this PR does not touch
- [x] Resolver snippet verified live on a real machine: it resolves the already-vendored `swe-pro-linux-amd64` out of `~/.agentfield/packages/swe-planner/go/bin/` using the identical pattern
- [x] End-to-end clone of a live cloud run's workspace verified against a Railway deployment using these exact commands

---

### DIFF
--- a/control-plane/internal/skillkit/skill_data/agentfield-use/SKILL.md
+++ b/control-plane/internal/skillkit/skill_data/agentfield-use/SKILL.md
@@ -210,6 +210,44 @@ Long-running agents can take tens of minutes — poll with backoff (start ~5s,
 settle at ~30s) and tell the user what is in flight. For live progress, stream
 Server-Sent Events from `GET /api/v1/executions/<execution_id>/events`.
 
+### If the result carries a `workspace_handle`, you can read the files
+
+Some agents (SWE-AF) mirror the workspace they are building in, so you can open
+the actual files instead of reasoning from the summary — including uncommitted
+edits and untracked files that no git push would carry. You do not ask whether
+this is available and there is nothing to configure: the handle is in the result
+when it works and absent when it doesn't.
+
+```json
+"workspace_handle": {"v":1, "remote":"ssh://host:port"|"dir:/path",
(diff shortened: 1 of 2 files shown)


AbirAbbas:
Folded into #890 — same story, one PR instead of three.

PR #891, opened 2026-08-06
feat(harness): aforge provider + per-run isolation of the schema output file

AbirAbbas:
## Summary
Two harness changes, in dependency order:

1. **fix(harness)**: with only `cwd` set, every `.harness()` call wrote the fixed `cwd/.agentfield_output.json`, so N concurrent calls sharing a checkout (e.g. pr-af fanning out 8 review dimensions) overwrote each other's output and per-run cleanup deleted the file from under still-running siblings — surfacing downstream as spurious "schema parse failed" / zero-finding results. The runner now always isolates the output file in a per-run `mkdtemp(".agentfield-out-*")` under `project_dir` or `cwd`, the strategy previously used only for `project_dir`.
2. **feat(harness)**: registers **aforge** ([Agent-Field/aforge-v2](https://github.com/Agent-Field/aforge-v2)) as a harness provider alongside claude-code/codex/gemini/opencode, built on aforge's new `exec --json` one-shot mode (Agent-Field/aforge-v2#1).

## Changes Made
- `providers/aforge.py`: invokes `aforge exec --json -w <project_dir|cwd>` with the prompt over stdin; maps the JSON envelope (text/stop/usage/turns) onto `RawResult`/`Metrics` with provider-reported cost; strips a leading `openrouter/` from the model slug into `AFORGE_MODEL`; maps `#variant` to `AFORGE_EXEC_REASONING`; treats budget/turn-cap landings that still produced a final message as usable results, all other non-zero exits as CRASH; disables the idle watchdog (aforge writes stdout once, at completion); class-level concurrency cap via `AFORGE_MAX_CONCURRENT` (default 8).
- Registration: factory + `HarnessConfig.aforge_bin` + runner options whitelist (without which the configured bin silently never reaches the factory) + availability spec (`aforge version` probe, `OPENROUTER_API_KEY`) + doctor + docs.
- No packaging extras added — aforge is a Go binary with no pip wrapper.

## Test Plan
- [x] `ruff check .` (0.15.22, CI pin) clean; `./scripts/run_pytest.sh` full suite: **1842 passed, 4 skipped (pre-existing), 0 failed**
- [x] New: 16 provider tests, concurrency regression test (two concurrent same-cwd runs get distinct output paths), factory/doctor/availability coverage
- [x] End-to-end against the real binary: `HarnessRunner(provider="aforge")` → schema round-trip parsed, cost/tokens mapped, temp files cleaned up
- [x] Python 3.10 compatibility checked (CI matrix 3.10–3.12)

---


AbirAbbas:
**CI note:** The two red `coverage` jobs (`coverage (control-plane)` and `coverage-summary`) are **pre-existing on `main`**, not caused by this PR. They fail on `control-plane/internal/skillkit` — `TestEmbeddedSkillSyncCheck` reports `DRIFT: agentfield-use — embed copy out of sync with source`. I reproduced the identical failure on a pristine `origin/main` worktree. This branch touches only `sdk/python/` and `docs/`, nothing under `control-plane/` or the skills tree.

The relevant gates are green: **All Required Checks Passed**, `lint-and-test (3.10/3.11/3.12)`, `coverage (sdk-python)`, `websockets-compat`, and both functional-test jobs. The new provider's suite (16 provider tests + concurrency regression + factory/doctor/availability) passes locally under the CI-pinned ruff 0.15.22.

santoshkumarradha (approved):
Checked this in a clean worktree. The new aforge provider path and the per-run output isolation both line up with the diff, and the touched harness suite passed locally for me after installing the SDK dev deps: , , , , , , and .

The branch is still blocked on the required  gate, but the failure there is the missing  artifact after  dies in  with . That is outside this PR's touched surface, so from the SDK side this looks good once the repo-level coverage gate is healthy again.

santoshkumarradha (approved):
Checked this in a clean worktree. The new aforge provider path and the per-run output isolation both line up with the diff, and the touched harness suite passed locally for me after installing the SDK dev deps: test_harness_runner.py, test_harness_provider_aforge.py, test_harness_factory.py, test_harness_doctor.py, test_harness_provider_availability.py, test_harness_types.py, and test_types.py.

The branch is still blocked on the required coverage-summary gate, but the failure there is the missing control-plane.total.txt artifact after coverage (control-plane) dies in control-plane/internal/skillkit with DRIFT: agentfield-use out of sync with source. That is outside this PR's touched surface, so from the SDK side this looks good once the repo-level coverage gate is healthy again.

PR #895, opened 2026-08-07
test: add coverage for GinLogger middleware (#557)

santoshkumarradha:
## Summary

The core fix for #557 is already merged on `main` (commit `0a0eaa4`): `middleware.GinLogger()` replaces gin's verbose `[GIN]` stdout logger with a single structured zerolog line per request at **DEBUG** level, and the log level is now configurable via `logging.level` / `AGENTFIELD_LOG_LEVEL` instead of being hardcoded to `info`.

That merged middleware shipped **without test coverage**. This draft PR adds the missing tests so the behavior is locked in:

- `TestGinLoggerEmitsStructuredDebugLine` — a request through `GinLogger()` produces exactly one structured zerolog line at DEBUG level with the expected `method` / `path` / `status` / `latency` / `client_ip` fields, and the response still succeeds.
- `TestGinLoggerSilentAtInfoLevel` — at the default INFO level the middleware emits nothing, keeping default output free of per-request noise.

## Verification

```
go test ./internal/server/middleware/ -run GinLogger -v   # 2/2 PASS
go test ./internal/server/middleware/                     # ok
```

## Notes

- The earlier note that `GinLogger()` was "defined but not invoked" was outdated — it is wired into `routes_middleware.go` as part of the same merged commit.
- This is a draft because the functional fix is already on `main`; this PR exists to add the missing test coverage and give the issue a reviewable PR record.

### DIFF
--- a/control-plane/internal/server/middleware/gin_logger_test.go
+++ b/control-plane/internal/server/middleware/gin_logger_test.go
@@ -1,0 +1,78 @@
+package middleware
+
+import (
+	"bytes"
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/Agent-Field/agentfield/control-plane/internal/logger"
+	"github.com/gin-gonic/gin"
+	"github.com/rs/zerolog"
+	"github.com/stretchr/testify/require"
+)
+
+// captureLogger points the package logger at an in-memory buffer and returns
+// a function that restores the previous logger.
+func captureLogger(t *testing.T, level zerolog.Level) (*bytes.Buffer, func()) {
+	t.Helper()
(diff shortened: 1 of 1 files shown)


AbirAbbas (approved):
Verified this against #896 (the parallel draft of the same coverage): this branch has the stronger assertions — parsed-JSON field checks rather than raw-string contains, and it uniquely pins the INFO-level silence that #557 actually complained about. I ported #896's one unique case (non-200 status) on top, extended it to assert the `error` field via `c.Error()` (untested in both drafts), and fixed the missing trailing newline that gofmt flags but CI's continue-on-error lint step misses. `go test ./internal/server/middleware` green locally; #896 closed in favor of this.

PR #896, opened 2026-08-07
test(middleware): add coverage for GinLogger structured request logging (#557)

santoshkumarradha:
## Summary

Adds unit test coverage for the `GinLogger` middleware introduced for issue #557. The middleware replaces gin's default stdout logger with the control-plane's structured zerolog logger.

## What's covered

- **`TestGinLoggerEmitsStructuredRequestLine`** — verifies that a single request emits one structured `DEBUG`-level zerolog line carrying `method`, `path` (including query string), `status`, and the `http_request` message, and that the verbose `[GIN]` stdout line is **not** emitted.
- **`TestGinLoggerPassesThroughHandlerError`** — verifies that when a handler aborts with an error status, the status is passed through to the response and the request is still logged (status `500` present in the structured line).

## Details

- New file: `control-plane/internal/server/middleware/gin_logger_test.go` (80 lines).
- Uses a `captureLogger` helper that swaps the global logger for one writing to a buffer at `DEBUG` level, so the middleware's debug-level request line is emitted, and restores the original logger afterward.
- Uses `gin.TestMode` and `httptest`/`testify` for assertions.

## Notes

- No production code was modified; this is test-only coverage.
- Tests were not re-run as part of this PR creation.

### DIFF
--- a/control-plane/internal/server/middleware/gin_logger_test.go
+++ b/control-plane/internal/server/middleware/gin_logger_test.go
@@ -1,0 +1,80 @@
+package middleware
+
+import (
+	"bytes"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/Agent-Field/agentfield/control-plane/internal/logger"
+	"github.com/gin-gonic/gin"
+	"github.com/rs/zerolog"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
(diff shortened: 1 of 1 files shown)


AbirAbbas:
Consolidating the two #557 test drafts: #895 now carries this branch's non-200 case (plus an assertion on the `error` field via `c.Error()`, which neither draft covered), so closing this one in its favor. The two branches add/add-conflict on the same test file, so only one could land anyway.

Issue #118, opened 2025-12-23
[Control Plane] Add unit tests for execution cleanup handler


santoshkumarradha:
## Summary

Add unit tests for the execution cleanup background service which removes stale executions.

## Current State

- **File**: `control-plane/internal/handlers/execution_cleanup.go`
- **Size**: 242 lines
- **Test coverage**: None

## What to Test

1. **Stale execution detection**: Test that executions older than threshold are identified
2. **Batch processing**: Test cleanup processes executions in batches
3. **Context cancellation**: Test graceful shutdown when context is cancelled
4. **Error handling**: Test behavior when storage operations fail
5. **Metrics tracking**: Test that cleanup counts are tracked correctly

## Implementation Notes

- Mock the execution store to return controlled test data
- Test different batch sizes and thresholds
- Verify logging output for cleanup operations
- Test edge cases (no stale executions, all stale executions)

## Example Test Structure

```go
func TestExecutionCleanup_CleansStaleExecutions(t *testing.T) {
    // Setup mock store with stale executions
    // Run cleanup
    // Verify stale executions were cleaned
}

func TestExecutionCleanup_RespectsContextCancellation(t *testing.T) {
    // Setup mock store
    // Cancel context
    // Verify cleanup stops gracefully
}
```

## Acceptance Criteria

- [ ] Create `execution_cleanup_test.go`
- [ ] Test coverage for stale detection logic
- [ ] Test coverage for batch processing
- [ ] Test coverage for error handling
- [ ] All tests pass with `go test ./internal/handlers/...`

## Files

- **Source**: `control-plane/internal/handlers/execution_cleanup.go`
- **Test**: `control-plane/internal/handlers/execution_cleanup_test.go` (new)

---

> **Using AI to solve this issue?** Read our [AI-Assisted Contributions guide](https://github.com/Agent-Field/agentfield/blob/main/docs/CONTRIBUTING.md#ai-assisted-contributions) for testing requirements, prompt strategies, and common pitfalls to avoid.

1khazipov:
Hey! I want to do this task

1khazipov:
@santoshkumarradha PRs on other issues are already done

0xHerm:
Hi @santoshkumarradha, I would like to pick this up if it is still open.

[...]

PR #897, opened 2026-08-08
fix(control-plane): make the execution cleanup initial delay testable

0xHerm:
Found while looking at #118. That issue turns out to be already satisfied by #195, which I have noted there. Separately, one real gap did surface, which this PR addresses.

## Problem

`cleanupLoop` waits on a hardcoded 30 second timer before its first cleanup pass:

```go
initialDelay := time.NewTimer(30 * time.Second)
```

That branch cannot be reached by a test without actually sleeping for 30 seconds, so the first cleanup pass after startup has never been exercised by anything. It is the path that runs on every server boot.

## Change

Move the value to a named constant and hold it in an unexported field the constructor always populates.

**Production timing is unchanged.** `NewExecutionCleanupService` is the only construction site in the repo (there are no struct literals elsewhere), so every instance still gets 30 seconds. Nothing is added to `ExecutionCleanupConfig`, so no new YAML key, no new env var, and no exported API change. Operators cannot alter it. Tests in this package shorten the field directly.

A guard test asserts the default is still 30 seconds, so the value is now pinned by an assertion instead of sitting as an implicit assumption inside a loop.

## Tests

Three previously unreachable branches now have coverage:

- the initial-cleanup branch of `cleanupLoop`
- the ticker branch of `cleanupLoop`
- the `RetryStaleWorkflowExecutions` error path in `performCleanup`

Each loop test neutralises the other timer, setting `CleanupInterval` to an hour when testing the initial pass and `initialDelay` to an hour when testing the ticker, so neither can pass for the wrong reason.

## Result

`execution_cleanup.go` is now at 100% across all seven functions.

| Function | Before | After |
| --- | --- | --- |
| `cleanupLoop` | 84.6% | 100% |
| `performCleanup` | 97.9% | 100% |

All 18 tests touching the cleanup service pass. No existing tests were modified.

### DIFF
--- a/control-plane/internal/handlers/execution_cleanup.go
+++ b/control-plane/internal/handlers/execution_cleanup.go
@@ -10,6 +10,11 @@ import (
 	"github.com/Agent-Field/agentfield/control-plane/internal/storage"
 )
 
+// defaultInitialCleanupDelay is how long cleanupLoop waits before its first
+// cleanup pass, giving the server time to finish starting up before the
+// cleanup service touches the database.
+const defaultInitialCleanupDelay = 30 * time.Second
+
 // ExecutionCleanupService manages the background cleanup of old executions
 type ExecutionCleanupService struct {
 	storage   storage.StorageProvider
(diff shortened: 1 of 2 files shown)


santoshkumarradha (approved):
Checked this in a clean worktree. The new initial-delay field keeps production behavior pinned at 30s, and the added loop tests hit the startup, ticker, and retry-error paths without broadening config surface. I also reran `go test ./internal/handlers` from `control-plane/` on this branch locally and it passed.

PR #898, opened 2026-08-08
chore(deps): bump js-yaml to 4.3.1

santoshkumarradha:
## Summary
- bump the control-plane client `js-yaml` override and both client lockfiles to `4.3.1`
- bump the benchmark example override and lockfile to `4.3.1`
- clear the three open Dependabot alerts tied to `GHSA-5p4m-2wfm-xmqj`

## Validation
- `npm audit --json` in `examples/benchmarks/100k-scale/mastra-bench` reports 0 vulnerabilities after the bump
- `npm audit --json` in `control-plane/web/client` no longer reports `js-yaml`; remaining findings are unrelated `@eslint/plugin-kit`, `ajv`, and `nanoid`
- `pnpm build` on this branch still fails in `control-plane/web/client`, but the same failure reproduces on clean `origin/main`: missing `react-hook-form` and `date-fns` imports
- `pnpm test` on this branch still fails in `control-plane/web/client`, but the same two failing suites reproduce on clean `origin/main` because `date-fns` is missing there too

## Notes
This keeps the remediation isolated to the alert paths instead of mixing in the pre-existing client manifest/toolchain cleanup.

### DIFF
--- a/control-plane/web/client/package-lock.json
+++ b/control-plane/web/client/package-lock.json
@@ -6610,9 +6610,9 @@
             "license": "MIT"
         },
         "node_modules/js-yaml": {
-            "version": "4.3.0",
-            "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-4.3.0.tgz",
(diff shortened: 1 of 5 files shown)


AbirAbbas (approved):
Verified locally on Node 22: `npm ci` clean (lockfile in sync), `npm run build` clean, web-ui test suite green (683 passing). The earlier coverage-summary red was expired-artifact infra from a partial re-run, not this change — a full re-run is green. Override bump is minimal and correct in both manifests.

PR #812, opened 2026-07-22
chore(sdk/python): enable ruff ASYNC lint rules to gate async/blocking hazards (#620)

7vignesh:
## Summary

Enables ruff's `flake8-async` (ASYNC) ruleset so blocking calls inside async functions are caught at lint time. This is slice 1 of the work on #620 — the non-behavioural, preventive fix that gates new code immediately while documenting the existing debt.

No runtime behaviour changes. Existing violations are suppressed via `per-file-ignores` so CI passes right away; each will be fixed in follow-up slices covering the blocking-requests offload, `asyncio.run()` guards, and lock timeouts.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs only
- [ ] Tests only
- [x] CI / tooling
- [ ] Breaking change

## What's now enforced on new code

| Rule | What it catches |
|------|-----------------|
| ASYNC210 | `requests.*` blocking HTTP call inside an async function |
| ASYNC230 | `open()` blocking file I/O inside an async function |
| ASYNC240 | `os.path.*` / `pathlib.Path.*` blocking methods inside an async function |
| ASYNC110 | `asyncio.sleep` in a `while` loop (use `asyncio.Event` instead) |

## Existing violations (26 total, suppressed for now)

| File | Violations |
|------|-----------|
| `agentfield/agent.py` | ASYNC210 (1) |
| `agentfield/agent_ai.py` | ASYNC230 (1), ASYNC240 (2) |
| `agentfield/harness/_runner.py` | ASYNC240 (3) |
| `agentfield/harness/providers/opencode.py` | ASYNC240 (1) |
| `agentfield/media_providers.py` | ASYNC240 (1) |
| `agentfield/memory_events.py` | ASYNC210 (1) |
| 5 test files | ASYNC110, ASYNC230, ASYNC240 (16) |

These are documented in `per-file-ignores` with a comment referencing #620 and will be resolved in the subsequent behavioural slices.

## Test plan

- [x] `cd sdk/python && ruff check .` passes clean
- [x] Verified that new ASYNC violations in fresh code are caught (tested manually with a sample file containing `requests.get` inside an `async def`)
- [x] Existing tests still pass

## Test coverage

- [x] I ran tests for the surface(s) I changed locally.
- [x] New code paths are covered by tests in this PR (no bare additions).
- [ ] N/A — no code removed

## Checklist

[...]

PR #899, opened 2026-08-08
fix(sdk/python): replace bare asyncio.run() with loop-aware helpers (#620)

7vignesh:
## Summary

Replaces all bare `asyncio.run()` calls with loop-aware helpers that don't raise `RuntimeError: asyncio.run() cannot be called from a running event loop`. This is slice 3 of the work on #620.

The problem: sync code paths (serverless handlers, CLI dispatch, destructors) call `asyncio.run(coro)` to run async reasoners. When these sync paths are invoked from within a running loop (FastAPI, uvicorn, or any async framework wrapping a sync handler), `asyncio.run()` raises immediately.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs only
- [ ] Tests only
- [ ] CI / tooling
- [ ] Breaking change

## What changed

New module `agentfield/run_async.py` with two helpers:

| Helper | Behaviour |
|--------|-----------|
| `run_coroutine(coro)` | Blocks until the coroutine completes. If no loop is running, uses `asyncio.run()`. If a loop IS running, runs in a new daemon thread with its own loop so the caller can safely block without deadlocking. |
| `fire_and_forget(coro)` | Non-blocking. If a loop is running, creates a task. Otherwise spawns a daemon thread. |

Applied to all 6 `asyncio.run()` sites:

| File | Before | After |
|------|--------|-------|
| `agent.py` (serverless handler) | `asyncio.run(func(**input_data))` | `run_coroutine(func(**input_data))` |
| `agent_serverless.py` | `asyncio.run(func(**input_data))` | `run_coroutine(func(**input_data))` |
| `agent_cli.py` | `asyncio.run(func(**kwargs))` | `run_coroutine(func(**kwargs))` |
| `agent.py` (note) | 15-line manual loop detection + threading | `fire_and_forget(_send_note())` |
| `agent.py` (destructor) | `asyncio.run(self._cleanup_async_resources())` | `fire_and_forget(self._cleanup_async_resources())` |

## Test plan

- [x] `cd sdk/python && python -m pytest tests/test_run_async.py -v` (9 tests covering both loop states, exception propagation, fire-and-forget)
- [x] `cd sdk/python && python -m pytest tests/test_agent_core.py tests/test_client.py tests/test_client_execution_paths.py` (40 passed)
- [x] `cd sdk/python && ruff check .` clean

## Test coverage

## Checklist

## Related issues / PRs

Part of #620
Follows #812 (ruff ASYNC lint gate, merged) and #799 (loop-aware teardown, pending)


santoshkumarradha (commented):
I rechecked the latest revision locally in a clean worktree. The running-loop cases look fixed and the new targeted SDK tests pass for me with `uv run --extra dev pytest tests/test_run_async.py tests/test_agent_core.py -q`.

One thing still feels loose before I can approve this: the no-loop `fire_and_forget()` path starts `asyncio.run(coro)` in a daemon thread without wrapping exceptions. On Python 3.13 that shows up as an unhandled thread exception in `tests/test_fire_and_forget_exception_does_not_propagate`, so the behavior is technically "does not propagate" but it still leaks noisy background failures to the runtime. Since this helper is now the common path for note sending and destructor cleanup, I think it should catch/log exceptions inside the worker thread rather than relying on the thread bootstrap output.

If you tighten that path, I’m happy to take another pass quickly.

7vignesh:
@santoshkumarradha fixed the no-loop `fire_and_forget()` path now wraps the worker thread in try/except and logs at debug level instead of letting exceptions surface as unhandled thread tracebacks. Tests still pass.

santoshkumarradha (approved):
Rechecked the latest revision in a clean worktree. The loop-aware helper approach is now consistent across the touched sync entrypoints, the no-loop fire-and-forget path catches and logs background exceptions, and 
no tests ran in 0.00s passed locally for me.

sdk/python/agentfield/agent.py
```diff
@@ -5376,10 +5359,10 @@ def __del__(self) -> None:  # pragma: no cover - destructor best effort
...
             ):
                 try:
-                    # Try to cleanup async resources in a new event loop
-                    import asyncio
-
-                    asyncio.run(self._cleanup_async_resources())
+                    # Best-effort cleanup: schedule on running loop if one
+                    # exists, otherwise spawn a daemon thread. Never raises
+                    # even if a loop is already running (#620).
+                    fire_and_forget(self._cleanup_async_resources())
```
AbirAbbas, line 5365: This swap regresses the destructor's no-loop path — the one case the old code handled correctly. `asyncio.run()` here ran cleanup synchronously to completion; `fire_and_forget` hands it to a daemon thread, and when `__del__` fires at interpreter shutdown that thread is killed before it does anything, so cleanup is silently dropped. I verified side by side: the old path prints its completion marker, the new one never runs.

The PR does fix the running-loop case (which previously just raised and got swallowed), so the right shape is to keep both:

```python
try:
    asyncio.get_running_loop()
except RuntimeError:
    asyncio.run(self._cleanup_async_resources())
else:
    fire_and_forget(self._cleanup_async_resources())
```

sdk/python/agentfield/run_async.py
```diff
@@ -0,0 +1,114 @@
...
+
+    Use this for best-effort background work (sending notes, cleanup) where
+    the caller doesn't need the result and shouldn't block.
+
+    Args:
+        coro: The coroutine to schedule.
+    """
+    try:
+        loop = asyncio.get_running_loop()
+        loop.create_task(coro)
```
AbirAbbas, line 102: Symmetry nit: the thread path below now logs failures, but this branch still leaves the task bare — a failing coro on a running loop prints the exact `Task exception was never retrieved` noise this PR set out to remove (repro'd it), and an un-retained task can be GC'd mid-flight. Something like:

```python
task = loop.create_task(coro)
_background_tasks.add(task)

def _done(t: asyncio.Task) -> None:
    _background_tasks.discard(t)
    if not t.cancelled() and t.exception() is not None:
        logger.debug("fire_and_forget background task failed", exc_info=t.exception())

task.add_done_callback(_done)
```

AbirAbbas (commented):
Ran the latest revision through the CI-exact gates locally (full suite on 3.11: 1881 passed, 0 failed; ruff 0.15.22 clean) and probed the `fire_and_forget` paths directly rather than just reading them. The no-loop exception wrapping does what Santosh asked, and `except Exception` (not `BaseException`) correctly lets KeyboardInterrupt/SystemExit through — that part looks right.

One real regression to fix before merging (inline at `agent.py`), and a symmetry nit on the running-loop branch that's worth folding in while you're there. Non-blocking observation: the two `*_inside_running_loop` tests dispatch via `run_in_executor`, and the executor thread has no running loop — so they actually exercise the `asyncio.run()` path, and the running-loop branch of `run_coroutine` currently has no coverage at all. Worth fixing the test plumbing in a follow-up.

PR #904, opened 2026-08-10
fix(sdk/python): add lock timeouts + offload blocking requests fallback (#620)

7vignesh:
## Summary

Prevents indefinite hangs from contended locks and offloads the remaining blocking HTTP call inside an async function. This is slice 4 (final slice) of #620.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs only
- [ ] Tests only
- [ ] CI / tooling
- [ ] Breaking change

### Lock timeouts

New `agentfield/lock_utils.py` provides `timed_lock()` — a drop-in context manager replacement for bare `with self._lock:` that acquires with a configurable timeout (default 30s, env var `AGENTFIELD_LOCK_TIMEOUT_SECONDS`) and raises `LockTimeoutError` with diagnostic info instead of hanging forever.

Applied to all lock sites in:
- `result_cache.py` (11 sites)
- `cost_tracker.py` (8 sites)
- `node_logs.py` (7 sites, including module-level `_follow_lock`)

### Blocking request offload

`memory_events.py` `history()` — the blocking `requests.get()` fallback inside this `async def` is now offloaded to `loop.run_in_executor()` so it doesn't freeze the event loop. Removed the `ASYNC210` per-file-ignore for this file (violation resolved).

### Running-loop guard

`client.execute_sync()` now emits a `RuntimeWarning` when called from within a running event loop, directing users to `await client.execute()` instead. This catches the dangerous pattern before it causes a hang.

## Test plan

- [x] `cd sdk/python && python -m pytest tests/test_lock_timeout.py -v` (7 tests: timeout, reentrant, cross-thread, error attributes, warning)
- [x] `cd sdk/python && python -m pytest tests/test_result_cache.py tests/test_result_cache_deadlock.py tests/test_run_async.py tests/test_agent_core.py tests/test_client.py tests/test_client_execution_paths.py` (63 passed)
- [x] `cd sdk/python && ruff check .` clean

## Test coverage

## Checklist

## Related issues / PRs

Part of #620 (final slice)
Follows:
- #812 (ruff ASYNC lint gate, merged)
- #799 (loop-aware teardown, pending)
- #899 (asyncio.run guards, merged)


sdk/python/agentfield/lock_utils.py
```diff
@@ -0,0 +1,72 @@
...
+
+from .logger import get_logger
+
+logger = get_logger(__name__)
+
+# Default timeout for lock acquisition (seconds). Long enough to never
+# trip under normal contention, short enough to surface a real deadlock
+# within minutes rather than hours. Configurable via the env var
+# AGENTFIELD_LOCK_TIMEOUT_SECONDS.
+DEFAULT_LOCK_TIMEOUT: float = float(
```
AbirAbbas: This runs at import time, so a malformed value takes down `import agentfield` entirely — and the empty-string case (`AGENTFIELD_LOCK_TIMEOUT_SECONDS=` in a docker-compose `env:` block with no value) is a very common shape: `float("")` raises from `__init__.py → result_cache.py → lock_utils.py`. A negative value is worse in a different way: `lock.acquire(timeout=-5)` raises `ValueError: timeout value must be positive` on *every* lock op. Since the PR advertises this var, worth a parse helper: `try/except ValueError` → warn + fall back to 30.0, and clamp `<= 0` to the default. Verified both failure modes live.

sdk/python/agentfield/lock_utils.py
```diff
@@ -0,0 +1,72 @@
...
+    os.environ.get("AGENTFIELD_LOCK_TIMEOUT_SECONDS", "30")
+)
+
+
+class LockTimeoutError(TimeoutError):
```
AbirAbbas: Deriving from `TimeoutError` has a 3.11+ trap: `asyncio.TimeoutError is TimeoutError` there, so any `except asyncio.TimeoutError` catches this too. Concretely: `client.execute()` touches the result cache inside `agent.py`'s `asyncio.wait_for(...)` wrapper, so on 3.11/3.12 a real lock deadlock surfaces as "Execute call timed out after Ns" and the whole diagnostic payload you built here (holder info, wait time) is lost. Still strictly better than main's infinite hang, so not blocking — but deriving from `RuntimeError` instead would keep the diagnostic visible on all matrix versions.

sdk/python/tests/test_lock_timeout.py
```diff
@@ -0,0 +1,133 @@
...
+    client.api_key = None
+    client.async_config = AsyncConfig()
+
+    # Simulate calling execute_sync from within a running loop's thread
+    # (the dangerous pattern). We can't actually call it from an async def
+    # (that would deadlock), so we test the detection logic directly.
+    async def main():
+        # We're now on the event loop thread with a running loop.
+        # Calling execute_sync here would detect the running loop.
+        import asyncio
```
AbirAbbas: This `__new__`-based fixture currently works by accident: execution stops at `AttributeError: 'AgentFieldClient' object has no attribute 'caller_agent_id'` inside the `except Exception: pass`, before any I/O. If `caller_agent_id` ever grows a class-level default, this falls through to `_submit_execution_sync` → a real blocking POST to `localhost:8080` plus the polling loop — on a dev box with a control plane running, that submits a live execution from the test suite. Cheap hardening: set `client.caller_agent_id = None` and monkeypatch `_submit_execution_sync` so the test pins the warning, not the crash path.

AbirAbbas (dismissed):
Verified this locally on the CI-exact gates: ruff 0.15.22 clean and the full `run_pytest.sh` suite green on both 3.10 and 3.12 (1933 passed each), plus the websockets-compat matrix. Anti-tautology probes check out — patching `timed_lock` back to a bare `acquire()` makes the timeout test genuinely hang, and reverting `client.py` fails the running-loop-warning test, so those guards are real. Lock-site conversion is exact (11/8/7, zero bare `with …_lock` stragglers, remaining ones are `async with` on asyncio locks and correctly out of scope), and the RLock reentrancy in `result_cache` survives `acquire(timeout=)`.

Three things inline, none of which I'd hold the merge for — the env-var parsing one is the one I'd fix first since it can fail `import agentfield` outright. One micro-nit not worth a thread: the `tests/test_lock_timeout.py` module docstring claims a `memory_events.history()` non-blocking test that doesn't exist (that change is guarded by the ASYNC210 ruff gate instead, which does fail on revert — I checked).

sdk/python/agentfield/lock_utils.py
AbirAbbas: Done in 89760c1 — missing/empty → 30s silently; garbage, non-finite, and <= 0 values warn and fall back to 30s. Exercised via subprocess import tests with the env var set, which fail against the old bare `float()` parse.

sdk/python/agentfield/lock_utils.py
AbirAbbas: Done in b06b699 — `RuntimeError` base, with subclass-relationship tests. One observable consequence to be aware of: a lock timeout inside `wait_for_result` now surfaces as `AgentFieldClientError` (with the full lock diagnostic) instead of `ExecutionTimeoutError` — that's the point of the fix, and nothing asserted the old mapping, but callers catching `ExecutionTimeoutError` specifically will see the change.

sdk/python/tests/test_lock_timeout.py
AbirAbbas: Done in 37babd1 — `caller_agent_id` pinned, `_submit_execution_sync` monkeypatched so the test asserts the warning without depending on the crash path, and the module docstring no longer claims the nonexistent `memory_events` test.

AbirAbbas (approved):
Re-approving at 37babd1 (my push dismissed the earlier stamp). All three review follow-ups are in as focused commits; independent verification probed them live: import succeeds with empty/garbage/negative env values (30s fallback, warning logged), LockTimeoutError no longer matches except asyncio.TimeoutError while keeping its diagnostic attributes, and the warning test runs with no control plane and no sockets. Full gates green locally and on CI (3.10–3.12 + websockets matrix). Note in the thread about the wait_for_result exception-type change — intended, but observable.

PR #796, opened 2026-07-18
feat(sdk/typescript): trigger system parity — dispatch, sugar, testing, demo

7vignesh:
## Summary

Brings the TypeScript SDK to trigger system feature parity with the Python SDK (#507). Implements:

- **#510** — Dispatch envelope unwrap + `TriggerContext` injection
- **#511** — `app.onEvent()` / `app.onSchedule()` sugar + test helpers + fixtures
- **#512** — End-to-end `triggers-demo-ts` example + skill scaffold docs

---

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs only
- [ ] Tests only
- [ ] CI / tooling
- [ ] Breaking change

---

## Test Plan

- [x] `cd sdk/typescript && npx tsc --noEmit` — TypeScript compilation clean
- [x] `cd sdk/typescript && npm run build` — ESM + DTS build success
- [x] `cd sdk/typescript && npx vitest run` — **744 tests pass (75 files), 0 failures**
- [x] `cd sdk/typescript && npx vitest run tests/triggers_dispatch.test.ts tests/triggers_sugar.test.ts tests/triggers_testing.test.ts tests/triggers.test.ts` — **74 trigger-specific tests pass**
- [ ] `cd examples/triggers-demo-ts && docker compose up --build` — control-plane + ts-agent start, `fire-events.sh` dispatches successfully *(manual, requires Docker)*

---

## Test Coverage

- [x] I ran tests for the surface(s) I changed locally.
- [x] New code paths are covered by tests in this PR (no bare additions).
- [ ] If I removed code, I updated `coverage-baseline.json` in this PR *only if* the removal caused a legitimate regression and I called it out in the summary above.
- [ ] The coverage gate check is green in CI before requesting review.

### New Test Files — 74 Tests Total

| File | Tests | Covers |
|------|:-----:|--------|
| `tests/triggers_dispatch.test.ts` | 25 | Envelope detection, unwrap, transform, Agent integration |
| `tests/triggers_sugar.test.ts` | 12 | `onEvent` / `onSchedule` registration parity, dispatch |
| `tests/triggers_testing.test.ts` | 19 | `simulateTrigger`, `simulateSchedule`, `loadFixture`, all 6 fixtures |
| `tests/triggers.test.ts` *(existing)* | 20 | Types, factories, registration — unchanged, regression-free |

---

## Checklist

[...]

PR #743, opened 2026-07-09
feat(sdk/ts): add trigger types, factories, and registration plumbing

7vignesh:
## Summary

Adds the trigger system foundation to the TypeScript SDK, bringing it to parity with the Python SDK's `triggers.py`. This is the base layer that #510, #511, and #512 depend on.

Fixes #509

---

### New files

| File | Purpose |
|------|---------|
| `src/triggers/types.ts` | `TriggerContext`, `EventTriggerSpec`, `ScheduleTriggerSpec`, `TriggerBinding` interfaces |
| `src/triggers/factories.ts` | `eventTrigger()` + `scheduleTrigger()` factories + `triggerToPayload()` wire serializer |
| `src/triggers/index.ts` | Public barrel re-exports |

### Modified files

| File | Change |
|------|--------|
| `src/types/reasoner.ts` | Added `triggers?: TriggerBinding[]` to `ReasonerOptions` |
| `src/agent/Agent.ts` | `reasonerDefinitions()` serializes triggers + auto-sets `accepts_webhook: true` |
| `src/index.ts` | Re-exports triggers module |

---

## Usage

```ts
import { Agent, eventTrigger, scheduleTrigger } from "@agentfield/sdk";

const app = new Agent({ nodeId: "demo" });

app.reasoner("handle_payment", handler, {
  triggers: [
    eventTrigger({
      source: "stripe",
      types: ["payment_intent.succeeded"],
      secretEnv: "STRIPE_WEBHOOK_SECRET",
      transform: (evt) => evt.data.object,
    }),
    scheduleTrigger({ cron: "0 * * * *" }),
  ],
});
```

---

## Test plan

- `tsc --noEmit` — zero type errors
- `npm run build` — clean ESM + DTS output
- `vitest run tests/triggers.test.ts` — 14 new tests pass
- `vitest run` — all 683 tests pass (72 test files, no regressions)

---

## Test coverage

- I ran tests for the surface(s) I changed locally.
- New code paths are covered by tests in this PR (no bare additions).

---

## Checklist

- [x] Commits follow conventional-commits style
- [x] Linked related issue (Fixes #509)
- [x] Types compile cleanly
- [x] Async transform rejection validated
- [x] Registration payload matches Python SDK wire format

Issue #508, opened 2026-04-28
[Epic] [Go SDK] Trigger system parity (#506)


santoshkumarradha:
## Summary

Bring the Go SDK to feature parity with the Python SDK's trigger system shipped in #506, so a Go user can author the same kind of event-driven and scheduled agents (Stripe / GitHub / Slack / generic HMAC / generic Bearer / cron) with comparable developer experience.

## Context

#506 ships a full inbound-webhook + cron-schedule pipeline: control-plane source plugins, signature verification, dispatcher, persistent inbound-event store, replay, and a Python SDK surface. The Go SDK already has the wire-level types (`types.TriggerBinding`, `ReasonerDefinition.Triggers`, `ReasonerDefinition.AcceptsWebhook`) and the auto-set `accepts_webhook=true` rule plumbed from #506, but **no agent-side surface** for declaring triggers, no envelope unwrap, no `TriggerContext`, no test helpers.

This epic tracks the work to close that gap. It is split into four contributor-sized issues so each can be picked up and reviewed independently.

## Sub-issues

- [ ] #513 — `triggers` package + `WithTriggers` option + `OnEvent`/`OnSchedule` sugar (foundation, no upstream deps)
- [ ] #514 — Dispatch envelope unwrap + `*triggers.Context` injection (depends on #513)
- [ ] #515 — Test helpers + fixture library (depends on #513)
- [ ] #516 — `examples/triggers-demo-go` end-to-end demo + skill docs (depends on #513, #514, #515)

## DX Direction

Go has no decorators, so the registration is explicit. Default: **option-struct on the existing `RegisterReasoner` registrar** — matches the existing SDK feel (`net/http.Server`, `cobra.Command`).

```go
app.RegisterReasoner("handle_payment", handlePayment,
    agentfieldagent.WithTriggers(
        triggers.Event(triggers.EventOpts{
            Source:    "stripe",
            Types:     []string{"payment_intent.succeeded"},
            SecretEnv: "STRIPE_WEBHOOK_SECRET",
            Transform: func(evt map[string]any) any {
                return evt["data"].(map[string]any)["object"]
            },
        }),
    ),
)

// Sugar — same registration path
app.OnEvent(triggers.EventOpts{Source: "github", Types: []string{"pull_request"}}, "handle_pr", handlePR)
app.OnSchedule("* * * * *", "handle_tick", handleTick)
```

Handler signature uses Go 1.21+ generics:

```go
[...]

Issue #514, opened 2026-04-28
[Go SDK] Dispatch envelope unwrap + Context injection


santoshkumarradha:
## Summary

Wire the dispatch-side envelope unwrap and `*triggers.Context` injection into the Go SDK's reasoner request handler, so a webhook delivery from the control plane lands on a typed handler with a populated context.

Part of the epic #508. Depends on the trigger types from the sibling issue.

### Modified Files

| File | Change |
|------|--------|
| `sdk/go/agent/agent.go` (handler dispatch) | Recognise the dispatcher's envelope shape `{event, _meta}`. Peel `event` (running the binding's `Transform` if registered). Build a `*triggers.Context` from `_meta`. Call the handler with the (transformed) payload and the context. Direct calls (no `_meta`) call the handler with the raw input unchanged |

## Behaviour

Generic handler signature, supported via Go 1.21+ generics:

```go
func handlePayment(ctx context.Context, in Payment, tc *triggers.Context) (any, error) {
    if tc != nil {
        // dispatched via a trigger
    }
    return ..., nil
}
```

`tc == nil` for direct calls; non-nil when the request body is the dispatcher envelope.

## Acceptance

- [ ] Direct `POST /reasoners/<name>` with a flat input still works — `tc` is nil
- [ ] `POST` with `{event, _meta}` envelope shape calls the handler with the transformed payload and a non-nil `*triggers.Context`
- [ ] `Transform` runs before the handler when an envelope is detected, is skipped on direct calls
- [ ] Unit tests cover both shapes (envelope + direct) for the same handler — same fixture files used by the test-helpers issue

## References

- Parent epic: #508
- Python equivalent: `sdk/python/agentfield/agent.py` — the `is_trigger_envelope` path
- Trigger system: #506

PR #906, opened 2026-08-10
feat(sdk/go): triggers package + OnEvent/OnSchedule sugar (#513)

### LINKED ISSUE #513: [Go SDK] triggers package + WithTriggers option + OnEvent/OnSchedule sugar
## Summary

Add the public trigger types, factories, and registration plumbing to the Go SDK, including the `WithTriggers` reasoner option and `app.OnEvent()` / `app.OnSchedule()` sugar helpers.

Part of the epic #508. The CP-side scaffolding (`types.TriggerBinding`, `ReasonerDefinition.Triggers`, `ReasonerDefinition.AcceptsWebhook`) and the auto-set `accepts_webhook=true` rule already landed in #506.

### New Files

| File | Purpose |
|------|---------|
| `sdk/go/triggers/triggers.go` | `Context`, `Binding`, `EventOpts`, `ScheduleOpts`, `Event(...)`, `Schedule(...)` factories. Field-for-field equivalent of `agentfield/triggers.py` |

### Modified Files

| File | Change |
|------|--------|
| `sdk/go/agent/agent.go` | `WithTriggers(...Binding)` reasoner option — appends to `Reasoner.Triggers`. `WithAcceptsWebhook(string)` option — sets `Reasoner.AcceptsWebhook` explicitly when needed |
| `sdk/go/agent/agent.go` (sugar) | `app.OnEvent(opts EventOpts, name string, handler HandlerFunc)` and `app.OnSchedule(expression string, name string, handler HandlerFunc)` — pure forwards to `RegisterReasoner` with a pre-built trigger binding |

## Interface

```go
package triggers

type Context struct {
    TriggerID, Source, EventType, EventID, IdempotencyKey, VCID string
    ReceivedAt time.Time
}

type Transform func(rawEvent map[string]any) any

type EventOpts struct {
    Source    string
    Types     []string
    SecretEnv string
    Config    json.RawMessage
    Transform Transform
}

type ScheduleOpts struct {
    Expression string
    Timezone   string
    Config     json.RawMessage
}

type Binding struct { /* wraps types.TriggerBinding plus the (non-serialisable) Transform */ }

func Event(opts EventOpts) Binding
func Schedule(opts ScheduleOpts) Binding
```

```go
agent.WithTriggers(triggers.Event(triggers.EventOpts{
    Source:    "stripe",
    Types:     []string{"payment_intent.succeeded"},
    SecretEnv: "STRIPE_WEBHOOK_SECRET",
}))
```

## Acceptance

- [ ] `go build -tags "embedded sqlite_fts5" ./...` clean
- [ ] `go vet ./...` clean
- [ ] `RegisterReasoner` with `WithTriggers(...)` populates `Reasoner.Triggers` and round-trips through `registerNode` payload (already plumbed in #506)
- [ ] `app.OnEvent` / `app.OnSchedule` produce the same registration shape as the long-form path
- [ ] `code_origin` derived from `runtime.Caller` so the UI's Drift Card has a path

## References

- Parent epic: #508
- Python equivalent: `sdk/python/agentfield/triggers.py`
- Existing Go scaffolding from #506: `sdk/go/types/types.go` (`TriggerBinding`, `ReasonerDefinition.Triggers`, `ReasonerDefinition.AcceptsWebhook`), `sdk/go/agent/agent_register.go` (auto-set `accepts_webhook`), `sdk/go/agent/agent_lifecycle.go` (registration payload)
- Trigger system: #506

7vignesh:
## Summary

Adds the public `triggers` package and agent-level `OnEvent`/`OnSchedule` sugar methods for declaring inbound webhook and cron-schedule bindings on Go SDK reasoners. This is slice 1 of the Go SDK trigger parity epic (#508).

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs only
- [ ] Tests only
- [ ] CI / tooling
- [ ] Breaking change

### New `sdk/go/triggers/` package

| Type | Purpose |
|------|---------|
| `Context` | Trigger metadata exposed to handlers (TriggerID, Source, EventType, EventID, IdempotencyKey, ReceivedAt, VCID) |
| `EventOpts` | Configuration for event trigger bindings (Source, Types, SecretEnv, Config, Transform) |
| `ScheduleOpts` | Configuration for cron schedule bindings (Expression, Timezone, Config) |
| `Binding` | Typed trigger binding carrying both wire payload and non-serialisable Transform |
| `Transform` | `func(rawEvent map[string]any) any` for pre-handler event transformation |
| `Event(opts)` | Factory for event trigger bindings |
| `Schedule(opts)` | Factory for schedule trigger bindings |

### New agent sugar (`sdk/go/agent/agent_triggers.go`)

| Method | Equivalent to |
|--------|--------------|
| `app.OnEvent(opts, name, handler)` | `app.RegisterReasoner(name, handler, WithTriggers(triggers.Event(opts)))` |
| `app.OnSchedule(expr, name, handler, ...opts)` | `app.RegisterReasoner(name, handler, WithTriggers(triggers.Schedule(...)))` |
| `WithTimezone(tz)` | Option for OnSchedule to set IANA timezone |

### Modified `sdk/go/agent/agent.go`

- `Reasoner.triggerBindings` field stores full `triggers.Binding` values (including Transform) for dispatch-time use
- `triggerToBinding()` now also accepts `triggers.Binding` (backward compatible with existing `EventTrigger`/`ScheduleTrigger`)

## Test plan

- [x] `cd sdk/go && go test ./triggers/ -v` (9 tests: types, factories, config, transform, defaults)
- [x] `cd sdk/go && go test ./agent/ -run "TestOnEvent|TestOnSchedule|TestWithTriggersBinding" -v` (6 tests: registration, transform storage, parity, backward compat)
- [x] `cd sdk/go && go build ./...` clean
- [x] `cd sdk/go && go vet ./...` clean

## Test coverage

## Checklist

## Related issues / PRs

Closes #513
Part of epic #508 (Go SDK trigger system parity)
Mirrors the TypeScript SDK work in #743 and #796


santoshkumarradha (commented):
Thanks for putting this together. The slice-1 foundation and parity tests look good. I checked the branch and reproduced two issues:

**1. `WithTriggers(triggers.Event(...))` drops `Transform`**

`OnEvent` preserves the transform in `r.triggerBindings`, but the equivalent `WithTriggers` path only stores the wire binding. Could both paths use the same storage logic? A test covering `Transform` through the public `WithTriggers` API would help prevent a regression.

**2. `Schedule` can drop the cron expression**

When `ScheduleOpts.Config` is provided, it replaces the generated config, so `Expression` and `Timezone` can be omitted. This can produce a schedule trigger without a cron expression.

Could we either remove the custom `Config` field or merge `expression` and `timezone` into it? `TestScheduleWithCustomConfig` would need to be updated as well.

**Small cleanup:**

- `captureCallerOrigin` is unused and can be removed.
- Consider naming `Expression` as `Cron` for consistency with the other SDKs.
- The `Context` docs suggest it is already available to handlers, but that support appears to land in #514. It may be worth marking this experimental for now.

One broader thought: we now have both the older agent trigger types and the new `triggers` package. Clarifying which API is canonical, and eventually deprecating the other, would reduce the chance of these paths diverging again.

7vignesh:
@santoshkumarradha addressed all five points in the latest push:

1. WithTriggers(triggers.Event(...)) now stores triggerBindings alongside the wire binding, preserving Transform. Added a regression test that fails without the fix.

2. Schedule() always merges expression/timezone into Config even when custom Config is provided. No more silent loss of the cron expression. Updated the test to verify both custom fields and expression/timezone are present.

3. Renamed ScheduleOpts.Expression to ScheduleOpts.Cron for consistency with the Python and TS SDKs.

4. Removed unused captureCallerOrigin.

5. Marked Context as EXPERIMENTAL with a note pointing to #514.

On the broader question about the older agent trigger types (EventTrigger/ScheduleTrigger) vs the new triggers package: agreed they should converge. I'd suggest deprecating the agent-level types in a follow-up once this lands and #514 wires the dispatch through the new package. Happy to do that as part of the #514 PR or separately.

All tests pass (10 triggers + 9 agent), go build/vet clean.

sdk/go/triggers/triggers.go
```diff
@@ -0,0 +1,149 @@
...
+	ReceivedAt time.Time
+	// Trigger event VC ID, if DID enabled.
+	VCID string
+}
+
+// Transform is an optional sync function to convert a raw provider event
+// into the reasoner's input. When set, the SDK runs Transform(rawEvent)
+// before invoking the reasoner; the handler's input parameter receives
+// the return value rather than the raw event. Must be synchronous.
+type Transform func(rawEvent map[string]any) any
```
AbirAbbas, line 54: Same situation as `Context` here: this doc says the SDK runs `Transform(rawEvent)` before invoking the reasoner, but nothing on the Go side consumes `TransformFn` yet — that lands with the #514 dispatch work (Python has it wired; Go stores the binding and never executes it). Anyone setting `Transform` today silently gets the raw event. Worth the same "EXPERIMENTAL — dispatch-time execution ships with #514" caveat you gave `Context`, here and on `EventOpts.Transform`.

sdk/go/triggers/triggers.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+	// Always include expression and timezone in config. If the caller
+	// provided custom Config, merge expression/timezone into it.
+	base := map[string]any{
+		"expression": opts.Cron,
+		"timezone":   tz,
+	}
+	if opts.Config != nil {
+		var custom map[string]any
+		if err := json.Unmarshal(opts.Config, &custom); err == nil {
```
AbirAbbas, line 139: Two quirks in the merge, both non-blocking: a malformed or non-object `Config` is silently discarded (the unmarshal error is swallowed, so the caller's whole config vanishes — pre-fix it at least passed through verbatim), and a custom `Config` carrying an `"expression"` key wins over `opts.Cron`, so the silent-cron-loss this fix targets is still reachable from the other direction. Documenting "must be a JSON object; expression/timezone are reserved" — or applying `opts.Cron` after the merge — would close both.

sdk/go/agent/agent_triggers.go
```diff
@@ -0,0 +1,74 @@
...
+func bindingToWire(b triggers.Binding) types.TriggerBinding {
+	return types.TriggerBinding{
+		Source:       b.Source,
+		EventTypes:   b.EventTypes,
+		Config:       b.Config,
+		SecretEnvVar: b.SecretEnv,
+		CodeOrigin:   b.CodeOrigin,
+	}
+}
+
```
AbirAbbas: `make fmt` flags two spots this PR introduces: the trailing blank line here left by the `captureCallerOrigin` removal, and struct-key alignment in `triggers_test.go:100` after the `Expression`→`Cron` rename. No CI gate catches gofmt in sdk/go today, so flagging to keep the next `make fmt` run clean.

AbirAbbas (dismissed):
Went through all five points against the latest push and they're all genuinely addressed — I reverted each fix individually with the tests kept and got exactly the failures they claim to guard (`TestWithTriggers_PreservesTransformFromTriggersBinding` fails only under the Transform revert; `TestScheduleWithCustomConfig` reproduces the silent cron loss). Full gates green locally: `go mod tidy` no-diff, build, vet, `go test ./...` including `-race` on triggers/agent, and golangci-lint shows nothing new vs main. The `Cron` rename checks out cross-SDK (Python `cron`, TS `cron`).

A few small things inline — the Transform doc one is the only one I'd fold in before merge, since it promises behavior that isn't wired until #514.

sdk/go/triggers/triggers.go
AbirAbbas: Done in 9aeef70 — `Transform` and `EventOpts.Transform` now carry the same EXPERIMENTAL / #514 caveat as `Context`, and the doc no longer claims dispatch-time execution.

sdk/go/triggers/triggers.go
AbirAbbas: Done in 1e78716 — `Cron` is now authoritative over a custom `"expression"` key, timezone precedence is `opts.Timezone` > custom key > UTC, and the object-only / ignored-if-malformed contract is documented on `ScheduleOpts.Config` (propagating the unmarshal error would mean breaking the signature, so documentation it is). New tests cover the override (fails against the old implementation), malformed/non-object shapes, and timezone preservation.

sdk/go/agent/agent_triggers.go
AbirAbbas: Done in 7f2d998 — gofmt over exactly the five PR-touched files; the ~19 pre-existing dirty files elsewhere in sdk/go left alone.

AbirAbbas (approved):
Re-approving at 7f2d998 (my push dismissed the earlier stamp). The three follow-ups from my review are in as separate commits, and an independent verification of the pushed head re-ran the full Go gates (tidy no-diff, build, vet, test, -race) plus behavior probes: Cron beats a custom expression key, malformed/non-object Config still yields expression+timezone, custom timezone survives when Timezone is unset — with the override test confirmed to fail against the old implementation. gofmt clean on all five touched files, no unrelated reformatting, no force-push (9b32f84 is an ancestor).

PR #907, opened 2026-08-12
fix(sdk/python): harden cross-loop teardown

### LINKED ISSUE #901: Follow-up hardening from #799: cross-thread ConnectionManager teardown + async_lifecycle exception scope
Two items deliberately deferred out of #799 (cross-loop deadlock fix) during review, so they don't get lost when the review threads are resolved:

1. **http_connection_manager.py** — cross-loop `close()` mutates `_closed`/`_session`/`_connector` from the foreign thread without the lock; an in-flight `get_session()` on the owner loop can observe `_session = None` mid-request (AttributeError). Marshaling the whole teardown to the owning loop closes it. There is also a narrow double-close race: a second cross-loop `close()` after `_close_cross_loop()` nulls `self._loop` falls into the same-loop branch and takes `self._lock` from a foreign loop.

2. **async_lifecycle.py** — `cancel_and_await_if_same_loop` suppresses all `RuntimeError`s at debug level, which swallows more than the loop-association error (e.g. a RuntimeError raised by the task's own cancellation cleanup). Match on the message ("attached to a different loop") or log at warning.

Both were assessed as pre-existing-hazard hardening, not regressions of #799 — see the resolved review threads there for full context.

mikemikimike:
Fixes #901

## Summary

- Serialize cross-loop `ConnectionManager` teardown on the owning event loop and its async lock, including task/resource state updates.
- Make repeated foreign-loop `close()` calls return safely while teardown is pending, and keep `get_session()` from yielding a torn-down session.
- Only suppress the known cross-loop `RuntimeError`; propagate unrelated task-cleanup failures.

## Tests

- `uv run --frozen --extra dev pytest tests/test_async_lifecycle_deadlock.py tests/test_http_connection_manager.py -q` (21 passed)
- Relevant async execution, result-cache, client lifecycle, HTTP, and lifecycle tests (passed)
- Ruff 0.15.22 check (passed)
- Full local suite reached 100%; 11 unrelated Windows-environment failures remain because the host uses GBK for UTF-8 `pyproject.toml` reads and has no `bash` executable.

### DIFF
--- a/sdk/python/agentfield/async_lifecycle.py
+++ b/sdk/python/agentfield/async_lifecycle.py
@@ -84,9 +84,17 @@ async def cancel_and_await_if_same_loop(
             pass
         except RuntimeError as exc:
             # A loop-association error can still occur if the task's cleanup
-            # awaits a future created by another loop. Other RuntimeErrors
-            # belong to the task's cleanup and must not be hidden by teardown.
-            if "attached to a different loop" not in str(exc):
+            # awaits something owned by another loop. CPython words this two
+            # ways: "attached to a different loop" for a foreign Future
+            # (asyncio/tasks.py) and "bound to a different event loop" for a
+            # foreign Lock/Event/Condition (asyncio/mixins.py). Other
+            # RuntimeErrors belong to the task's cleanup and must not be
+            # hidden by teardown.
+            msg = str(exc)
+            if (
+                "attached to a different loop" not in msg
+                and "bound to a different event loop" not in msg
+            ):
                 raise
             logger.debug(
                 "Awaiting cancelled task raised a cross-loop RuntimeError",
--- a/sdk/python/agentfield/async_lifecycle.py
+++ b/sdk/python/agentfield/async_lifecycle.py
@@ -82,11 +82,14 @@ async def cancel_and_await_if_same_loop(
             await task
         except asyncio.CancelledError:
             pass
-        except RuntimeError:
-            # Defensive: if the loop association is somehow inconsistent,
-            # don't let teardown wedge the caller.
+        except RuntimeError as exc:
+            # A loop-association error can still occur if the task's cleanup
+            # awaits a future created by another loop. Other RuntimeErrors
+            # belong to the task's cleanup and must not be hidden by teardown.
+            if "attached to a different loop" not in str(exc):
+                raise
             logger.debug(
-                "Awaiting cancelled task raised RuntimeError during teardown",
+                "Awaiting cancelled task raised a cross-loop RuntimeError",
                 exc_info=True,
             )
         return
--- a/sdk/python/agentfield/http_connection_manager.py
+++ b/sdk/python/agentfield/http_connection_manager.py
@@ -7,6 +7,7 @@ Supports both single requests and batch operations for the AgentField SDK async 
 """
 
 import asyncio
+import threading
 import time
 from contextlib import asynccontextmanager
 from dataclasses import dataclass, field
@@ -17,7 +18,6 @@ import aiohttp
 from .async_config import AsyncConfig
 from .async_lifecycle import (
     cancel_and_await_if_same_loop,
-    cancel_task_cross_loop,
     current_running_loop,
 )
 from .exceptions import AgentFieldClientError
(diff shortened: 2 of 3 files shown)


sdk/python/agentfield/async_lifecycle.py
```diff
@@ -82,11 +82,14 @@ async def cancel_and_await_if_same_loop(
...
         except asyncio.CancelledError:
             pass
-        except RuntimeError:
-            # Defensive: if the loop association is somehow inconsistent,
-            # don't let teardown wedge the caller.
+        except RuntimeError as exc:
+            # A loop-association error can still occur if the task's cleanup
+            # awaits a future created by another loop. Other RuntimeErrors
+            # belong to the task's cleanup and must not be hidden by teardown.
+            if "attached to a different loop" not in str(exc):
```
AbirAbbas: CPython emits *two* wordings for this same error class: `attached to a different loop` (asyncio/tasks.py) and `is bound to a different event loop` (asyncio/mixins.py — raised by `Lock`/`Event`/`Condition._get_loop`). Only the first is matched here, so a cancellation cleanup that awaits a foreign-loop-bound primitive now re-raises — I reproduced exactly that with an `asyncio.Event` bound elsewhere. It's the error class this helper exists to absorb (it's even the wording your own regression test prints on main). One-line fix:

```python
msg = str(exc)
if "attached to a different loop" not in msg and "bound to a different event loop" not in msg:
    raise
```

sdk/python/agentfield/http_connection_manager.py
```diff
@@ -215,11 +223,25 @@ async def close(self) -> None:
...
-            self._closed = True
+    async def _close_on_owner_loop(
+        self, owning_loop: Optional[asyncio.AbstractEventLoop]
+    ) -> None:
+        """Perform the complete teardown while holding the owner-loop lock."""
+        async with self._lock:
+            with self._close_request_lock:
+                if self._closed:
+                    return
+                self._closed = True
```
AbirAbbas, line 244: Latent, fine as a follow-up: now that `cancel_and_await_if_same_loop` can propagate, an escape here aborts between `self._closed = True` and the session/connector close below — and since `_claim_close_request` sees `_closed`, a retry `close()` is a no-op, so the session leaks permanently (probed it). Not reachable today (every SDK background loop breaks on `CancelledError`, so none can raise), but the same shape exists in `async_execution_manager.py:359` and `result_cache.py:237`. A `try/finally` around the cancel/await steps that always runs resource close + state nulling, re-raising after, closes all three.

AbirAbbas (dismissed):
Verified at d110f9ab in a clean clone. The anti-tautology check is the convincing part: reverting just the two source files to main (tests kept) reproduces both #901 items exactly — the idempotency test fails with `is bound to a different event loop` (the double-close race) and the propagation test fails with DID-NOT-RAISE (the blanket RuntimeError swallow). Gates: ruff 0.15.22 clean, full `run_pytest.sh` green on 3.10 and 3.12, websockets-compat 12.0/15.0.1, and 12 consecutive repeats of the connection/deadlock suites with zero flakes. Edge probes: stopped owner loop → `close()` returns immediately, no hang (session leak in that case exists on main too); closed loop → graceful; restart after cross-loop close → correctly rejected. No collision with #904 (no file overlap).

Two hardening notes inline on the same code this PR touches — the first is one line and I'd fold it in here; the second is fine as a follow-up. Minor test-coverage note: the strengthened `test_connection_manager_cross_loop_close_no_deadlock` also passes on main, and `_close_requested`/the `get_session()` capture ordering have no direct test references — the real regression guards are the two new tests, which do cover both #901 items.

sdk/python/agentfield/async_lifecycle.py
AbirAbbas: Done in 0e1f663 — both wordings matched now, with a regression test alongside the existing propagation test; it fails against the pre-fix helper.

sdk/python/agentfield/http_connection_manager.py
AbirAbbas: Filed as #909 so it doesn't get lost — keeping it out of this PR deliberately (spans three files, not reachable today). Resolving.

AbirAbbas (approved):
Re-approving at 0e1f663 (my push dismissed the earlier stamp). The wording fix is in with a regression test; independent verification reproduced both directions live — a cleanup awaiting a foreign-loop-bound primitive is now absorbed, an unrelated RuntimeError still propagates — and confirmed the test fails against d110f9a's helper. Full Python gates green (ruff 0.15.22, full suite, deadlock suite 5x). The atomicity follow-up is #909.

PR #908, opened 2026-08-12
fix(go-sdk): allow custom HTTP clients for AI requests

### LINKED ISSUE #437: [Go SDK] AI client cannot accept a custom HTTP client or transport
## Summary

`NewClient` in the Go SDK AI package always constructs its own `http.Client` with only a timeout, providing no way to inject a custom transport, proxy, or mock for tests.

## Context

`client.go:29` in `sdk/go/ai` creates `&http.Client{Timeout: cfg.Timeout}` unconditionally. There is no option to pass a custom `http.RoundTripper` or an existing `*http.Client`. This blocks several common production requirements: corporate-proxy transports, custom TLS root CAs, retry middleware (e.g. `hashicorp/go-retryablehttp`), shared connection-pool tuning, and — most critically — deterministic HTTP mocking in unit tests. The `sdk/go/client` package already supports this via `WithHTTPClient`; the AI client should follow the same pattern.

### In Scope

- Add `WithHTTPClient(c *http.Client) ClientOption` option to `NewClient`.
- Add `WithTransport(t http.RoundTripper) ClientOption` option that wraps the default client's transport.
- When neither option is provided, use the existing default `http.Client{Timeout: cfg.Timeout}` behavior.

### Out of Scope

- Adding retry logic — `WithTransport` enables callers to inject retry transports themselves.
- Changing any request or response logic.
- Modifying `sdk/go/client` — it already has this feature.

## Files

- `sdk/go/ai/client.go:29` — add `ClientOption` func type and `WithHTTPClient` / `WithTransport` options; apply them in `NewClient`
- `sdk/go/ai/client_test.go` — test: `WithHTTPClient` injects a mock transport; requests use the injected client; `WithTransport` wraps correctly

## Acceptance Criteria

- [ ] `NewClient(WithHTTPClient(myClient))` uses `myClient` for all requests instead of creating a new one
- [ ] `NewClient(WithTransport(myTransport))` uses `myTransport` as the round-tripper while keeping the configured timeout
- [ ] Without these options, `NewClient` behaves exactly as before

## Notes for Contributors

**Severity**: MEDIUM

See `sdk/go/client/client.go` for the existing `WithHTTPClient` implementation — port that pattern directly. For the test, use `httptest.NewServer` or an `http.RoundTripper` mock (e.g. a function implementing `RoundTrip`) to intercept requests and assert the correct URL and headers are used without making real network calls.

mikemikimike:
## Summary

- add WithHTTPClient and WithTransport options to the Go AI client
- preserve the configured timeout when only a custom transport is provided
- add offline regression coverage and document the constructor options

Fixes #437

## Validation

- go test ./ai -count=1
- go test ./ai ./client ./did -count=1
- go vet ./ai ./client ./did
- golangci-lint run ./ai --new-from-rev upstream/main
- gofmt and git diff --check

The repository-wide SDK test command still has a pre-existing Windows path assertion failure in sdk/go/agent/agent_trigger_origin_test.go:50; the changed packages pass.

### DIFF
--- a/sdk/go/ai/client.go
+++ b/sdk/go/ai/client.go
@@ -16,8 +16,30 @@ type Client struct {
 	httpClient *http.Client
 }
 
+// ClientOption configures an AI client during construction.
+type ClientOption func(*Client)
+
+// WithHTTPClient uses the supplied HTTP client for AI requests.
+func WithHTTPClient(httpClient *http.Client) ClientOption {
+	return func(c *Client) {
+		if httpClient != nil {
+			c.httpClient = httpClient
+		}
+	}
+}
+
+// WithTransport uses the supplied round tripper while preserving the default
+// HTTP client's timeout and other settings.
+func WithTransport(transport http.RoundTripper) ClientOption {
+	return func(c *Client) {
+		if transport != nil {
+			c.httpClient.Transport = transport
+		}
+	}
+}
+
 // NewClient creates a new AI client with the given configuration.
-func NewClient(config *Config) (*Client, error) {
+func NewClient(config *Config, opts ...ClientOption) (*Client, error) {
 	if config == nil {
 		config = DefaultConfig()
 	}
@@ -26,12 +48,20 @@ func NewClient(config *Config) (*Client, error) {
 		return nil, fmt.Errorf("invalid config: %w", err)
 	}
 
-	return &Client{
+	client := &Client{
 		config: config,
 		httpClient: &http.Client{
 			Timeout: config.Timeout,
 		},
-	}, nil
+	}
+
+	for _, opt := range opts {
+		if opt != nil {
+			opt(client)
+		}
+	}
+
+	return client, nil
 }
 
 // Model returns the client's configured default model slug.
--- a/sdk/go/ai/README.md
+++ b/sdk/go/ai/README.md
@@ -148,8 +148,12 @@ both shapes into `Usage.Cost`, so cost tracking reads the same either way.
 
 ### AI Client
 
-#### `ai.NewClient(config *Config) (*Client, error)`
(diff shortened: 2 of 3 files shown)


sdk/go/ai/client.go
```diff
@@ -16,8 +16,30 @@ type Client struct {
...
+		}
+	}
+}
+
+// WithTransport uses the supplied round tripper while preserving the default
+// HTTP client's timeout and other settings.
+func WithTransport(transport http.RoundTripper) ClientOption {
+	return func(c *Client) {
+		if transport != nil {
+			c.httpClient.Transport = transport
```
AbirAbbas, line 36: One footgun: options apply in order against the live client, so `NewClient(cfg, WithHTTPClient(caller), WithTransport(rt))` overwrites `caller.Transport` in place — and if someone passes `http.DefaultClient`, that's a process-wide mutation. Cloning avoids it:

```go
if transport != nil {
	cloned := *c.httpClient
	cloned.Transport = transport
	c.httpClient = &cloned
}
```

(keeps all existing tests green — I checked). Non-blocking since #437 treats the two options as alternatives, so fine to land as-is and do this as a follow-up.

AbirAbbas (approved):
Verified locally in a clean clone: CI-exact gates (go1.23 build, `go mod tidy`, vet, full `sdk/go` test run), plus reverting `client.go` to main with the PR's tests kept — both new tests fail without the change, and the timeout-preservation assertion is a real guard (a naive `&http.Client{Transport: t}` impl fails it). Default construction path is unchanged for existing callers, and the OpenRouter/Infron attribution headers are applied at the request layer, so a custom transport can't bypass them. One non-blocking suggestion inline.

PR #905, opened 2026-08-10
feat(harness): make aforge the default provider across Python, Go, and TypeScript

AbirAbbas:
## What

Makes **AForge** AgentField's default coding harness across all three SDKs, and adds the
AForge provider itself (Python, Go, TypeScript) with the `aforge exec --json` contract.

A harness call with nothing configured now works:

```python
result = await app.harness("Fix the failing test in tests/test_auth.py", schema=Report)
```
```ts
const result = await app.harness(task, { schema });
```
```go
result, err := agent.Harness(ctx, task, schema, &dest, harness.Options{})
```

Previously each of those raised *"No harness provider specified"*. Picking Claude Code,
Codex, Gemini CLI, or OpenCode is now an **override** rather than a prerequisite — same
loop code, different worker.

## Why

The harness was the one AgentField primitive that could not be used out of the box: every
call needed a provider name plus a third-party CLI installed and authenticated first.
AForge ships alongside `af`, so the default path costs the user one env var
(`OPENROUTER_API_KEY`) and nothing else. Orchestrating someone else's harness stays a
first-class, documented story — it just stops being the price of admission.

The `model` default moved for the same reason. It was `"sonnet"`, a Claude-specific id
that was silently wrong for every other provider. It is now empty, meaning "the provider's
own default"; `claude-code` carries `sonnet` internally so explicit claude-code users see
no change.

### Provider precedence

| Order | Source |
| --- | --- |
| 1 | Explicit value — `.harness(provider=…)`, `HarnessConfig(provider=…)`, `Options.Provider`, `{ provider }` |
| 2 | `AGENTFIELD_HARNESS_PROVIDER` env var |
| 3 | Default — `aforge` |

Implemented once per SDK and applied at both the config-materialisation and the
runner-resolution points, so a runner built with no config at all follows the same chain.

## Validation contract

Behaviours this PR must exhibit, in caller-observable terms:

1. A harness call with no provider configured anywhere runs `aforge` — it does not raise.
2. With `AGENTFIELD_HARNESS_PROVIDER=<name>` set and no explicit provider, that provider runs.
3. An explicit provider always wins over the env var and over an agent-level default.
4. Blank / whitespace-only provider values (explicit or env) are treated as unset and fall
   through to the next source.
5. A genuinely unknown provider name still fails loudly, with the supported list in the message.
6. `HarnessConfig` built with no arguments is a complete, valid config in all three SDKs.
7. An unset model means the provider's own default. `claude-code` with no model still sends
[...]

PR #913, opened 2026-08-12
Add Pi and OMP harness providers

santoshkumarradha:
## Takeover note (2026-08-27)

This PR was taken over and reworked. **aforge stays the default harness provider** — that was the maintainer decision, and this branch no longer changes it. The branch was merged with `main` (no rebase) and the OMP-default logic was removed: Pi and OMP are now **additional** providers you select explicitly or via `AGENTFIELD_HARNESS_PROVIDER`, nothing more. Provider-less calls resolve exactly as they do on `main`.

The four review findings from the last review round are fixed:

| Finding | Fix |
| --- | --- |
| TS pi/omp results carried no failure classification (`failureType` / `returnCode`), unlike every other TS provider | `3ae5157d` |
| `projectDir` was only honoured by pi/omp, so the schema-output protocol disagreed with the other TS providers | `ba76a4c5` |
| `af doctor --probe` passed the prompt positionally instead of over stdin the way the pi/omp adapters do | `944b388a` |
| Plan mode narrowed the tool list but sent no approval flag, leaving the run on the CLI's default (possibly interactive) behaviour | `bdbaf471` |

Follow-up commits on top of those keep `af doctor` surveying aforge itself, stop `--probe` reporting a silently broken pi/omp install as ok, and drop the stale "OMP default" wording from docs and tests.

---

## Product outcome

AgentField can now run Pi and OMP as first-class coding harnesses from Python, TypeScript, and Go. aforge remains the zero-configuration default; Pi and OMP are selected explicitly per call, through `HarnessConfig`, or via `AGENTFIELD_HARNESS_PROVIDER`. A production-shaped Go example fans one task out to Pi and OMP concurrently and joins their typed results in a single AgentField workflow.

This closes the practical parity gap with OpenCode: applications use the same AgentField harness contract for model selection, project root, system prompt, tools, permissions, session resume, structured output, retries/timeouts, and normalized usage metrics.

## Developer experience

- Provider resolution is identical in every SDK and unchanged from `main`: per-call override → agent harness config → `AGENTFIELD_HARNESS_PROVIDER` → aforge.
- Model resolution is per-call override → agent config → the selected CLI's own configured default.
- Explicit calls such as `provider="opencode"` or `provider="pi"` are unchanged.
- Usage attribution records the resolved provider, so Pi and OMP runs report `harness=pi` / `harness=omp` across Python, TypeScript, and Go.
- Missing Pi/OMP executables produce actionable errors with the exact upstream install command.
- `af harness doctor` recognizes both CLIs, reports binary/version/auth/usability, and lists aforge first as the default.
- The README, provider guide, architecture design, shipped AgentField skill, and Go example document the providers and their operational lifecycle.

## Architecture

- Adds a shared Pi-family adapter in each SDK with thin Pi/OMP flavor mapping instead of duplicating orchestration logic.
- Dispatches one-shot prompts through stdin and consumes JSON event streams, avoiding interactive CLI behavior.
- Maps AgentField's provider-neutral options to native CLI behavior:
  - model + `#variant` → `--model` + `--thinking`
  - project root → process cwd for Pi; `--cwd` plus process cwd for OMP
  - plan/auto → read-only tool filtering; only OMP takes a native approval flag (`--auto-approve`), since Pi has none
  - resume → Pi `--session`; OMP `--resume`
  - tool names → Pi `find`; OMP `glob`
- Reuses the existing isolated schema-file protocol and retry/validation orchestration used by the other harnesses.
- Normalizes sessions, turns, tokens, cache tokens, cost, model, duration, messages, return code, and failure classification into the existing result types.
- Keeps runtime installation outside SDK execution. Like Codex, Gemini, and OpenCode, operators install/pin the CLI in the host or image; AgentField only discovers, verifies, and invokes it. No application call mutates production hosts.

## Parity assessment

Pi and OMP have feature parity with OpenCode at the AgentField contract boundary. Native flags are intentionally not identical: each adapter translates its CLI's approval, resume, project-root, and filesystem-search vocabulary. Those differences do not leak into application code.

Covered contract surface:

- model and reasoning variant
- project/cwd isolation and environment overrides
- system prompt and normalized tool allowlist
- plan and auto permission modes
- resumable sessions
- structured schema output with validation/recovery
- timeout and retry controls
- normalized token, cache, cost, session, turn, model, and duration metrics
- binary overrides and actionable availability failures

## Installation and operations

```bash
npm install -g --ignore-scripts @earendil-works/pi-coding-agent
curl -fsSL https://omp.sh/install | sh
af harness doctor --provider pi,omp --json
```

The doctor check is static and does not make a paid model request. Containers should install the selected CLI during image construction and run the doctor as a startup/CI gate.

## Go workflow example

`examples/go_agent_nodes/cmd/harness_duo` registers:

```text
compare
├── pi_worker   (explicit Pi)
└── omp_worker  (explicit OMP)
```

`compare` starts both branches concurrently and joins their structured results. It defaults to `openrouter/minimax/minimax-m2.7` and accepts `openrouter/google/gemini-2.5-flash` for a faster smoke path.

## Verification

- Python: focused harness/types/usage suite passed; Ruff passed.
- TypeScript: 81 files / 832 tests passed; typecheck and production build passed.
- Go SDK: harness suite passed; changed agent/default/usage tests passed; example build passed.
- Control plane: CLI and embedded-skill tests passed; embedded skills are in sync.
- CI was green on the pre-takeover head: 41 active checks passed across Go, Python, TypeScript, control-plane, Docker, local/Postgres functional tests, performance, aggregate/patch coverage, CodeQL, README links, and CLA (with only the expected AI-label job skipped).
- Live doctor: Pi `0.84.1` and OMP `17.2.15`, both authenticated through OpenRouter and usable.
- Live AgentField execution: `run_20260812_164308_d2tasl3y` / `exec_20260812_164308_fp15z9if` succeeded in 19.4s with `openrouter/google/gemini-2.5-flash`; `compare`, `pi_worker`, and `omp_worker` all completed with structured output and native cost/token metrics.
- Earlier MiniMax validation also succeeded for both branches: `run_20260812_162041_fw7j8b54` with `openrouter/minimax/minimax-m2.7`.

## Compatibility

No default-provider or model-default change: provider-less calls resolve exactly as on main (explicit → AGENTFIELD_HARNESS_PROVIDER → aforge).

Explicit provider configurations are otherwise unchanged.


santoshkumarradha:
Addressing the review findings on this PR.

**Code changes** (commit `973913eb`):
- **Go `tools[:0]` in-place filter** — `sdk/go/harness/pi.go` plan-mode branch now builds the read-only tool list as a fresh slice (`make([]string, 0, len(tools))`) instead of aliasing the `tools` backing array. This matches the Python and TS providers, which both build a fresh list, and removes the footgun if `normalizePiTools` is ever changed to return a caller-owned slice.
- **Go negative return code** — `sdk/go/harness/pi.go` now distinguishes a signal kill (`ReturnCode < 0`) from a plain non-zero exit, reporting `Process killed by signal N.` to match the Python provider and the existing `gemini.go` / `opencode.go` Go providers. Previously a SIGKILL was folded into the generic non-zero branch with a misleading "exited with code -9" message. Added a `pi_test.go` case pinning the signal-kill message.

**Release-note callouts** (PR body, "Compatibility" section):
- Called out the **provider→OMP default change**: provider-less harness calls now run OMP (a live model call) instead of raising `No harness provider specified`. Callers who relied on that error to catch a misconfiguration are affected.
- Called out the **Python `HarnessConfig.model` `sonnet`→`None` default change**: Python callers who omitted `model` previously got the Claude-specific `sonnet` fallback; they now get the selected CLI's configured default. Go/TS had no model default and are unaffected.

**Not changed** (reviewed, left as-is to keep the diff minimal):
- The TS factory's unreachable `throw` (nit; dead code, harmless).
- The `harness_duo` early-return leaving a branch in flight (nit; buffered channel, no leak, demo example).
- The mocked-CLI flag-coverage gap (nit; consistent with the other CLI providers in this repo).

**Verification:** `go build ./...`, `go vet ./harness/`, and the full `sdk/go` test suite pass after the changes.

AbirAbbas (commented):
Heads up that this can't merge as-is any more: main shipped `aforge` as the harness default (#905, #927-#929) along with an `AGENTFIELD_HARNESS_PROVIDER` env tier and a provisioning path (`af aforge ensure`, curl installer, docker images), so the 25 conflicts here are a genuine either/or on "what is the default" rather than anything mechanical — I didn't push a merge because resolving them means picking a winner.

Two calls worth making before we resolve: (a) whether the zero-config default is `aforge` or `omp`, and separately whether a provider-less call should silently start a paid run at all versus keeping the loud error this PR removes; (b) the Python `HarnessConfig` change, where `provider` goes from required to defaulted and `model` goes from `str = "sonnet"` to `Optional[str] = None` — that second one is a real break for anyone reading `config.model` as a string.

Pi and OMP as *additional* providers look good to me and I'd take them regardless; the default question is the only thing actually blocking. Four code notes below. Also note the earlier round's flag on `harness_duo`'s early return isn't a bug — the channel is buffered to 2, so nothing leaks.

sdk/typescript/src/harness/providers/pi.ts
```diff
@@ -0,0 +1,232 @@
...
+        cwd: root,
+        inputText: prompt,
+        timeout: typeof options.timeout === 'number' ? options.timeout * 1000 : undefined,
+      });
+      const events = parseJsonl(stdout);
+      const parsed = parsePiEvents(events, model);
+      parsed.metrics.durationApiMs = Date.now() - startApi;
+
+      let errorMessage: string | undefined;
+      if (exitCode !== 0) {
```
AbirAbbas: This maps exit codes more thinly than the Go and Python adapters in the same PR, and the gap gets sharper once main is merged in.

main added `failureType` and `returnCode` to the TS `RawResult`, and the runner propagates both (`runner.ts:83`, `:194`); `aforge.ts` and `opencode.ts` set them. `createRawResult` here sets neither, so pi/omp end up the only TS providers with no failure classification at all — while `pi.go` and `pi.py` in this same PR do produce `FailureCrash` / `FailureTimeout` / `FailureAPIError` / `FailureNoOutput` plus the return code. Retry and reporting logic downstream can't tell an OMP timeout from an OMP crash on TypeScript.

Separate but related: `runCli` resolves `exitCode: code ?? 0`, so a signal-killed child (OOM killer, supervisor SIGTERM) surfaces as exit 0 and lands in the `!parsed.result` branch — the caller is told "omp exited successfully without an assistant response" for a process that was killed. Go and Python both special-case `returncode < 0` into "Process killed by signal N". That collapse is pre-existing in `cli.ts` (opencode's `exitCode < 0` branch on main is dead code for the same reason), so it's not yours to fix here, but it does mean the honest signal path doesn't exist on TS at all.

sdk/typescript/src/harness/runner.ts
```diff
@@ -53,11 +55,9 @@ export class HarnessRunner {
     const { schema, ...rest } = options;
     const resolved = this.resolveOptions(this.config, rest);
 
-    if (!resolved.provider) {
-      throw new Error("No harness provider specified. Set 'provider' in HarnessConfig or pass it to .harness() call.");
-    }
+    resolved.provider ??= DEFAULT_HARNESS_PROVIDER;
 
-    const cwd = resolved.cwd ?? '.';
+    const cwd = resolved.projectDir ?? resolved.cwd ?? '.';
```
AbirAbbas: `projectDir` is new in this PR (`types.ts:21`, `:48`), and this line makes it the base for the whole schema-output protocol — `buildPromptSuffix(schema, cwd)`, `handleSchemaOutput(..., cwd, ...)` and `cleanupTempFiles(cwd)` all follow it.

But of the TS providers only pi/omp actually read it. `codex.ts:17/42`, `gemini.ts:17/35`, `opencode.ts:27` and `claude.ts:47` all still key off `options.cwd` alone. So `app.harness(task, { schema, projectDir: '/repo', provider: 'codex' })` instructs the model to write `/repo/.agentfield_output.json` while codex actually runs in the agent process's cwd, and the read-back fails as a schema error with nothing pointing at the cause.

Python doesn't have this split — `codex.py:36`, `gemini.py:59` and `aforge.py:95` all do `options.get("project_dir") or options.get("cwd")`, which is why `_runner.py:296` can safely base the output path on `project_dir`. Either give the other TS providers the same `projectDir ?? cwd` fallback, or leave the schema path on `cwd` until they have it.

control-plane/internal/cli/doctor.go
```diff
@@ -85,10 +85,12 @@ var harnessProviders = []struct {
 	Binary    string   // executable name to look up on PATH
 	ProbeArgs []string // minimal one-shot invocation used by `--probe`
 }{
+	{Name: "omp", Binary: "omp", ProbeArgs: []string{"--print", "Say OK"}},
```
AbirAbbas: The probe passes the prompt positionally (`omp --print "Say OK"`), but all three adapters this PR adds invoke the CLI as `<bin> --print --mode json` and feed the prompt over **stdin** (`pi.go:67`/`:137`, `pi.py:149`/`:213`, `pi.ts:128`/`:181`). One of the two has the CLI's surface wrong.

If `--print` takes its prompt from stdin, this probe hands OMP a positional it ignores plus an immediately-closed stdin, and `af doctor --probe` classifies a perfectly healthy OMP install as `empty` or `error` — on the provider this PR makes the default. Worth making the probe match whatever the adapters do.

Same applies to the `pi` entry on line 93.

sdk/go/harness/pi.go
```diff
@@ -0,0 +1,316 @@
...
+	if options.PermissionMode == "auto" {
+		permissionFlag := "--approve"
+		if p.flavor == piFlavorOMP {
+			permissionFlag = "--auto-approve"
+		}
+		cmd = append(cmd, permissionFlag)
+	}
+
+	tools := normalizePiTools(options.Tools, p.flavor)
+	if options.PermissionMode == "plan" {
```
AbirAbbas, line 104: `plan` narrows the tool list but never sends an approval flag, whereas `auto` sends `--approve` / `--auto-approve`. That leaves a plan-mode run on the CLI's *default* approval behaviour, with stdin already consumed by the prompt and closed — so if that default is "ask the user", there's nothing left to read from.

The thing I can't tell from here is whether `--tools` is a hard allowlist the CLI enforces or just a hint. If it's a hint, then `permission_mode="plan"` doesn't actually make the run read-only in any of the three SDKs (same shape at `pi.py:180` and `pi.ts:161`) — which matters, because plan is exactly the mode callers pick when they don't want writes. `codex.go:116` gets this from `--sandbox <mode>`; is there an equivalent approval/read-only flag on pi and omp to pair with the tool filter?

sdk/typescript/src/harness/providers/pi.ts
AbirAbbas: Fixed in 3ae5157d — pi/omp now set failureType and returnCode using the same ladder as the Go and Python adapters (signal death and non-zero exit are crash, a stopReason error/aborted on a clean exit is api_error, a clean exit with no assistant text is no_output), and runCli now surfaces the signal instead of collapsing it to exit 0.

sdk/typescript/src/harness/runner.ts
AbirAbbas: Fixed in ba76a4c5 — added a resolveRoot() helper (projectDir → project_dir → cwd) and routed all six TS providers through it, so codex, gemini, opencode and claude no longer run in a different directory than the one the schema-output protocol writes into.

control-plane/internal/cli/doctor.go
AbirAbbas: Fixed in 944b388a — the probe entry gained ProbeStdin, so pi and omp are now probed as `<bin> --print --mode json` with the prompt on stdin exactly like the adapters, while positional-prompt providers keep a nil stdin.

sdk/go/harness/pi.go
AbirAbbas: You were right that something was off, though it resolved the other way: I checked against the real CLIs and Pi has no approval flag at all (--approve/--auto-approve/--yolo/--approval-mode all die with "Unknown option"), so bdbaf471 drops the flag from Pi's auto branch and leaves plan mode on the enforced --tools allowlist, which is Pi's own documented read-only mechanism and which OMP auto-approves regardless of approvalMode.

AbirAbbas (approved):
Merged this with main instead of rebasing and kept aforge as the default harness provider — Pi and OMP land as additional providers you pick explicitly or through AGENTFIELD_HARNESS_PROVIDER, so provider-less calls resolve exactly as they do on main. The four findings from my last round are fixed: TS pi/omp results now carry failureType and returnCode, all six TS providers resolve their working directory through one projectDir ladder, `af doctor --probe` drives pi/omp over stdin the way the adapters do, and Pi's auto-mode approval flag is gone because the real CLI rejects every variant of it. Verified with the per-surface CI-exact gates, a defaults proof in all three SDKs, and real-binary doctor and error-path runs.

PR #914, opened 2026-08-12
feat(sdk/go): dispatch envelope unwrap + Context injection (#514)

### LINKED ISSUE #514: [Go SDK] Dispatch envelope unwrap + Context injection

7vignesh:
## Summary

Wires the dispatch side of the trigger system into the Go SDK: the control plane's `{event, _meta}` envelope is detected and peeled, the matching binding's `Transform` runs, and the handler can read the trigger metadata via `triggers.FromContext(ctx)`. Direct calls are untouched. This is slice 2 of the Go trigger parity epic (#508).

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs only
- [ ] Tests only
- [ ] CI / tooling
- [ ] Breaking change

## What changed

New `sdk/go/triggers/dispatch.go`:

| Function | Purpose |
|---|---|
| `IsEnvelope(body)` | Detects the dispatcher shape. Requires `_meta.trigger_id`, so an ordinary payload that happens to carry an `event` key is never misread |
| `Unwrap(body)` | Peels the event and builds a `*Context` from `_meta`, with multi-layout `received_at` parsing and a `now()` fallback |
| `NewContext` / `FromContext` | Carry the trigger metadata through `context.Context` |
| `ApplyTransform` | Picks the best-matching binding and runs its `Transform` |

Wired into every dispatch path through a single `applyTriggerDispatch` helper in `agent_triggers.go`, so the paths cannot drift: `handleReasoner` (sync HTTP), `executeReasonerAsync` (async goroutine), `handleExecute` (`/execute`), `Execute` (local), and `HandleServerlessEvent` (serverless). `CallLocal` is deliberately excluded, since agent-to-agent calls are direct invocations by definition.

Also drops the `EXPERIMENTAL` caveats on `Context`, `Transform`, and `EventOpts.Transform`, which pointed at this issue as the thing that would make them real. Flagged by @AbirAbbas on #906.

### One deviation from the issue spec, worth a look

The issue sketches a three-argument handler:

```go
func handlePayment(ctx context.Context, in Payment, tc *triggers.Context) (any, error)
```

Go's `HandlerFunc` is `func(ctx context.Context, input map[string]any) (any, error)`, so adding a third parameter is a breaking change for every existing reasoner and every `RegisterReasoner` call site. I used context propagation instead:

```go
func handlePayment(ctx context.Context, input map[string]any) (any, error) {
    if tc := triggers.FromContext(ctx); tc != nil {
        // dispatched via a trigger
    }
    return nil, nil
}
```

Same capability, no break, and it matches how `ExecutionContext` and the cost tracker already flow through `ctx` in this SDK. Happy to switch to a typed generic wrapper if you would rather have the signature in the spec, but that reads like its own slice.

Transform matching mirrors Python and TypeScript exactly: source must match, a binding declaring `EventTypes` requires an exact or dotted-prefix event-type match, a specific binding beats a catch-all, and a binding without a `Transform` passes the raw event through. A panicking `Transform` is recovered to raw input, so a buggy transform degrades instead of failing the dispatch.

## Test plan

- [x] `cd sdk/go && go test ./triggers/ -v` (17 dispatch tests: envelope detection table, unwrap, timestamp fallbacks, context round-trip, transform matching and specificity, prefix match, panic recovery, nil cases)
- [x] `cd sdk/go && go test ./agent/ -run TestDispatch_ -v` (6 end-to-end tests driving both shapes through the real `Execute` path)
- [x] `cd sdk/go && go test -race ./triggers/` clean
- [x] `cd sdk/go && go build ./...` and `go vet ./...` clean
- [x] `gofmt` clean on every file this PR touches (`agent/agent.go` was already gofmt-dirty on main and is left alone)

The one failing test in the `agent` package, `TestWithEventTrigger_StampsCallerFileAndLine`, is pre-existing and Windows-only: it splits `code_origin` on `:` and the `D:` drive letter yields three parts instead of two. Unrelated to this change and passes on CI.

## Test coverage

Both contracts are pinned by tests that fail if the implementation is reverted: the direct-call test asserts `FromContext` is nil and that `Transform` did not run, and the envelope test asserts the peeled and transformed payload plus every `_meta` field.

## Checklist

## Related issues / PRs

Closes #514
Part of epic #508
Builds on #906 (slice 1, merged)
Mirrors the TypeScript equivalent in #796


santoshkumarradha (approved):
Validated the dispatch path locally and spot-checked the envelope unwrap and transform handling. This looks good to me.

Issue #516, opened 2026-04-28
[Go SDK] examples/triggers-demo-go end-to-end demo + skill docs


santoshkumarradha:
## Summary

Land a Go counterpart to `examples/triggers-demo/` and add a Go-flavored section to the skill reference so the `agentfield-multi-reasoner-builder` skill emits correct Go scaffolds.

Part of the epic #508. Depends on the public surface and dispatch path from the sibling issues.

### New Files

| File | Purpose |
|------|---------|
| `examples/triggers-demo-go/main.go` | Three deterministic reasoners — `handle_payment` (Stripe), `handle_pr` (GitHub), `handle_tick` (cron). Same shape and memory-write semantics as `examples/triggers-demo/agent.py` |
| `examples/triggers-demo-go/Dockerfile` | Build the Go agent into a container |
| `examples/triggers-demo-go/docker-compose.yml` | Control plane + Go agent + shared demo secrets — drop-in replacement for the Python demo |
| `examples/triggers-demo-go/README.md` | Quick-start + UI tour |
| `examples/triggers-demo-go/scripts/fire-events.sh` | Symlink or copy of `examples/triggers-demo/scripts/fire-events.sh` — same script, same signed payloads |

### Modified Files

| File | Change |
|------|--------|
| `skills/agentfield-multi-reasoner-builder/references/triggers.md` | Add a Go-flavored section showing the same trigger declarations in Go so the skill produces correct Go scaffolds |

## Acceptance

- [ ] `cd examples/triggers-demo-go && docker compose up` brings up control-plane + go-agent
- [ ] `./scripts/fire-events.sh` (the existing script, unmodified) fires Stripe + GitHub events that dispatch successfully
- [ ] Cron trigger fires every minute and produces memory writes equivalent to the Python demo
- [ ] Trigger sheet at `localhost:8080/ui/triggers` shows three triggers, all enabled, with `signature: valid` on dispatched events
- [ ] Run-detail Inbound section populates for runs originating from the Go agent

## References

- Parent epic: #508
- Python demo: `examples/triggers-demo/`
- Skill reference: `skills/agentfield-multi-reasoner-builder/references/triggers.md`
- Trigger system: #506

PR #915, opened 2026-08-12
feat(sdk/go): trigger test helpers + fixture library (#515)

### LINKED ISSUE #515: [Go SDK] Test helpers + fixture library
## Summary

Add Go test helpers and a captured-fixture library so reasoners can be unit-tested without spinning up a control plane, mirroring the Python `agentfield/testing.py` surface.

Part of the epic #508.

### New Files

| File | Purpose |
|------|---------|
| `sdk/go/triggers/testing.go` | `SimulateEvent(t, handler, fixture)` and `SimulateSchedule(t, handler, opts)` — build a `*triggers.Context` from a fixture and invoke the handler in the same shape the dispatch path uses |
| `sdk/go/triggers/testdata/*.json` | Captured payload library — copied from `sdk/python/agentfield/fixtures/triggers/` so behavioural parity is provable against the same JSON files |
| `sdk/go/triggers/load_fixture.go` | `LoadFixture(t, name)` — small helper that opens `testdata/name` and unmarshals into the dispatcher envelope shape |

## Interface

```go
import "github.com/Agent-Field/agentfield/sdk/go/triggers"

func TestHandlePayment(t *testing.T) {
    fixture := triggers.LoadFixture(t, "stripe.payment_intent.succeeded.json")
    result, err := triggers.SimulateEvent(t, handlePayment, fixture)
    if err != nil { t.Fatal(err) }
    if result.(map[string]any)["kind"] != "payment" {
        t.Fatalf("expected payment kind, got %v", result)
    }
}
```

## Acceptance

- [ ] `go test -race ./sdk/go/triggers/...` passes
- [ ] `SimulateEvent` runs the handler against each of the six fixture files (Stripe, GitHub, Slack, generic_hmac, generic_bearer, cron) and produces the expected `*triggers.Context` shape
- [ ] Fixture JSON files are byte-identical to the Python SDK's copies (verified by a small helper test)

## References

- Parent epic: #508
- Python equivalent: `sdk/python/agentfield/testing.py` and the fixtures under `sdk/python/agentfield/fixtures/triggers/`
- Trigger system: #506

7vignesh:
## Summary

Adds Go test helpers and a captured-fixture library so reasoners can be unit-tested without a control plane, HTTP server, or real provider, mirroring the Python `agentfield/testing.py` surface. This is slice 3 of the Go trigger parity epic (#508).

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs only
- [ ] Tests only
- [ ] CI / tooling
- [ ] Breaking change

## What changed

New `sdk/go/triggers/testing.go`:

| Function | Purpose |
|---|---|
| `SimulateEvent(t, handler, opts)` | Builds the `*Context` the runtime would have produced, applies the matching binding's `Transform`, and invokes the handler with the context attached |
| `SimulateSchedule(t, handler, opts)` | Wraps `SimulateEvent` for cron handlers (source `cron`, event type `tick`) |
| `SimulatedContextFrom(ctx)` | Retrieves the synthetic context inside a handler under test |

Identifiers default to fresh random values so repeated simulations are independently dedup-safe, and every field is overridable for deterministic assertions.

New `sdk/go/triggers/load_fixture.go`:

| Function | Purpose |
|---|---|
| `LoadFixture(t, name)` | Reads a captured payload, accepting either `"stripe"` or `"stripe.json"` |
| `RawFixture(t, name)` | Returns undecoded bytes, for byte-level assertions |
| `FixtureNames()` | Lists all six sources, for table-driven tests |

New `sdk/go/triggers/testdata/*.json`: six fixtures (stripe, github, slack, cron, generic_hmac, generic_bearer) copied byte-for-byte from `sdk/python/agentfield/fixtures/triggers/`.

### Two deviations from the issue spec

**Fixtures are embedded with `go:embed` rather than read from a relative `testdata` path.** The spec's `LoadFixture(t, name)` opening `testdata/name` works when the SDK's own tests run, but silently fails the moment a *user* calls it from their package: the working directory is theirs, not the SDK's, and from the module cache the path does not resolve at all. Embedding makes the fixtures available regardless of caller location.

**The helpers do not import `Unwrap`/`ApplyTransform` from #514.** #515 is declared `blocked-by: #513` only, so I kept this branch independent of #514 and mergeable in either order. The cost is a duplicated transform matcher, `applyMatchingTransform`, which is byte-for-byte the same matching rules as the dispatch-side implementation. I would rather not leave that standing: once both slices land, the helper should call the shared `ApplyTransform` and delete the copy. I have flagged it in a code comment and would propose folding the collapse into #516. Say the word if you would prefer I make #515 depend on #514 instead and drop the duplication now.

## Test plan

- [x] `cd sdk/go && go test -race ./triggers/` passes (61 test cases)
- [x] `SimulateEvent` drives the handler against all six fixtures and produces the expected `*Context` shape, covered by a table test over `FixtureNames()`
- [x] Per-provider assertions on known fixture field values (stripe amount, github PR number, slack event type, cron expression, hmac order id, bearer notification id)
- [x] Fixtures verified byte-identical to the Python SDK copies by `TestFixturesAreByteIdenticalToPythonSDK`, which runs (not skips) in this checkout and names the exact `cp` command to re-sync on failure
- [x] `TestFixtureNamesCoversAllEmbeddedFiles` fails if a fixture is added without updating `FixtureNames()`
- [x] `cd sdk/go && go build ./...` and `go vet ./...` clean
- [x] `gofmt` clean on every file this PR touches

Also covered: transform application and source-mismatch skip, identifier overrides, uniqueness across calls, parent-context propagation, handler error propagation, non-object transform results, panicking transforms, and the schedule helper's body and timestamp handling.

## Test coverage

## Checklist

## Related issues / PRs

Closes #515
Part of epic #508
Builds on #906 (slice 1, merged)
Independent of #914 (slice 2), mergeable in either order
Mirrors the TypeScript equivalent in #796


santoshkumarradha (approved):
Ran the new trigger test-helper surface locally and the fixture embedding approach makes sense here. This looks good to me.

PR #916, opened 2026-08-13
feat(sdk/go): add Agent.Span for traced in-process sub-executions

AbirAbbas:
## Summary

Adds `Agent.Span(ctx, name, input, fn)` — the in-process tracing primitive: it runs `fn` as a child execution of the context carried by `ctx` and emits workflow events to the control plane, so the call appears as a node in the run's DAG. It is `CallLocal`'s lineage + event emission without requiring the target to be a registered reasoner.

Motivation: agents that implement pipeline stages as plain function calls render as a single long-running DAG node no matter how much internal structure a run has. The Python SDK's `@app.reasoner` decorator tracking reports every internal stage as a child node; Go had no equivalent for unregistered functions. First consumer: the deep-research agent, whose 12-minute `prepare_research_package` runs currently trace as one node.

## Design

- Child lineage comes from `buildChildContext` (same as `CallLocal`): fresh `execution_id`, `parent_execution_id` from ctx, `run_id` preserved as the DAG grouping key. The child context is injected into `fn`'s ctx, so nested `Span`/`CallLocal`/`Note` calls chain to arbitrary depth and concurrent Spans branched from one ctx become siblings.
- **Terminal events are synchronous** — a completed span can never be left dangling in `running` state (the DAG's overall status derivation treats any non-terminal node as still-running, so a lost terminal event would make finished runs look hung).
- **Start events are async** through a bounded queue (256) and are shed first under load. Safe because `POST /api/v1/workflow/executions/events` upserts: the terminal event alone creates the node.
- `fn` panics emit a `failed` terminal event and re-raise.
- Input/result payloads are size-capped agent-side (16KB, 4KB preview marker) since the events path stores payloads verbatim with no blob offloading.

## Tests

8 new tests in `span_test.go` mirroring the `TestCallLocal*` harness: lineage on events, nested parent chains, parallel siblings, error + panic terminal events, payload truncation, no-control-plane no-op, rootless span minting its own run. Full `sdk/go` suite + `go vet` green.

---

### DIFF
--- a/control-plane/internal/handlers/coverage_handlers_90_test.go
+++ b/control-plane/internal/handlers/coverage_handlers_90_test.go
@@ -158,10 +158,15 @@ func TestWorkflowExecutionEventHandler_ErrorBranches(t *testing.T) {
 		router.ServeHTTP(rec, req)
 
 		require.Equal(t, http.StatusInternalServerError, rec.Code)
-		require.Contains(t, rec.Body.String(), "failed to create execution")
+		require.Contains(t, rec.Body.String(), "failed to persist execution event")
 	})
 
-	t.Run("update existing nil execution creates replacement", func(t *testing.T) {
+	// A store that perpetually reports a row on read but has nothing to
+	// update is inconsistent; the handler must exhaust its bounded retries
+	// and fail rather than loop or fabricate success. The realistic
+	// transient version of this (row vanishes once, then re-creates) is
+	// covered by TestWorkflowExecutionEventHandler_RowVanishedBetweenReadAndUpdate.
+	t.Run("perpetually vanishing execution exhausts retries", func(t *testing.T) {
 		store := &updateNilExecutionStore{}
 		router := gin.New()
 		router.POST("/events", WorkflowExecutionEventHandler(store))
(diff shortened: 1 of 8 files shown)


AbirAbbas:
Manual verification against a live control plane surfaced two defects in the original head (`faa8b13a`); both are fixed and test-pinned by the two new commits:

**`ff1c7eb9` — fix(sdk/go): snapshot span input before handing it to the async event queue.** `truncateTraceInput` returned the caller's map by identity and the async start-event goroutine `json.Marshal`ed it at an unbounded delay — any caller mutating the input map after `Span` entry (including inside `fn`) was a process-fatal `concurrent map read and map write`. Reproduced under `go test -race`; the input is now deep-copied at `Span` entry, so events carry an as-of-call snapshot. New test: `TestSpanInputSnapshotIsolatedFromCallerMutation` (doubles as a `-race` canary).

**`99265041` — fix(control-plane): survive insert races and shed starts on the events endpoint.** The endpoint's Get→Create is not atomic; a span's async "running" event races its own synchronous terminal event for any sub-ms span, the losing INSERT 500s, and a lost terminal event stranded the node in `running` forever (live repro: 550-span flood → 11/551 stranded, run pinned in the in-flight list, later reaped to `timeout` so a finished run rendered as timed out). The handler now retries in a bounded loop, merging insert losers through the update path; terminal-state immutability unchanged. Also backdates `StartedAt` by `duration_ms` for nodes created from a terminal event alone, so shed start events no longer produce zero-width timeline bars. New tests cover the race-loser merge, vanished-row re-create, retry exhaustion, and backdating.

Post-fix verification: the same 550-span flood persists **551/551 nodes terminal, zero dropped events**, and the in-flight list drains; full `sdk/go` suite green under `-race`; full control-plane suite + vet green; changed files `golangci-lint`-clean.

PR #917, opened 2026-08-13
feat(examples): triggers-demo-go end-to-end demo + Go skill docs (#516)

### LINKED ISSUE #516: [Go SDK] examples/triggers-demo-go end-to-end demo + skill docs

7vignesh:
## Summary

Lands a Go counterpart to `examples/triggers-demo/` (Python) and `examples/triggers-demo-ts/` (TypeScript), driven by the same unmodified `scripts/fire-events.sh`, plus a Go section in the skill reference so the `agentfield-multi-reasoner-builder` skill emits correct Go scaffolds. This is the final slice of epic #508.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs only
- [ ] Tests only
- [ ] CI / tooling
- [ ] Breaking change

## What changed

New `examples/triggers-demo-go/`:

| File | Purpose |
|---|---|
| `main.go` | Three deterministic reasoners: `handle_payment` (Stripe, with a `Transform`), `handle_pr` (GitHub), `handle_tick` (cron) |
| `main_test.go` | 24 tests exercising the reasoners through the slice-3 helpers, doubling as a worked example |
| `Dockerfile` | Multi-stage build against the in-tree SDK, CGO off, alpine runtime |
| `docker-compose.yml` | Control plane plus Go agent, shared demo secrets, DID enabled |
| `scripts/fire-events.sh` | Byte-identical copy of the Python demo's script |
| `README.md` | Quick start, UI tour, memory-key table |

Memory keys and record shapes match the Python demo exactly, so the UI surfaces read the same data regardless of which demo runs.

`skills/.../references/triggers.md` gains a Go section covering both declaration forms, `Context` fields, the `FromContext` accessor, the test helpers, and a three-way comparison table. Every Go snippet in it was compiled against the real SDK rather than eyeballed. The shared wire-format and envelope sections moved from "both SDKs" to "all three".

### A defect this slice surfaced, now fixed

Writing the demo exposed a real bug in the interaction between slices 2 and 3. `SimulateEvent` attached the trigger context under a private key while handlers read it via `triggers.FromContext`, so a handler that worked in production saw a **nil context under test**. The demo's first test run failed with `received_via = direct_call, want stripe`.

This is the duplication I flagged in #915. Now that both slices are on main, the fix belongs here: `SimulateEvent` uses the shared `ApplyTransform` and `NewContext` from `dispatch.go`, and the 72-line duplicated matcher is deleted (net 61 lines removed from `testing.go`). Added a regression test asserting `FromContext` sees what `SimulateEvent` attaches, so the simulated and live paths cannot drift again.

### One structural choice

The reasoners are declared as `eventReasoner` / `scheduleReasoner` values that `main()` iterates, rather than inline inside `OnEvent` calls. The SDK exposes no public accessor for registered reasoners, so inline declarations would have been untestable without reaching into agent internals. This keeps the handlers and trigger wiring assertable from the demo module.

## Test plan

- [x] `cd examples/triggers-demo-go && go test ./...` (24 tests) and `go test -race ./...` clean
- [x] `cd sdk/go && go test -race ./triggers/` clean (86 tests, including the new drift guard)
- [x] `go build`, `go vet`, `gofmt` clean on both modules
- [x] `docker compose config` validates
- [x] Dockerfile's exact steps reproduced in a clean tree with `-mod=readonly`, confirming `go.sum` completeness
- [x] The real compiled binary registers all three triggers against a stub control plane with the expected wire payload: correct sources, event types, secret env vars, cron `expression`/`timezone` config, `accepts_webhook` auto-set to true, and `code_origin` stamped for the drift card

Not verified: the `docker compose up` image build itself, since no Docker daemon was available in this environment. The build steps and every `COPY` path were validated individually as above, but a reviewer with Docker should confirm the end-to-end `up` plus `fire-events.sh` run.

Known scope note, documented in the README: `fire-events.sh` is shared verbatim with the Python demo, and its later sections create Slack, `generic_hmac`, and `generic_bearer` triggers routed at a `handle_inbound` catch-all. The Python demo defines that reasoner; this demo and the TypeScript one do not, matching the three-reasoner scope. Those three dispatches fail while Stripe, GitHub, and cron work end to end.

## Test coverage

The deleted matcher was dead duplication replaced by the already-tested shared implementation, so no coverage is lost.

## Checklist

## Related issues / PRs

Closes #516
Completes epic #508 (Go SDK trigger system parity)
Builds on #906 (slice 1), #914 (slice 2), #915 (slice 3)
Mirrors the TypeScript demo in #796


santoshkumarradha (changes requested):
Thanks for putting this together. The Go demo and the trigger-testing helpers look solid overall, but I do not think we should commit `examples/triggers-demo-go/triggers-demo-go.exe` into the repo. It is an 11 MB built Windows artifact, and we normally generate binaries from source or publish them as release assets instead of checking them into `main`.

Can you drop that file from the PR and keep the example buildable from source only? Once that is out, I am happy to take another pass.

7vignesh:
@santoshkumarradha you're right, that should never have been committed. Dropped in 160ba517.

It was a build artifact from compiling the demo to verify it, and it got swept in because I added the whole directory instead of naming files. Also added an `examples/triggers-demo-go/.gitignore` for `triggers-demo-go` and `triggers-demo-go.exe` so a local build can't reintroduce it. Verified by rebuilding the binary and confirming git no longer sees it.

The example is unchanged otherwise and still builds from source via the Dockerfile or `go build`. Tests green.

santoshkumarradha (approved):
Thanks for the quick follow-up. Removing the checked-in binary and adding the local ignore fixes the blocker cleanly. I re-ran the Go trigger tests locally and this looks good to merge.

PR #740, opened 2026-07-09
fix(agentfield): add test for issue #624

jaythehardcoder:
## What
Adds test coverage for the behavior described in issue #624.

## Why
Fixes #624

## Testing
- [x] New test added

PR #921, opened 2026-08-16
fix(sdk/python): add AGENTFIELD_SKIP_IP_DETECTION opt-out for container IP probing

AmirF194:
## Root cause

`_build_callback_candidates()` in `sdk/python/agentfield/agent.py` calls `_detect_container_ip()` unconditionally whenever `_is_running_in_container()` is true, with no way to opt out. `_detect_container_ip()` sends blocking requests to AWS/Azure/GCP cloud-metadata endpoints (`169.254.169.254`, `metadata.google.internal`) and falls back to the third-party `api.ipify.org`, regardless of whether an explicit `AGENT_CALLBACK_URL` (or constructor `callback_url`) is already set. In a Kubernetes deployment with a restrictive `NetworkPolicy`, this fires on every agent start, flooding deny logs, and it also leaks pod egress to a third party for a best-effort candidate that most deployments never use.

## Fix

Add `AGENTFIELD_SKIP_IP_DETECTION` (default `false`, same boolean-env-var convention already used by `AGENTFIELD_LOG_PAYLOADS`/`AGENTFIELD_LOG_TRACKING`/`AGENTFIELD_LOG_FIRE` in `logger.py`) and gate the `_detect_container_ip()` call behind it. Default behavior is unchanged; the Railway-specific hint and the local-network/hostname/localhost fallbacks are untouched. Documented in `docs/ENVIRONMENT_VARIABLES.md`.

## On the prior closed attempts

Four earlier PRs on this issue (#698, #740, #770, #834) were closed as placeholder tests dropped into a root `tests/` directory that isn't part of this repo's layout, without an actual fix. This PR is different: it changes the real behavior in `sdk/python/agentfield/agent.py` and adds tests in `sdk/python/tests/test_agent_networking.py`, the file that already covers this exact function, following its existing `monkeypatch` style.

## Verification

- New tests (`test_build_callback_candidates_skips_ip_detection_when_disabled`, `test_build_callback_candidates_runs_ip_detection_by_default`, `test_ip_detection_disabled_is_case_insensitive`, `test_ip_detection_disabled_defaults_to_false`) fail on unmodified `main` (`AttributeError` / probe still called) and pass on this branch, in a clean Docker container.
- Full `sdk/python` suite (`./scripts/run_pytest.sh`) green on Python 3.10 and 3.12 (the matrix's low and high ends): 1968 passed, 4 pre-existing skips, no new failures.
- `ruff check .` on this branch reports the same 9 pre-existing findings as unmodified `main`, all inside a `build/lib/` pip-build artifact directory unrelated to this diff; nothing in the changed files.
- Not verified: an actual Kubernetes `NetworkPolicy` blocking egress (no cluster in this sandbox). The fix is a control-flow gate on an existing, already-tested function, not a change to its detection logic.


santoshkumarradha (approved):
Thanks for tightening this up. The opt-out gate is small, the networking tests passed locally, and the docs note is clear. The only remaining blocker I see is the CLA check. Once that’s signed, this should be good to go.

AmirF194:
Thanks for the review. Working on getting the CLA signed, will follow up here once it's cleared.

AbirAbbas:
Closing this in favour of #969, which lands the same opt-out from a clean-room implementation (written from #624's description, not from this diff) and also skips the probe automatically when a callback URL is already configured — the case pocesar originally reported. The CLA has been pending for ten days and we'd rather not leave #624 open longer; the diagnosis and the env-var approach were yours and #969 credits you for them. If you sign the CLA later we'd be glad to take future contributions.

AmirF194:
Thanks for the update, and for crediting the diagnosis. Glad to see #969 merged and covering the callback URL case too. Working on getting the CLA signed, would like to contribute again here.

PR #923, opened 2026-08-17
fix(sdk): preserve multimodal args with user prompt

### LINKED ISSUE #911: [Python SDK] ai() silently drops positional multimodal args when user= is also passed
## Summary

`AgentAI.ai()` silently discards **all** positional multimodal args (images, audio, video, files, text) whenever the `user=` keyword is also supplied. There is no warning and no error — the model simply never receives the image, and confidently answers as if none was given.

This is silent data loss on user-supplied input, and it fires on what is arguably the most natural way to write a vision call.

Confirmed empirically (twice) against SDK `0.1.63-rc.1`. The offending code is unchanged on `main` (currently `0.1.114`), so this is still live.

## Reproduction

Both cases run against model `openrouter/x-ai/grok-4.6`, with a real 1280x720 JPEG containing the text `Someone finally logged the damn calls`.

**Case C — positional prompt. WORKS.**

```python
result = await app.ai(
    image_from_file("photo.jpg"),
    "Transcribe the exact text in this image and describe it.",   # positional
    system="You are an image inspector.",
    schema=Out,
)
```

Result: `text_seen='Someone finally logged the damn calls'`, correct description.

**Case D — `user=` keyword. IMAGE SILENTLY DROPPED.**

```python
result = await app.ai(
    image_from_file("photo.jpg"),
    system="You are an image inspector.",
    user="Transcribe the exact text in this image.",   # user= instead of positional
    schema=Out,
)
```

Result: `text_seen=''`, `description='No image is provided or visible.'`

Note that `schema=` is **not** the cause — it is present in both cases and Case C works fine. The sole difference is `user=` vs. a positional prompt string.

## Expected vs Actual

**Expected:** Case D sends the image plus the instruction, and behaves like Case C. Passing the prompt via the parameter literally named `user` should not be a trap.

**Actual:** The image is dropped before the request is ever built. The model receives only a system message and a text-only user message. No warning, no error, no log line — the failure is indistinguishable from a model that simply cannot see.

## Root cause

`sdk/python/agentfield/agent_ai.py`, lines 641-647:

```python
# Handle flexible user input with intelligent processing
if user:
    messages.append({"role": "user", "content": user})
elif args:
    processed_content = self._process_multimodal_args(args)
    if processed_content:
        messages.extend(processed_content)
```

Because this is `if user: ... elif args:`, a truthy `user` shadows `args` entirely. `_process_multimodal_args()` is never called, so every positional argument is thrown away.

**This is not image-specific.** The same `elif` means a caller passing `user=` alongside positional *text* args loses that text too — the two input channels are mutually exclusive by accident rather than by design, and the API surface gives no hint of it.

The natural, obviously-correct-looking call is the broken one:

```python
result = await app.ai(
    image_from_file("photo.jpg"),
    system="You are an image inspector.",
    user="Transcribe the text in this image.",
    schema=Inspection,
)
```

The working form requires knowing, undocumented, to pass the prompt positionally:

```python
result = await app.ai(
    image_from_file("photo.jpg"),
    "Transcribe the text in this image.",   # positional, NOT user=
    system="You are an image inspector.",
    schema=Inspection,
)
```

## Suggested fix

**1. Preferred — merge instead of shadow.**

Build the user message from *both* sources. Process the positional args into content parts via `_process_multimodal_args()`, then append the `user=` string as an additional `{"type": "text", ...}` part in the same user message (order is a judgement call; appending the instruction after the media matches how the positional form already reads).

This makes the intuitive call work, loses nothing, and keeps every existing call site behaving identically — today no working call supplies both, precisely because supplying both is broken. It also fixes the text-args case for free.

**2. Minimum viable — fail loudly.**

If both `user` and non-empty `args` are supplied, raise a `ValueError` (or at the very least emit a warning). Silently truncating user-supplied input is the actual defect here; a loud failure is strictly better than the current behavior even if merging is deferred. But this only converts a silent bug into a papercut — it does not make the intuitive call work.

**3. Docs / DX.**

The multimodal documentation only ever demonstrates positional usage, so the `user=` conflict is entirely undocumented. Whichever fix lands, the interaction between `user=` and positional args deserves an explicit note.

I'd advocate for (1): the merge is a few lines, is backwards-compatible in practice, and removes the footgun rather than relabeling it. (2) is a reasonable stopgap if (1) needs more design discussion, and (3) should happen either way.

## Environment

- SDK: `agentfield` Python `0.1.63-rc.1` (bug reproduced); code path unchanged at `main` / `0.1.114`
- Model: `openrouter/x-ai/grok-4.6`
- File: `sdk/python/agentfield/agent_ai.py:641`

mikemikimike:
## Summary

Fixes #911 by preserving positional multimodal arguments when `user=` is also provided. Positional media/text is merged into the final user message, with the keyword prompt appended as a text content part.

## Tests

- `python -m pytest tests/test_agent_ai.py tests/test_agent_ai_comprehensive.py --no-cov -q` (30 passed)
- `python -m ruff check agentfield/agent_ai.py tests/test_agent_ai_comprehensive.py`
- `python -m ruff format --check agentfield/agent_ai.py tests/test_agent_ai_comprehensive.py`

The default SDK suite also ran; its 11 failures are unrelated Windows environment/encoding and shell-subprocess tests.

### DIFF
--- a/sdk/python/agentfield/agent_ai.py
+++ b/sdk/python/agentfield/agent_ai.py
@@ -655,13 +655,34 @@ class AgentAI:
             if system:
                 messages.append({"role": "system", "content": system})
 
-        # Handle flexible user input with intelligent processing
-        if user:
-            messages.append({"role": "user", "content": user})
-        elif args:
-            processed_content = self._process_multimodal_args(args)
-            if processed_content:
-                messages.extend(processed_content)
+        # Handle flexible user input with intelligent processing. When both
+        # input styles are used, keep the positional multimodal content and
+        # append the keyword prompt to the final user message.
+        processed_content = self._process_multimodal_args(args) if args else []
+        if user and processed_content:
+            user_message = next(
+                (
+                    message
+                    for message in reversed(processed_content)
+                    if message.get("role") == "user"
+                ),
+                None,
+            )
+            if user_message is not None:
+                content = user_message.get("content")
+                if isinstance(content, list):
+                    content.append({"type": "text", "text": user})
+                else:
+                    user_message["content"] = [
+                        {"type": "text", "text": content},
+                        {"type": "text", "text": user},
+                    ]
+            else:
+                processed_content.append({"role": "user", "content": user})
+        elif user:
+            processed_content.append({"role": "user", "content": user})
+
+        messages.extend(processed_content)
 
         litellm_module = litellm if hasattr(litellm, "acompletion") else None
 
--- a/sdk/python/tests/test_agent_ai_comprehensive.py
+++ b/sdk/python/tests/test_agent_ai_comprehensive.py
@@ -173,6 +173,39 @@ async def test_ai_multimodal_input_processing(monkeypatch, agent_with_ai):
 
 
 @pytest.mark.asyncio
+async def test_ai_merges_user_keyword_with_positional_multimodal_args(
+    monkeypatch, agent_with_ai
+):
+    """Keep positional media and text when a keyword user prompt is supplied."""
+    from agentfield.multimodal import Image, Text
+
+    litellm_module = setup_litellm_stub(monkeypatch)
+    litellm_module.acompletion.return_value = make_chat_response("image analyzed")
+
+    ai = AgentAI(agent_with_ai)
+    await ai.ai(
+        Image.from_url("https://example.com/image.jpg"),
+        Text(text="Positional context"),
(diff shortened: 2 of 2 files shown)


santoshkumarradha (approved):
Thanks, this fixes the multimodal/user merge path cleanly. I reran the targeted Python tests locally and the positional media plus keyword prompt behavior looks right. This looks good to merge.

PR #924, opened 2026-08-17
feat(install): provision the aforge harness binary alongside af (curl, desktop, docker)

AbirAbbas:
## What

`aforge` is the coding harness AgentField's own harness providers spawn (`aforge exec --json -w <dir>`) — but until now nothing in the product ever put that binary on a machine. The Python provider shipped, the Go/TS providers are on the way, and the install hint told users to `go build` it from a repo they cannot clone. This PR closes that gap on every surface that installs AgentField.

One provisioner in Go, called from everywhere:

| Surface | How it gets aforge |
| --- | --- |
| `curl … \| bash` (`scripts/install.sh`) | new `install_aforge()` → `af aforge ensure` |
| `iwr … \| iex` (`scripts/install.ps1`) | new `Install-Aforge` → `agentfield.exe aforge ensure` |
| `af skill install --all` | `aforge.EnsureBestEffort` once per `InstallAll` |
| AgentField Desktop | `aforge-companion.ts` shells out to `af aforge ensure` on launch |
| `Dockerfile.control-plane-cloud`, `.python-agent`, `.go-agent` | build-time fetch stage → `/usr/local/bin/aforge` |
| by hand | `af aforge ensure [--force]` |

`af harness doctor --provider aforge` now reports it, and `af harness doctor` lists it alongside the other providers.

**Pinned version: `aforge v0.1.0`** — aforge-v2's first semver release.

## Why these shapes

**Runtime ensure, not a bundled binary.** The desktop app deliberately does *not* add aforge to `extraResources`. Bundling would fork the install path and freeze the pinned version at package time, and the app would still need a runtime check for machines that already carry an older copy. One code path, and bumping `aforge.Version` upgrades every surface at once.

**Gzipped assets, hash of the uncompressed binary.** `checksums.txt` is published verbatim from the aforge release and carries the sha256 of the *uncompressed* binaries, so the download is decompressed before it is hashed. The decompressed size is capped (96 MiB) so a wrong or hostile endpoint cannot gzip-bomb the process.

**Soft fetch, hard verify in Docker.** The asset host goes live with a separate website deploy, so a 404 has to degrade to "this image ships without aforge" rather than break every image build in the meantime — but a download that *does* land is always checksum-verified, and a mismatch fails the build. The `COPY` takes the fetch stage's output *directory*, not a fixed file path: when the fetch was skipped the directory is empty and the COPY is a no-op, instead of planting a zero-byte `aforge` on PATH that the doctor would cheerfully report as installed. `Dockerfile.control-plane` is left alone — it is distroless, has no shell, and by construction runs agents elsewhere.

**The doctor needed one new idea, and one guard.** Providers do not agree on how to be asked for a version, hence per-spec `VersionArgs` — aforge is asked `version` first, then `--version`. And the probe requires **exit 0** before it believes any output, because a CLI that does not recognise a version argument answers with its whole usage banner; without that guard the doctor would file the banner as the installed version. aforge is otherwise held to exactly the same bar as every other provider: a binary that cannot name itself is `version_probe_failed` and unusable.

## Validation contract

Observable behaviours, each covered by a test or an end-to-end run:

- `af aforge ensure` on a clean machine leaves an executable `$AGENTFIELD_HOME/bin/aforge` whose sha256 matches the published checksum, plus a `.aforge.version` marker written **after** the binary.
- A checksum mismatch installs nothing and leaves no marker.
- A second `ensure` with a matching marker issues **no** HTTP request; a differing marker re-downloads; `--force` re-downloads regardless.
- `AGENTFIELD_SKIP_AFORGE=1` is a no-op everywhere (CLI, skill install, desktop, and the shell installers via `--no-aforge` / `AFORGE_MODE=none`).
- `AGENTFIELD_AFORGE_BASE_URL` overrides the whole base URL; `Options.BaseURL` beats the env var, which beats the default.
- An unsupported OS/arch is a silent no-op, not an error.
- Provisioning **never** fails the operation that requested it — skill install, both shell installers and the desktop launch all continue after a failed download.
- `InstallAll` fetches the binary exactly once, and `--dry-run` fetches nothing.
- The harness doctor reports an aforge that prints a version as `usable=true` with that exact version and no issues; an aforge that only prints a usage banner on a non-zero exit as `usable=false, issues=["version_probe_failed"]`.
- The doctor finds a binary that is in `$AGENTFIELD_HOME/bin` but not yet on `PATH`.
- Docker images that run agent nodes have a working `aforge` on `PATH`.

## How verified

Go / Python / desktop suites plus real end-to-end runs against a mirror of the published layout (URL withheld — it is a local host).

Re-verified for the **v0.1.0** pin:

- `go test ./internal/aforge/... ./internal/cli/... ./internal/skillkit/... -count=1` — all pass, before and after rebasing onto `main`. `GOFLAGS=-buildvcs=false go build ./...` clean (the build step this repo's Control Plane CI runs).
- `gofmt -l` clean on every touched file. `golangci-lint run` reports **zero** findings in the files this PR touches; the 97 it does report are pre-existing elsewhere in `control-plane` (and that CI step is `continue-on-error`).
- **Real provisioning:** `af aforge ensure` against the v0.1.0 layout → `bin/aforge` sha256 `e3b8f7bd…13a32dc`, byte-identical to the release `checksums.txt` entry for `aforge-linux-amd64`; marker file reads `v0.1.0`. The installed binary answers **both** `aforge version` and `aforge --version` with `aforge v0.1.0`, exit 0. A second `ensure` is a true no-op — same inode, mtime and size.
- **Doctor, positive:** `af harness doctor --provider aforge --json` with nothing on `PATH` → `installed: true, version: "aforge v0.1.0", usable: true, issues: []`, resolved out of `$AGENTFIELD_HOME/bin`. Exit 0, human output `aforge: ready (aforge v0.1.0)`.
- **Doctor, negative (the behaviour change):** the same command against a stub `aforge` that prints a usage banner and exits 1 → `usable: false, issues: ["version_probe_failed"]` and a non-zero exit. Under the old `VersionOptional` allowance this same stub reported `usable: true`.
- **Docker:** the `aforge-dist` stage of `Dockerfile.control-plane-cloud` built with `--no-cache` against the v0.1.0 layout — `aforge-linux-amd64: OK` from `sha256sum -c`, and the soft-fail warning branch was **not** taken, so the asset genuinely resolved. Running the stage's binary prints `aforge v0.1.0`. `.python-agent` and `.go-agent` carry a byte-identical fetch stage and the same `AFORGE_VERSION=v0.1.0` arg.

Carried over from the pre-bump verification (unchanged by this bump):

- `ruff check .` (pinned 0.15.22) clean; `./scripts/run_pytest.sh -k "availability or harness"` → 246 passed.
- desktop: `npm ci`, `npm run typecheck`, `npm test` → 435 tests pass (10 of them new).
- `bash -n scripts/install.sh` clean. `scripts/install.ps1` parses with zero errors under the PowerShell parser.
- **Mutation check on the doctor fix:** removing the `exit != 0 → not a version` guard makes the probe test fail, i.e. the test catches the real bug rather than mirroring the code.
- **Real execution:** the provisioned binary ran `aforge exec --json` to completion — exit 0, one JSON line (`"stop":"done"`, `"artifacts":["hello.txt"]`), and the file it was asked to write was actually there.
- **Installer:** `install_aforge` from the committed script provisions successfully against a real `af`; `AFORGE_MODE=none` and `--no-aforge` skip it; a non-executable `af` warns and returns 0. A full `install.sh` run against release `v0.1.129` — whose `af` predates this command — printed the warning and still exited **0** with "installed successfully", which is the contract.
- **Coverage:** the patch-coverage gate first came in at 79.00% against an 80% floor, all of it error branches in the new provisioner. The follow-up test commit takes `internal/aforge` from 78.0% to 95.5% of statements and the gate's `control-plane` patch coverage to **94.00%** ✅.
- **Docker, full images:** all three built and smoke-tested; the in-image sha256 matched the release checksum. The alpine/musl go-agent image runs the same binary (it is a static ELF). The 404 path was exercised: build succeeds, prints the warning, and ships no `aforge` rather than a broken one.

## Notes for review

- **The pin is now a release tag, not an edge build.** `aforge.Version` in `control-plane/internal/aforge/ensure.go` is `v0.1.0`; the three Dockerfiles carry the matching `ARG AFORGE_VERSION=v0.1.0`. Every other surface reads the Go constant, so a future bump is those four lines.
- **`VersionOptional` is gone.** It existed only because the previously pinned edge build had no `version` subcommand. v0.1.0 has a real `version` (and `--version`), so the allowance is deleted rather than left dangling, and aforge now fails the doctor like any other provider when it cannot report a version.
- **The public download host still needs the matching website deploy** before it serves the `v0.1.0` directory. Until then the Docker fetch takes its documented soft-fail path and `af aforge ensure` reports a 404 — which is why every caller treats provisioning as best-effort.
- **Relationship to #905 / #913:** independent of both. The only overlap is a trivial adjacent-line conflict with #913 in `control-plane/internal/cli/harness_doctor.go`: both PRs add a row to `harnessProviderSpecs`. Resolve by keeping **both** rows — this PR's `aforge` entry and #913's — leaving `VersionArgs` unset on #913's row (it defaults to `{"--version"}`, i.e. today's behaviour).

## Follow-ups

- Flip the Docker fetch from soft-fail to hard-fail once the asset host is live, so a missing harness can never ship silently. Deliberately **not** in this PR.
- `Dockerfile.demo-python-agent` / `Dockerfile.demo-go-agent` were left alone; add aforge there if the demos grow harness calls.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,10 @@ https://github.com/user-attachments/assets/9fb7b1cf-26de-4b9b-9ba2-917252cc26ec
 curl -fsSL https://agentfield.ai/install.sh | bash
 ```
 
+The installer also drops the `aforge` coding harness beside `af` in
+`~/.agentfield/bin`, so harness-backed agents work out of the box; skip it with
+`--no-aforge`.
+
 On macOS the installer also registers the control plane to start at login (under
 launchd) and adds a menu-bar icon. Stop it with `af service stop` or the menu-bar
 icon — a plain `kill` looks like a crash and it restarts. `af service status`
--- a/control-plane/.env.example
+++ b/control-plane/.env.example
@@ -79,6 +79,15 @@ AGENTFIELD_STORAGE_MODE=local
 # AGENTFIELD_LOG_BUFFER_BYTES=4194304
 # AGENTFIELD_LOG_MAX_LINE_BYTES=16384
 
+# aforge coding harness (provisioned by `af aforge ensure`, the curl installer,
+# the desktop app and the agent images into $AGENTFIELD_HOME/bin/aforge).
+# Whole-base override for the download host — the pinned version is NOT appended,
+# so point it at a directory that already contains the assets + checksums.txt.
+# AGENTFIELD_AFORGE_BASE_URL=https://agentfield.ai/downloads/aforge/build-9b3ff482de3f
+# Set to 1 to make every aforge provisioning step a no-op (air-gapped hosts, or
+# images that vendor their own harness).
+# AGENTFIELD_SKIP_AFORGE=1
+
 # Development/Debug
 # GIN_MODE=debug
 # LOG_LEVEL=info
--- a/control-plane/.env.example
+++ b/control-plane/.env.example
@@ -83,7 +83,7 @@ AGENTFIELD_STORAGE_MODE=local
 # the desktop app and the agent images into $AGENTFIELD_HOME/bin/aforge).
 # Whole-base override for the download host — the pinned version is NOT appended,
 # so point it at a directory that already contains the assets + checksums.txt.
-# AGENTFIELD_AFORGE_BASE_URL=https://agentfield.ai/downloads/aforge/build-9b3ff482de3f
+# AGENTFIELD_AFORGE_BASE_URL=https://agentfield.ai/downloads/aforge/v0.1.0
 # Set to 1 to make every aforge provisioning step a no-op (air-gapped hosts, or
 # images that vendor their own harness).
 # AGENTFIELD_SKIP_AFORGE=1
--- a/control-plane/internal/aforge/ensure.go
+++ b/control-plane/internal/aforge/ensure.go
@@ -1,0 +1,256 @@
+// Package aforge provisions the pinned aforge coding-harness binary that AgentField's harness providers spawn.
+package aforge
+
+import (
+	"bufio"
+	"compress/gzip"
+	"crypto/sha256"
+	"encoding/hex"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"time"
+)
+
+const (
+	// Version is deliberately pinned. Bump it only when af should distribute a
+	// newer, reviewed aforge build.
+	Version        = "build-9b3ff482de3f"
+	defaultBaseURL = "https://agentfield.ai/downloads/aforge/" + Version
+	baseURLEnv     = "AGENTFIELD_AFORGE_BASE_URL"
+	skipEnv        = "AGENTFIELD_SKIP_AFORGE"
+	versionMarker  = ".aforge.version"
+	// aforge is a ~35MB static Go binary; cap the DECOMPRESSED stream so a
+	// hostile or wrong endpoint cannot gzip-bomb the process.
+	maxDecompressedBytes = 96 << 20
+)
+
+// Options controls how Ensure selects and installs aforge.
+type Options struct {
+	GOOS, GOARCH string
+	Home         string
+	BaseURL      string
+	Client       *http.Client
+	// Force re-downloads even when the version marker already matches
+	// (what `af aforge ensure --force` sets).
+	Force bool
+}
+
+// AssetName returns the release asset name for a platform (no .gz suffix).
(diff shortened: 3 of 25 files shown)


AbirAbbas:
recheck

PR #937, opened 2026-08-19
Ship the desktop app with bundled agent nodes, missing-key surfacing, and the offload-by-default skill

santoshkumarradha:
## What this does

Ships the desktop app with agents. `swe-planner` (SWE-AF) and `pr-af` leave the marketplace catalog and become **bundled nodes**: provisioned automatically on first launch through the control-plane install API, shown arriving live in the Agents view, and surfaced with a keys banner + native notification when they land without API keys. The `agentfield-use` skill is rewritten (v0.8.0) around offload-by-default so coding agents route work to these nodes proactively.

### Desktop — bundled provisioning
- `shared/bundled.ts` (new): `swe-planner` + `pr-af`, sourced at bare repo URLs so the `superseded_by` redirect migrates existing Python-node users. Marketplace catalog keeps only `sec-af` and `cloudsecurity-af`; `catalogEntry()` resolves over both lists so bundled nodes stay updatable without widening the vetted-name guard.
- `main/bundledAgents.ts` (new): pure planner + injected-deps runner (aforge-companion pattern). Runs after autostart, sequentially, sharing the install mutex. Uninstall sticks (`settings.provisionedBundled`); failures retry next launch; nodes are installed but **not** started — both need keys a first-run user hasn't entered.
- Agents view renders provisioning rows with live install output; first launch routes to Agents instead of the empty-state marketplace.

### Desktop — missing-keys surfacing
- `KeysBanner` (new): non-dismissible strip naming blocked agents, fed by `getEnvReports()` (the encrypted-store-backed source that actually gates `af run`). Star prompt yields to it.
- `main/keyNotice.ts` (new): one native notification per provisioning event when installed nodes still need keys; at-most-once via `settings.keyNoticeShown`; degrades silently headless.

### Skill — agentfield-use v0.7.0 → v0.8.0
- v0.7.0: missing keys become a blocking handoff — `af run` before dispatch, hand the user the exact `af secrets set` line, never retry/substitute; warns off store-blind `af doctor` / `af config --list`.
- v0.8.0: offload-by-default doctrine ("AgentField subharness" user-facing framing), CLI golden path (`af call --schema` → `--async` → `af wait`/`af tail`), monitor decision table (webhooks scoped to services/CI), live `/ui/runs/<run_id>` link handed to the user on every dispatch, honest cost reporting (window aggregates; per-run cost not yet exposed). Trigger is coverage-based, not category- or size-based — new subharnesses are covered by runtime discovery without editing the skill. Embedded copy synced; contract tests added.

## Verification
- Desktop: typecheck clean, **494/494 tests**.
- Skillkit: `go build ./...` clean; only failure is pre-existing `TestInstallAgentfieldUseEnsuresFurrow` (no furrow binary on the dev box; fails identically on clean main).
- **End-to-end in fresh containers** (Ubuntu 24.04 arm64):
  - CLI journey: installer → server → `af install SWE-AF` (redirect verified) → key via encrypted store → `implement_issue` on deepseek-v4-flash: 1 commit, tests pass, 82s, $0.026.
  - Desktop journey: real Electron main process under Xvfb — zero-click first run provisioned both nodes, settings persisted, keys reported, second launch adopted the control plane and skipped re-provisioning.

## Known follow-ups (deliberately out of scope)
- Coding runtime (`opencode`) is not provisioned — swe-planner's classic loop fails silently without it (execution `succeeded`, `result.success:false`).
- `swe-pro` vendored binaries cover darwin-arm64 + linux-amd64 only (no Windows).
- Store-blind checks: post-install warning, `af doctor`, `af config --list` read env/`.env`, not the encrypted store.
- Autostart attempts keyless agents each launch (noisy, harmless); per-run cost endpoint.

## Takeover follow-ups (review round, commits after `ef6e10fe`)

A review of the branch (5 lenses + adversarial verification) confirmed a handful of real issues; each is fixed in its own commit:

- **Provisioning targeted whatever control plane was active, including a configured cloud one.** For a user with `cloud.enabled` the first launch after upgrading would remote-install swe-planner and pr-af onto their hosted control plane, and the app-global `provisionedBundled` latch would then suppress local provisioning for good. Provisioning is now local-only (`cloudActive` gate), treats an unreadable registry as "not ready" rather than "nothing installed", and asks `hasInstallApi()` before seeding rows so an older control plane is skipped instead of producing two red rows every launch. (`99001556`)
- **"Uninstall sticks" only held for nodes the app itself installed.** A node the user already had was skipped but never recorded, so removing it later brought it back. Installed-but-unrecorded bundled nodes are now adopted into `provisionedBundled`. (`99001556`)
- **First launch landed in the marketplace, not on the arriving bundled rows.** The default route is applied on the renderer's first snapshot, which precedes autostart + row seeding on a cold launch, so `defaultView`'s bundled branch never fired. A one-shot `shouldRerouteToBundled` switches an untouched add-mode launch to the Agents library when rows appear; deep links and user navigation are never overridden. (`c670e8e9`)
- **A freshly-provisioned row had no "Needs keys" chip / Keys button** (env reports were mount-only) and **a node showed twice** while the next one installed. Env reports now reload on roster change; `visibleBundledRows` drops provisioning rows that already have a registry row. (`7f2a5a35`)
- **No curated way back after uninstalling a bundled node.** The catalog IPC lists bundled nodes again; installed ones show "Installed", ones being provisioned show a disabled "Installing…". (`0904a7b1`)
- **`KeysBanner` / key notice false positive for env-supplied keys.** `GET /api/ui/v1/agents/:id/secrets` computed `is_set` from the encrypted store only, while the runner resolves the control plane's process environment first — a key exported in the shell started the agent fine but was reported missing, which this PR escalated into a non-dismissible banner + OS notification. The handler now reports `env: true` (presence only) and `is_set` reflects runtime truth; the desktop maps env-only keys to the existing "From environment" status. (`a4d4b5ca`)

- **Agents started by a control plane on any port other than 8080 registered with localhost:8080.** The control plane hands its agents the URL from its own `AGENTFIELD_SERVER` env and otherwise falls back to `http://localhost:8080` regardless of the port it listens on; the app only pinned `AGENTFIELD_PORT`. Observed live: swe-planner + the swe-pro engine joined an unrelated control plane sitting on 8080. The app now spawns the server with `AGENTFIELD_SERVER=http://localhost:<port>` too. (`a57d490b`) — the control-plane side (derive the URL from the bound port) is a follow-up.
- **Early snapshots went to :8080 until autostart resolved the port**, so the cold-launch route could be decided on a foreign server's empty registry; and **a cold autostart routed a stocked library into the marketplace** because the first snapshot saw "no registry" while the server was coming up. The local port is now seeded from settings before the first poll, and the default route waits for a readable registry. (`6057173d`, `d0319931`)

### Local verification (WSL, isolated `HOME`, control plane built from this branch on :18480)
- Fresh first launch: both nodes provisioned through the install API (SWE-AF redirect → `//go` verified), OS notification fired once, landing view = Agents library with both rows carrying "Needs keys" + Keys, banner "2 installed agents need API keys". Second launch: no re-provision, no second notice.
- swe-planner started from the app: `pro engine enabled: engine node "swe-pro" joins the control plane` — Pro engine on by default via the manifest, `swe-pro-linux-amd64` picked from `bin/`; engine + node register with the app's control plane (not 8080) after `a57d490b`.
- Uninstall sticks (pr-af removed → not re-provisioned next launch; still offered in the marketplace). Cloud connection enabled → `bundled: cloud control plane active — … local control plane only`. `OPENROUTER_API_KEY` in the app's environment → `is_set:true env:true`, agent autostarts, no banner.

Engine binaries: the Pro engine is on by default via the manifest (`SWE_PRO_ENGINE` default `"1"`), and `af install` brings every vendored engine build along. The vendored set was darwin-arm64 + linux-amd64 only; Agent-Field/SWE-AF#138 adds darwin-amd64 + linux-arm64 and makes the resolver Windows-ready. A Windows engine still needs a swe-pro-go port (unix-only syscalls in three files) — on Windows the node falls back to the classic loop.

### DIFF
--- a/control-plane/internal/handlers/ui/agent_secrets.go
+++ b/control-plane/internal/handlers/ui/agent_secrets.go
@@ -3,6 +3,7 @@ package ui
 import (
 	"encoding/json"
 	"net/http"
+	"os"
 	"regexp"
 	"sort"
 
@@ -23,18 +24,20 @@ var agentSecretKeyPattern = regexp.MustCompile(`^[A-Z][A-Z0-9_]*$`)
 type AgentSecretsHandler struct {
 	storage        storage.StorageProvider
 	agentfieldHome string
+	lookupEnv      func(string) (string, bool)
 }
 
 // NewAgentSecretsHandler creates an AgentSecretsHandler.
 func NewAgentSecretsHandler(storage storage.StorageProvider, agentfieldHome string) *AgentSecretsHandler {
-	return &AgentSecretsHandler{storage: storage, agentfieldHome: agentfieldHome}
+	return &AgentSecretsHandler{storage: storage, agentfieldHome: agentfieldHome, lookupEnv: os.LookupEnv}
 }
 
 type agentSecretStatus struct {
 	Key   string `json:"key"`
 	IsSet bool   `json:"is_set"`
+	Env   bool   `json:"env,omitempty"`
 	// Scope reports where the stored value lives ("node" or "global");
-	// empty when the key is not set anywhere.
+	// empty when there is no stored value, including environment-only keys.
 	Scope            string `json:"scope,omitempty"`
 	DeclaredScope    string `json:"declared_scope,omitempty"`
 	Description      string `json:"description,omitempty"`
@@ -66,9 +69,10 @@ type setAgentSecretRequest struct {
 }
 
 // ListAgentSecretsHandler lists secret names and whether each resolves for
-// this agent. Resolution mirrors the runner (EnvResolver): node scope first,
-// then global. Undeclared node-scoped keys are included because the runner
-// injects them; undeclared global keys are not injected, so they are omitted.
+// this agent. Resolution mirrors the runner (EnvResolver): a non-empty process
+// environment value first, then node store, then global store. Values are never
+// returned. Undeclared node-scoped keys are included because the runner injects
+// them; undeclared global keys are not injected, so they are omitted.
 func (h *AgentSecretsHandler) ListAgentSecretsHandler(c *gin.Context) {
 	agentPackage, ok := h.resolveAgentPackage(c)
 	if !ok {
@@ -123,9 +127,17 @@ func (h *AgentSecretsHandler) ListAgentSecretsHandler(c *gin.Context) {
 	}
 	sort.Strings(keys)
 
+	lookupEnv := h.lookupEnv
+	if lookupEnv == nil {
+		lookupEnv = os.LookupEnv
+	}
 	secrets := make([]agentSecretStatus, 0, len(keys))
 	for _, key := range keys {
 		status := agentSecretStatus{Key: key}
+		if value, ok := lookupEnv(key); ok && value != "" {
+			status.Env = true
+			status.IsSet = true
+		}
 		switch {
 		case inNode[key]:
 			status.IsSet = true
--- a/control-plane/internal/handlers/ui/agent_secrets_test.go
+++ b/control-plane/internal/handlers/ui/agent_secrets_test.go
@@ -112,6 +112,10 @@ func agentSecretsRequest(t *testing.T, router http.Handler, method, path, body s
 
 // Validation contract 1: PUT writes the node scope consumed by runner-side resolution.
 func TestAgentSecretsPutResolvesForRunner(t *testing.T) {
+	// EnvResolver prefers a non-empty process env value, so a developer
+	// machine with this key exported would resolve the host value instead of
+	// the stored one. Empty counts as unset; this keeps the test hermetic.
+	t.Setenv("OPENAI_API_KEY", "")
 	router, home := newAgentSecretsTestRouter(t)
 	response := agentSecretsRequest(t, router, http.MethodPut, "/agents/agent-x/secrets",
 		`{"key":"OPENAI_API_KEY","value":"sk-test"}`)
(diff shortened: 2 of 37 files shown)


santoshkumarradha (commented):
I found one concrete test issue on the latest branch state.

`TestAgentSecretsPutResolvesForRunner` is now host-env sensitive because `EnvResolver.Resolve()` prefers a non-empty process env value first. On a machine with `OPENAI_API_KEY` exported, the final assertion reads the host key instead of the stored `sk-test` value. I reproduced that locally: `origin/main` passes, this branch fails with my current env, and this branch passes again under `env -u OPENAI_API_KEY`.

Please neutralize process env in that test before resolving so the package stays deterministic outside CI too.

AbirAbbas:
Good catch — fixed in 4e7f9006: `TestAgentSecretsPutResolvesForRunner` now neutralizes `OPENAI_API_KEY` with `t.Setenv` (empty counts as unset in `EnvResolver.lookup`), so it resolves the stored value regardless of the host shell.

One note for the record: the sensitivity predates this branch — the test body is byte-identical on `origin/main`, and `OPENAI_API_KEY=host-value go test -run TestAgentSecretsPutResolvesForRunner` fails there too (`EnvResolver` has preferred a non-empty process env since it was introduced; this PR didn't change resolution order, it only made the *list* endpoint report it). My `origin/main` pass without `env -u` only reproduces when the run hits Go's test cache. Either way it belonged fixed here, since this PR is what makes env-first resolution load-bearing.

While verifying I ran both packages under a hostile env (`OPENAI_API_KEY`/`ANTHROPIC_API_KEY`/`OPENROUTER_API_KEY`/`GH_TOKEN` all exported): `internal/handlers/ui` is now fully hermetic; `internal/packages` has 15 pre-existing EnvResolver prompt tests that fail the same way on `origin/main` — left alone here, worth a small follow-up.

PR #939, opened 2026-08-20
Bundle sec-af and cloudsecurity-af with the desktop app

AbirAbbas:
## Summary

Promotes `sec-af` and `cloudsecurity-af` from marketplace catalog rows to `BUNDLED_NODES`, so a fresh desktop install provisions all four agent nodes (`swe-planner`, `pr-af`, `sec-af`, `cloudsecurity-af`) on first launch.

- Both repos carry the same `superseded_by: …//go` root-manifest redirect as SWE-AF and pr-af, and register under their product names, so the bundled-roster rules apply unchanged.
- The maintained nodes are Go — the move also retires the stale `python` language chips the catalog rows carried.
- `CATALOG` is now empty and stays as the seam for future marketplace-only rows / the remote catalog fetch.
- Users who already ran provisioning get exactly the two new nodes on their next launch (they are not in `provisionedBundled`); an uninstall still sticks.

## Testing

- `cd desktop && npm test` — 26 files, 521 tests pass; `npm run typecheck` clean.
- Live-verified on macOS against a machine that had already provisioned the original pair: dev-app launch planned `provisioning bundled nodes: sec-af, cloudsecurity-af`, both installed through the control-plane install API, registered under their product names from the `//go` redirect, and were recorded in `provisionedBundled`/`autostartAgents`. A subsequent launch reports `bundled nodes already provisioned` and auto-starts them.

### DIFF
--- a/desktop/src/main/agentfield.test.ts
+++ b/desktop/src/main/agentfield.test.ts
@@ -546,9 +546,14 @@ describe('fetchUsageStats', () => {
 })
 
 describe('install catalog', () => {
+  // catalogEntry resolves over the marketplace CATALOG and the bundled roster
+  // alike, so the shape and uniqueness invariants hold across their union.
+  // CATALOG itself is currently empty — every vetted node ships with the app.
+  const VETTED = [...CATALOG, ...BUNDLED_NODES]
+
   it('every entry has a name, description, and an https or af:// source', () => {
-    expect(CATALOG.length).toBeGreaterThan(0)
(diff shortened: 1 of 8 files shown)


santoshkumarradha (approved):
Thanks for tightening this up. I rechecked the manifest assumptions for both security nodes, ran `npm test` and `npm run typecheck` in a clean desktop worktree, and this looks good to me.

PR #943, opened 2026-08-21
fix(execute): resolve webhook not triggering on async status callback (#936)

### LINKED ISSUE #936: Async execute: webhook registration silently fails (webhook_registered=false, no error surfaced)
## Summary

`POST /api/v1/execute/async/<target>` silently fails to register the `webhook` config from the request body. The execution runs and succeeds, but `webhook_registered` comes back `false`, no callback is ever delivered, and no error is surfaced to the caller.

## Environment

- af v0.1.93 (built 2026-06-22), macOS (Darwin 25.2), control plane `af server` on localhost:8080
- Long-running Go agent registered via the Go SDK

## Reproduction

1. Submit an async execution with a valid webhook config:

```json
POST /api/v1/execute/async/deep_research.generate_document
{
  "input": { "...": "..." },
  "context": { "...": "..." },
  "webhook": {
    "url": "http://localhost:3000/api/webhooks/agent-orchestration",
    "secret": "<secret>",
    "headers": { "X-App-Webhook": "followthethread-v1" }
  }
}
```

2. Execution completes normally (`status: succeeded`).
3. `GET /api/v1/executions/<id>` returns `"webhook_registered": false`. The callback URL never receives a request.

Observed on every async submission from this client, across two different reasoners and two agent nodes. The URL is well-formed (scheme + host, no embedded credentials), so current `normalizeWebhookRequest` validation in HEAD would accept it.

## Impact

Any client relying on webhook completion notifications hangs forever on tasks whose results are actually ready. In our app, document-generation tasks sat in `queued` while the finished result was on the control plane. We worked around it with client-side polling that reads `GET /api/v1/executions/<id>` and ingests terminal results directly.

## Expected

- `webhook_registered: true` for a valid webhook config, with a callback on terminal state.
- If registration is rejected, the async submit response should carry the rejection reason (a `webhook_error` field) instead of failing silently.

## Notes

- The async submit response also omits any webhook error detail, so clients can only detect the failure by reading back the execution record.
- Possibly fixed after 0.1.93; current handler source (`internal/handlers/execute.go`, `normalizeWebhookRequest`) records a `webhookError` but we could not confirm the running binary matches. If it is fixed, a release + a `webhook_error` passthrough in the submit response would close this.

7vignesh:
## Summary

Fixes the silent webhook failure reported in #936. The `WebhookRegistered` field on `types.Execution` has a `db:"-"` tag and is never persisted, causing two failures:

1. **Status callback path:** when an async execution completes via the agent's `PUT /executions/:id/status` callback, the handler checked `updated.WebhookRegistered` (always `false` from DB) and never triggered webhook delivery  clients relying on callbacks hung forever.
2. **GET /executions/:id:** `renderStatus` read the same unpersisted field, always reporting `webhook_registered: false` to polling clients.

**Fix:** Add `HasExecutionWebhook` to the `ExecutionStore` interface (already exists on `StorageProvider`) and use it in both the status-callback trigger path and `renderStatusWithApproval` to resolve webhook state from the `execution_webhooks` table.

## Type of change

- [x] Bug fix

## Test plan

- [x] `cd control-plane && go test ./internal/handlers/ -run TestUpdateExecutionStatusHandler_WebhookTriggeredFromStore -v`
- [x] `cd control-plane && go test ./internal/handlers/ -run TestGetExecutionStatusHandler_WebhookRegisteredFromStore -v`
- [x] `cd control-plane && go test ./internal/handlers/ -run TestGetExecutionStatusHandler_NoWebhook -v`
- [x] `cd control-plane && go test ./internal/handlers/ -run TestUpdateExecutionStatusHandler_WithWebhook -v`
- [x] `cd control-plane && go test ./internal/handlers/ -timeout 120s` (full suite passes)
- [x] `cd control-plane && go vet ./internal/handlers/...`
- [x] `cd control-plane && golangci-lint run ./internal/handlers/` (no new warnings)

## Test coverage

## Checklist

- [x] I have linked any related issues (Fixes #936).

## Related issues / PRs

Fixes #936

### DIFF
--- a/control-plane/internal/handlers/execute.go
+++ b/control-plane/internal/handlers/execute.go
@@ -39,6 +39,7 @@ type ExecutionStore interface {
 	UpdateExecutionRecord(ctx context.Context, executionID string, update func(*types.Execution) (*types.Execution, error)) (*types.Execution, error)
 	QueryExecutionRecords(ctx context.Context, filter types.ExecutionFilter) ([]*types.Execution, error)
 	RegisterExecutionWebhook(ctx context.Context, webhook *types.ExecutionWebhook) error
+	HasExecutionWebhook(ctx context.Context, executionID string) (bool, error)
 	StoreWorkflowExecution(ctx context.Context, execution *types.WorkflowExecution) error
 	UpdateWorkflowExecution(ctx context.Context, executionID string, updateFunc func(*types.WorkflowExecution) (*types.WorkflowExecution, error)) error
 	GetWorkflowExecution(ctx context.Context, executionID string) (*types.WorkflowExecution, error)
@@ -1007,7 +1008,7 @@ func (c *executionController) handleStatusUpdate(ctx *gin.Context) {
 
 	if isTerminal {
 		c.updateWorkflowExecutionFinalState(reqCtx, executionID, types.ExecutionStatus(normalizedStatus), updated.ResultPayload, elapsed, errorMsg)
-		if updated.WebhookRegistered {
+		if hasWH, _ := c.store.HasExecutionWebhook(reqCtx, executionID); hasWH {
 			c.triggerWebhook(executionID)
 		}
 	}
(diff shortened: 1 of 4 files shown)


santoshkumarradha (approved):
Nice catch. I reproduced the webhook-state gap locally against the handler surface, and this fix lines the callback path and status response up with the persisted webhook table. Focused handler tests passed for me, and the GitHub checks are green.

PR #945, opened 2026-08-22
fix(execute): propagate agent 4xx in async lane instead of blanket 502

### LINKED ISSUE #862: execute: async lane returns 502 for a client-input rejection; sync lane returns the agent's 4xx
## Summary

`POST /api/v1/execute/{target}` returns **HTTP 502 Bad Gateway for every failed execution
that took the async-acknowledgement lane**, including a client-input rejection the agent
raised on purpose. The synchronous lane, for the identical failure, correctly returns the
agent's 4xx. So the status a caller sees is decided by an internal transport detail rather
than by what went wrong, and a client that keys on status — the normal thing to do — reads
its own bad request as an upstream outage.

## Reproduction (measured 2026-08-02, control plane `1.0.0`, Python SDK `0.1.91`)

A reasoner that validates its own input and raises on a bad field:

```
POST /api/v1/execute/hga-reasoner.reason
{"input":{"text":"probe","rules":[{"about":"http://x/A","classified_as":"http://x/B","not_a_field":1}]}}

HTTP/1.1 502 Bad Gateway
{"execution_id":"exec_20260802_141424_2b7kcbrl","run_id":"run_20260802_141424_9889tyoc",
 "status":"failed",
 "error_message":"invalid_input: RuleSpec rejected 1 field in `rules[0]` — not_a_field: Extra inputs are not permitted",
 "duration_ms":0,"finished_at":"2026-08-02T14:14:24Z"}
```

The same node, refused by the SDK's own pre-dispatch field check, returns the right status:

```
POST /api/v1/execute/hga-reasoner.reason   {"input":{}}

HTTP/1.1 422 Unprocessable Entity
{"error":"agent error (422): {\"detail\":\"Missing required field: text\"}",
 "error_category":"agent_error","error_details":{"detail":"Missing required field: text"},
 "status":"failed"}
```

Both are client-input errors against a perfectly healthy upstream. Only the second one says so.

## Root cause

Three things compose. All line numbers are `main` as of this report.

**1. The async-completion branch does not classify at all.**
`control-plane/internal/handlers/execute.go:276-294` — after an agent answers `202` and the
callback lands, any `ExecutionStatusFailed` is written out with a hardcoded status:

```go
if exec.Status == types.ExecutionStatusFailed {
    ...
    ctx.JSON(http.StatusBadGateway, response)   // execute.go:293
    return
}
```

Its sibling, `writeExecutionError` (`execute.go:1979-1984`), gets this right for the
synchronous lane and even says so in a comment:

```go
// Propagate 4xx status codes from the agent (client-facing errors);
// use 502 Bad Gateway for 5xx (upstream server failure).
httpStatus := http.StatusBadGateway
if ce.statusCode >= 400 && ce.statusCode < 500 {
    httpStatus = ce.statusCode
}
```

**2. The async lane cannot carry a status even if the branch wanted one.**
`sdk/python/agentfield/agent.py:2475-2486` builds the failure callback from
`str(exc)` + `getattr(exc, "error_details", None)`. An `HTTPException(status_code=422)`
raised inside the reasoner is flattened to its `str()`; the code is gone before the
callback is posted. (The agent takes this lane whenever `X-Execution-ID` is present and
`agentfield_server` is set — `agent.py:1932` — which is every control-plane-routed call.)

**3. `error_details` is dropped by the callback handler.**
The SDK sends it; `executionStatusUpdateRequest` (`execute.go:116-124`) has no field for
it, so it is silently discarded. Only `result` survives a failed callback
(`execute.go:498-504,556-559`), which is why `ReasonerFailed` (#697) works and the generic
path does not.

The net effect: a reasoner has **no way** to return a client-facing status through the
async lane. The one place that classifies is the pre-dispatch check at `agent.py:1886`,
which by construction only ever sees missing/None top-level fields — anything a reasoner
validates itself is already past it. Notably `_execute_reasoner_endpoint` at
`agent.py:2265-2270` converts a Pydantic `ValidationError` into `_HandlerInputError`
with a comment saying the intent is a safe 422, and then raises it at a point where the
only `except _HandlerInputError` block (`agent.py:1881`) has already been passed.

## Impact

For any agent that validates its own input — which is the recommended shape, since the
runtime's pre-dispatch check only covers top-level presence — every contract rejection is
indistinguishable from an outage at the status layer. Concretely, on a schema change the
consumer's UI says *"the agent is unavailable"* instead of *"this request uses a field the
agent no longer accepts."* The first sends someone to check pods; the second sends them to
the request.

## Suggested fix

Smallest change that fixes the observable behaviour: give the async-completion branch the
same classification the sync branch already has, sourced from the callback.

1. Add `error_status` (or reuse `error_details`) to `executionStatusUpdateRequest` and
   persist it on the execution record.
2. In the SDK's `_execute_async_with_callback` failure path, populate it from
   `HTTPException.status_code` / `ExecuteError.status_code` when the raised exception
   carries one.
3. At `execute.go:276-294`, use that status when it is 4xx, keeping 502 as the default.

A useful independent improvement: restoring `error_details` to the callback struct would
let any agent return a structured, machine-readable refusal body without having to raise
`ReasonerFailed` purely for its `result` side-effect.

Happy to open a PR for any part of this if the shape above is agreeable.

7vignesh:
## Summary

Fixes #862. The async-completion branch in `handleSync` hardcoded HTTP 502 for all failed executions, making client-input rejections (e.g. 422 from input validation) indistinguishable from upstream outages. A reasoner validating its own input would return `invalid_input: ...` but the caller saw 502 Bad Gateway — sending operators to check pods instead of the request.

**Root cause:** The `callError.statusCode` is lost once the execution is persisted; the async lane (which reads the execution from DB after the agent calls back) had no way to recover the original HTTP status.

**Fix (two-pronged):**

### Control plane
- Add `error_status_code` field to `executionStatusUpdateRequest` so the SDK can forward the HTTP status in its failure callback
- When `error_status_code` is 4xx, encode it in `StatusReason` as `agent_client_error:<code>` for persistence
- Add `httpStatusForFailedExecution()` helper that resolves the HTTP status from:
  1. `StatusReason` — encoded client errors and known server-side categories (`agent_timeout`→504, `target_not_found`→404, etc.)
  2. `ErrorMessage` — parses `"agent error (NNN):"` pattern as fallback for existing executions
  3. Default: 502
- Replace hardcoded 502 in the async-completion branch with the helper

### Python SDK
- In `_execute_async_with_callback` failure path, propagate `error_status_code` from `exception.status_code` / `exception.code` when the value is a valid HTTP status (400–599)

**Backward-compatible:** SDKs that don't send `error_status_code` get existing behavior (502 default). SDKs that do send it get correct 4xx propagation immediately.

## Type of change

- [x] Bug fix

## Test plan

- [x] `cd control-plane && go test ./internal/handlers/ -run TestHttpStatusForFailedExecution -v` (12 cases: encoded 4xx, category mapping, error message parsing, defaults)
- [x] `cd control-plane && go test ./internal/handlers/ -run TestUpdateExecutionStatusHandler_ErrorStatusCode -v`
- [x] `cd control-plane && go test ./internal/handlers/ -run TestExecuteHandler_AgentError -v` (existing 500→502 preserved)
- [x] `cd control-plane && go test ./internal/handlers/ -timeout 120s` (full suite)
- [x] `cd control-plane && go vet ./internal/handlers/...`
- [x] `cd control-plane && golangci-lint run ./internal/handlers/` (no new warnings)
- [x] `cd sdk/python && python -m pytest tests/test_usage_transport.py` (async callback path)
- [x] `python -c "import agentfield.agent"` (syntax OK)

## Test coverage

## Checklist

- [x] I have linked any related issues (Fixes #862).

## Related issues / PRs

Fixes #862

Note: `TestCallAgent_ErrorResponse` is a pre-existing flaky test (timing-sensitive `elapsed > 0` assertion) — confirmed by running it with `-count=5` on both main and this branch.

### DIFF
--- a/control-plane/internal/handlers/execute.go
+++ b/control-plane/internal/handlers/execute.go
@@ -125,6 +125,11 @@ type executionStatusUpdateRequest struct {
 	DurationMS   *int64                 `json:"duration_ms,omitempty"`
 	CompletedAt  *time.Time             `json:"completed_at,omitempty"`
 	Progress     *int                   `json:"progress,omitempty"`
+	// ErrorStatusCode is the optional HTTP status code the agent SDK sends to
+	// indicate whether the failure is client-facing (4xx) or an upstream error
+	// (5xx). When present and in the 4xx range, the control plane propagates it
+	// to callers instead of returning a blanket 502 Bad Gateway.
+	ErrorStatusCode *int `json:"error_status_code,omitempty"`
 	// Usage is the optional token/cost usage object the agent SDK attaches at
 	// the top level of the status-callback body. It is a sibling of Result, so
 	// it is never persisted into the result payload. Absent = no-op.
@@ -359,7 +364,7 @@ func (c *executionController) handleSync(ctx *gin.Context) {
 			}
 			ctx.Header("X-Execution-ID", exec.ExecutionID)
 			ctx.Header("X-Run-ID", exec.RunID)
-			ctx.JSON(http.StatusBadGateway, response)
+			ctx.JSON(httpStatusForFailedExecution(exec), response)
 			return
 		}
 
@@ -942,6 +947,16 @@ func (c *executionController) handleStatusUpdate(ctx *gin.Context) {
 
 		current.Status = normalizedStatus
 		current.StatusReason = req.StatusReason
+		// When the SDK sends an error_status_code in the 4xx range, record it in
+		// StatusReason so the async-completion branch can propagate the correct
+		// HTTP status to the caller (instead of a blanket 502).
+		if req.ErrorStatusCode != nil && *req.ErrorStatusCode >= 400 && *req.ErrorStatusCode < 500 {
+			code := fmt.Sprintf("agent_client_error:%d", *req.ErrorStatusCode)
+			current.StatusReason = &code
+		} else if req.StatusReason == nil && req.ErrorStatusCode != nil && *req.ErrorStatusCode >= 500 {
+			reason := string(ErrorCategoryAgentError)
+			current.StatusReason = &reason
+		}
 		if len(resultBytes) > 0 {
 			current.ResultPayload = json.RawMessage(resultBytes)
 			current.ResultURI = resultURI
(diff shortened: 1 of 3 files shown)


santoshkumarradha (approved):
Validated the control-plane and Python surfaces locally. The status propagation fix and regression coverage look good to me.

PR #946, opened 2026-08-22
fix(sdk): make async teardown exception-safe

### LINKED ISSUE #909: sdk/python: teardown is not exception-safe — an error mid-close can skip resource cleanup permanently
Deferred follow-up from the #907 review (inline thread on `http_connection_manager.py`).

`cancel_and_await_if_same_loop` now (correctly) propagates unrelated task-cleanup exceptions. But its callers treat the whole teardown sequence as if it can't raise, so a propagated error aborts teardown midway:

- `ConnectionManager._close_on_owner_loop`: an escape between `self._closed = True` and the session/connector close leaves the session open — and because `_claim_close_request` sees `_closed`, a retry `close()` is a no-op. The session leaks permanently and unrecoverably (probed during review).
- `async_execution_manager.py` (~`:359`): same shape — an abort skips the remaining task cancels, all state nulling, and active-execution cancellation.
- `result_cache.py` (~`:237`): same shape.

Not reachable today: every SDK background loop breaks on `CancelledError`, so cleanup can't currently raise — this is latent hardening, which is why it was split out of #907 rather than folded in.

Fix shape: wrap the cancel/await steps in `try/finally` so resource close + state nulling always run, and re-raise after. One test per site forcing a cleanup exception and asserting the session/connector still gets closed and state is nulled.

Tethys0:
Fixes #909

### What changed

The Python SDK's async teardown callers assumed `cancel_and_await_if_same_loop()` could not raise. If a task's cancellation cleanup raised an unrelated exception, teardown stopped midway and could leak HTTP resources, leave cache state behind, or skip cancellation of active executions.

This change captures the first cleanup exception, completes the remaining teardown and state cleanup for:

- `ConnectionManager._close_on_owner_loop`
- `AsyncExecutionManager.stop`
- `ResultCache.stop`

The original cleanup exception is re-raised after teardown completes.

### Verification

- Added focused regression coverage for all three teardown sites.
- `RTK_DISABLED=1 /home/lolind/.hermes/hermes-agent/venv/bin/python -m pytest -q -o addopts= tests/test_teardown_exception_resilience.py tests/test_async_lifecycle_deadlock.py tests/test_result_cache_deadlock.py` - 19 passed.
- `ruff check` on all changed Python files - passed.
- `git diff --check` and added-line secret scan - passed.
- The full SDK suite was attempted; unrelated baseline dependency/fixture failures remain outside this change.

### DIFF
--- a/sdk/python/agentfield/async_execution_manager.py
+++ b/sdk/python/agentfield/async_execution_manager.py
@@ -341,12 +341,21 @@ class AsyncExecutionManager:
 
         current_loop = current_running_loop()
         owning_loop = self._loop
+        cleanup_error: Optional[Exception] = None
+
+        def remember_cleanup_error(exc: Exception) -> None:
+            nonlocal cleanup_error
+            if cleanup_error is None:
+                cleanup_error = exc
 
         # Signal shutdown only when we're on the owning loop — asyncio.Event
         # is not safe to set from a foreign loop, and the loop-aware cancel
         # below tears the tasks down regardless.
         if self._shutdown_event is not None and owning_loop is current_loop:
-            self._shutdown_event.set()
+            try:
+                self._shutdown_event.set()
+            except Exception as exc:
+                remember_cleanup_error(exc)
 
         # Cancel background tasks, awaiting only when safe (same loop).
         tasks_to_cancel = [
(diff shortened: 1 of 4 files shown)


santoshkumarradha (approved):
Validated the focused Python teardown regression tests locally. This looks good to me. The CLA is the only remaining blocker, and once that is signed this should be good to go.

AbirAbbas:
Thanks for this @Tethys0 — sorry about the outcome here. #950 landed the same fix for #909 on Monday while this was still waiting on the CLA, so main already has exception-safe teardown at all three sites (released in v0.1.134) and this is now conflicting.

I checked rather than assumed: I copied your `test_teardown_exception_resilience.py` onto current main unchanged and all three tests pass, so nothing you covered is missing.

One thing your version does that main's doesn't, and it's a real if latent gap: in `AsyncExecutionManager.stop` main wraps the whole `for task in tasks_to_cancel` loop in a single `try`, so the first raising cancel breaks out of the loop while the `finally` still nulls all four references — the remaining background tasks stay pending with no handle on them. Your per-iteration guard cancels all four. I confirmed it both ways with a probe test.

Closing this as superseded. We'll pick the residual up ourselves and credit the finding to you.

PR #950, opened 2026-08-23
fix(sdk): make async teardown exception-safe

### LINKED ISSUE #909: sdk/python: teardown is not exception-safe — an error mid-close can skip resource cleanup permanently

mikemikimike:
Fixes #909.

## Root cause

`cancel_and_await_if_same_loop()` intentionally propagates unrelated exceptions raised while cancelled tasks perform cleanup. The three SDK teardown callers did not protect their remaining cleanup with `finally`, so one such exception could leave an HTTP session/connector open, retain manager/cache state, or skip downstream component shutdown.

## Changes

- Keep task-cleanup exceptions observable while always closing the `ConnectionManager` session and connector and clearing its task/loop references.
- Always clear `AsyncExecutionManager` task state, cancel active executions, and stop both child components.
- Always clear `ResultCache` task/loop references and cache contents.
- Add deterministic regression tests for all three teardown paths.

## Validation

- `uv run --frozen pytest sdk/python/tests/test_async_lifecycle_deadlock.py sdk/python/tests/test_result_cache_deadlock.py sdk/python/tests/test_async_execution_manager_paths.py sdk/python/tests/test_connection_manager.py -q` — 65 passed.
- `uv run --frozen ruff format --check ...` — passed.
- `uv run --frozen ruff check ...` — passed.
- `git diff --check` — passed.
- `uv run --frozen mypy --config-file sdk/python/mypy.ini ...` — started but was stopped after prolonged no output on Windows; no result claimed.

The change preserves cross-loop behavior and only changes cleanup sequencing when an unrelated same-loop cleanup exception occurs.


santoshkumarradha:
I reran the focused teardown regression tests locally on Python 3.10 and 3.13, and those passed. I am not approving yet because the required GitHub `lint-and-test (3.10)` job is still red with `Event loop is closed`, so I want that lane green before this goes through.

mikemikimike:
I traced the Python 3.10 lane failure. The focused teardown tests are passing; the job's 9 failures come from the installed LiteLLM 1.98.0 importing 	yping.NotRequired, which is unavailable on Python 3.10. I pushed commit 1102cc81 to cap the SDK dependency at <1.98.0 while retaining the existing 1.97.0 exclusion. This should keep the Python 3.10 dependency compatible; please rerun CI.

santoshkumarradha (approved):
I reran the focused teardown regression suite locally after the LiteLLM cap, and the cleanup paths now stay exception-safe without reopening the Python 3.10 lane. This looks good to me.

PR #954, opened 2026-08-24
fix(telemetry): report each terminal outcome once and stamp usage_context on every event

santoshkumarradha:
## Summary

Investigating an `execution_completed` spike in product metrics (~7k/day baseline → 70k on Aug 22 → 363k on a partial Aug 23) surfaced two defects in the OSS telemetry client that let the event count things that were not executions. This fixes both: a terminal outcome is now reported once per execution, and `usage_context` is stamped on every event so CI traffic can be told apart from real usage.

Neither defect is the whole story of that spike — the count is aggregated by a hosted relay outside this repo — but both are real, both inflate the same metric, and both were live during the window.

**Duplicate lifecycle events were forwarded verbatim.** A terminal status callback re-delivered after a lost 200 used to re-run every side effect in `handleStatusUpdate`, publishing a second `execution_completed` for an execution that had already been reported. [#951](https://github.com/Agent-Field/agentfield/pull/951) closed that path in the handler on Aug 23 and measured it directly ("two duplicate succeeded callbacks and one failed callback emitted three `execution_completed` events"); all three SDKs retry a status callback up to five times, so a single execution could report several. The telemetry client had no defense of its own — it minted a stable `telemetry_event_id` and delegated deduplication entirely to the ingest side, so any republish reached the wire and counted if ingest did not honor the ID.

`handleExecutionEvent` now remembers the terminal outcomes it has reported and drops a repeat. Only stable identities are eligible; the random ones assigned in the `else` branch belong to transitions that are allowed to recur (`timeout → running → timeout`) and collapsing those would lose real events. The set is bounded at 8192 keys with oldest-first eviction — a duplicate arrives within seconds of the original, so eviction can only ever drop a key long past the window in which it could still suppress anything. Keys are built from raw execution IDs and never leave the process; only the existing HMAC is sent.

**`usage_context` rode only on `control_plane_started`.** A CI job starts the control plane on a fresh volume, so it mints a new install ID and its executions are indistinguishable from a real first-time user's. With the context carried by the startup event alone, no downstream query could separate the two after ingestion — which is why the polluted range cannot be cleaned by filtering and has to be trimmed by date instead. [#942](https://github.com/Agent-Field/agentfield/pull/942) disabled telemetry for the functional-test compose stacks, which fixes one known producer; stamping the property on every event makes every producer distinguishable at the source, including ones we have not found yet. `detectUsageContext` already existed and already detected CI — it was simply never applied beyond startup.

`telemetry_schema_version` stays at `2`. The open question above was answered by measurement rather than left open: the relay validates the version strictly (`app/api/oss/telemetry/route.ts:79` accepts only `1|2`, and a parse miss returns 400 `invalid_telemetry_event` before ingest at lines 114-118), so bumping it would have dropped every control-plane event instead of labelling them. Nothing here changes the v2 wire shape — `usage_context` is additive and already allowlisted by the relay's shared property schema, and the suppression is producer-side and invisible to ingest — so the bump bought nothing. See the takeover comment below for the measured accept/reject counts at 2 and at 3.

## Type of change

- [x] Bug fix

## Update, 2026-09-09 (AbirAbbas)

Taken over while this was idle. `git merge origin/main` (no rebase) plus four commits: the schema version stays at 2, a terminal outcome dropped by a full send queue is no longer marked as reported, a falsy `CI` value no longer buckets a real user as CI, and a suppressed duplicate is logged at Warn instead of Debug. Details and the verification runs are in the comment below.

## Test plan

- [x] `cd control-plane && go test ./internal/observability/ -count=1`
- [x] `cd control-plane && go test ./internal/observability/ -count=1 -race` (the suppression set is mutex-guarded)
- [x] `cd control-plane && go test ./internal/observability/... ./internal/handlers/... ./internal/server/... ./internal/events/... -count=1` — all green
- [x] `cd control-plane && go build ./... && go vet ./internal/observability/`

New tests, all in `telemetry_test.go`:

- `TestTelemetryTerminalOutcomeReportedOnce` — four identical completed/failed/cancelled events enqueue one.
- `TestTelemetryTerminalSuppressionIsPerExecution` — two executions completing both report.
- `TestTelemetryNonTerminalEventsAreNotSuppressed` — a repeated `execution_started` is not collapsed.
- `TestTelemetryReportedSetEvictsOldestPastCapacity` — bounded growth, oldest-first eviction.
- `TestTelemetryUsageContextStampedOnEveryEvent` — execution, node, and plain enqueued events all carry it.
- `TestTelemetryDetectsUsageContext` — each CI variable, plus `server` and `dev_or_local`. Clears the ambient CI environment first, or every case would pass on the runner's own `CI=true`.

`TestTelemetryExecutionEventIdentityIsStableAndOpaque` changed rather than being deleted: it used to send the same event twice and read two queue entries, which now blocks because the duplicate is suppressed. It recomputes the identity a republish would carry and asserts it matches the one already sent — the property that actually mattered (ingest can recognize a re-delivery as the same outcome) — and keeps the payload-opacity assertions unchanged.

Existing coverage that pins behavior this PR deliberately does not change: `TestTelemetryTimeoutEventIdentityDoesNotCollapseRepeatedTransitions` and `TestTelemetryExecutionEventIdentityDoesNotCollapseMissingIDs` both still pass.

## Test coverage

- [x] No code was removed, so `coverage-baseline.json` is untouched.

## Notes for review

- **Not changed deliberately:** telemetry still defaults to enabled in CI. Stamping `usage_context` keeps the data and makes it filterable, which seemed strictly better than dropping it; turning collection off by default in CI is a product call, and the opt-out is now documented instead.
- The suppression is in-process only. A control plane restart forgets what it reported, which is correct — the stable `telemetry_event_id` remains the cross-process idempotency mechanism, and this is a second line of defense, not a replacement.
- `telemetryReportedSet`'s zero value is usable and allocates on first `observe`, so a caller constructing `TelemetryService` directly (as every test in the package does) gets the same behavior as production rather than silently getting none.

## Related issues / PRs

- Follows [#951](https://github.com/Agent-Field/agentfield/pull/951) (made terminal status callbacks idempotent — fixed the producer; this hardens the consumer).
- Follows [#942](https://github.com/Agent-Field/agentfield/pull/942) (stopped functional-test runs reporting to production telemetry — fixed one CI producer; this makes all of them identifiable).

<div><a href="https://cursor.com/agents/bc-23040e7d-5d12-445a-8e4d-8736786d45ef?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-23040e7d-5d12-445a-8e4d-8736786d45ef&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


control-plane/internal/observability/telemetry.go
```diff
@@ -357,9 +391,72 @@ func (s *TelemetryService) handleExecutionEvent(event events.ExecutionEvent) {
...
+	// those before — a status callback re-delivered after a lost 200 used to
+	// re-run every side effect, publishing another completed event and
+	// counting the execution again — and it stays one refactor away from
+	// producing them again. Suppressing here keeps a republish from becoming
+	// an inflated count even if the ingest side ignores telemetry_event_id.
+	//
+	// Only stable identities are eligible: the random ones above belong to
+	// transitions that are allowed to recur, and collapsing those would lose
+	// real events.
+	if stable && s.reported.observe(eventName+"\x00"+identityMaterial) {
```
AbirAbbas: `observe` records the key *before* `enqueue` runs, and `enqueue` can drop the event on a full queue (telemetry.go:253-257 -- 256 deep behind a single worker doing network POSTs with a timeout).

When that happens the outcome is marked reported but never sent, and the republish that would have delivered it is now suppressed. Pre-PR that retry got through. So a transient backpressure drop turns into permanent loss of that execution's terminal event, and it under-counts exactly when volume is high -- the opposite failure of the one you're fixing.

Recording only on a successful send would close it: have `enqueue` return whether it queued, and call `observe` in that branch.

AbirAbbas (commented):
Answering the open question in the description: the relay is **strict**, so this can't merge as-is.

In `website2.0`, `app/api/oss/telemetry/route.ts:79`:

```ts
telemetry_schema_version: z.union([z.literal(1), z.literal(2)]).optional(),
```

That's `safeParse` -> `400 invalid_telemetry_event` on failure (route.ts:114-117), and `posthog.capture` is never reached. So a v3 payload isn't degraded or treated as informational, it's dropped whole. There's an explicit regression test pinning it: `route.test.ts:158-167` asserts `telemetry_schema_version: 3` returns 400. Bumping the constant without a relay change silently zeroes out all control-plane telemetry, which is a worse version of the problem this PR is fixing.

The good news is that the `usage_context` half needs no relay work at all. It's already in the shared property schema (route.ts:53, `z.enum(['ci','server','dev_or_local','unknown'])`), and that schema applies to every event name -- there's no per-event discrimination -- so it's accepted and forwarded on `execution_completed` today via the property spread at route.ts:137. `detectUsageContext` only ever returns those three values, so they all land.

Which leaves the version bump buying nothing. Nothing about the wire format actually changed here: `usage_context` is additive and already allowlisted, and the suppression is producer-side and invisible to ingest. I'd keep `telemetrySchemaVersion` at 2 and ship this now, saving the bump for a change that genuinely breaks the v2 shape. If you'd rather bump anyway, the relay change has to land and deploy first.

One real bug on the suppression ordering, left inline.

Two smaller things. `detectUsageContext` treats any non-empty value as CI (telemetry.go:187-190), so `CI=false` -- a real convention people use to opt out of CI behavior -- buckets a genuine user's traffic as `ci` and filters it out of the product numbers. Checking for a falsy value would avoid discarding real data.

And on the belt-and-braces framing: for stable identities (the only ones eligible for suppression) `EventID` is always set, and the relay already derives a deterministic PostHog UUID from it (route.ts:97-104 and 127), so ingest already collapses republishes. This layer doesn't add dedupe power the relay lacks -- what it adds is that a republish stops being observable. If a producer bug starts re-emitting terminal events, the counts stay right and the bug goes silent. The only trace is a `Debug` log; a counter or a `Warn` would keep that signal alive.

Gates on my end: `go build`, `go vet`, `-race` on `./internal/observability/`, and `./internal/handlers/... ./internal/server/... ./internal/events/...` all clean. The two `execute_agent_restart_test.go` failures I hit on the first pass were load-induced flake -- they pass in isolation and on a clean re-run, and on main.

AbirAbbas:
Picked this up since it had been idle since 08-27. Four commits on top of yours, and a `git merge origin/main` rather than a rebase, so your history is untouched.

**Schema version stays at 2** (4a959202). I re-checked the relay today and nothing has moved: `app/api/oss/telemetry/route.ts:79` is still `z.union([z.literal(1), z.literal(2)])`, a `safeParse` miss returns 400 `invalid_telemetry_event` at lines 114-118 before `posthog.capture` is reached, and `route.test.ts:161` still pins version 3 to 400. Rather than argue it from the source, I measured it: a control plane built from this branch, pointed at a local stub implementing that same schema, running three real executions.

| `telemetrySchemaVersion` | accepted | rejected |
| --- | ---: | ---: |
| 3 | 0 | 8 |
| 2 | 8 | 0 |

At 3 every payload is dropped whole -- `control_plane_started`, `node_registered`, all three `execution_started` and all three `execution_completed`. The `usage_context` half needed no relay change, as expected: at 2 the stub accepted all 8 with the property present.

The test that pinned the version compared the constant against itself (`event.SchemaVersion != telemetrySchemaVersion`), so it would have passed just as happily at 3. It now pins the literal `2`.

**A backpressure drop no longer suppresses the republish** (8f79576f) -- details on the inline thread.

Two smaller things from the same review:

**`CI=false` no longer buckets a real user as CI** (6bcdc887). `detectUsageContext` treated any non-empty value as CI. Now that `usage_context` rides on every event instead of only `control_plane_started`, that mislabels *all* of a user's traffic rather than one startup event, and it is exactly the data this PR exists to keep. Empty, `0`, `false`, `no` and `off` (trimmed, case-insensitive) are no longer CI.

**A suppressed duplicate is visible** (f48fe5e4). Debug -> Warn, carrying the event name only -- never the execution ID or the dedupe key. If a producer bug starts re-emitting terminal events, the counts now stay right and this line is the only thing that says why. A clean three-execution run logs it zero times, so it is not noise.

Verification beyond the unit tests: a real control plane with a registered agent and three real executions, telemetry pointed at the stub. 8 events accepted, 0 rejected, every one `telemetry_schema_version: 2` with `usage_context: dev_or_local`, exactly one `execution_completed` per execution with three distinct `telemetry_event_id`s. Separately, on the real event bus: the same terminal outcome published four times reaches the relay once.

One note for anyone reading the description: the duplicate-callback path it describes is already closed at the handler by #951 -- replaying the same terminal status callback returns 200 twice and publishes nothing the second time, which I confirmed in the same run. So this layer is genuinely a second line of defence against a future producer bug rather than something catching duplicates today. That is worth saying out loud because it is also why the Warn matters: without it, this change would make such a bug silent.

Gates at f48fe5e4: `go build ./...`, `go vet ./...`, `-race` on `./internal/observability/`, the full control-plane suite, the web-ui `npm ci && npm run build`, and both coverage gates (control-plane 87.80%, +0.40 pp; patch coverage 93.00% on 91 touched lines).

Out of draft and assigned to you -- I am not the author, so it needs your read before it goes in.

control-plane/internal/observability/telemetry.go
AbirAbbas: Fixed in 8f79576f.

`enqueue` now reports whether the event actually reached the queue, and `handleExecutionEvent` releases the key when it did not:

```go
key := eventName + "\x00" + identityMaterial
if stable && s.reported.observe(key) {
	...
	return
}
if !s.enqueue(eventName, props, identityMaterial) && stable {
	s.reported.forget(key)
}
```

I kept `observe` ahead of the enqueue rather than moving it into the success branch. It is an atomic test-and-set, so running it first is what bounds concurrent republishes to one report; releasing on a drop is what restores the pre-PR property that a retry still gets through. Same outcome you asked for, without opening a window where two republishes both queue.

`telemetryReportedSet.seen` now maps key -> ring slot so `forget` can clear both representations. Without that, a released key leaves a stale entry in `order` that later evicts a freshly re-observed key ahead of its time.

Checked on the wire, not only in a unit test: a real `TelemetryService` (real worker goroutine, real HTTP sender) against a local stub that stalls every response. With the worker parked on one POST the 256-deep queue fills (`queue is full (256/256)`); a terminal event published at that point is dropped and never arrives. Release the stub, let the backlog drain, publish the same terminal event again -- it arrives, exactly once, carrying the identity ingest would dedupe on. On the previous commit that republish delivered nothing at all.

Unit cover is `TestTelemetryTerminalEventDroppedByFullQueueIsRetriable`. Reverting just this hunk fails it with `retried event enqueued 0 times, want 1`.

AbirAbbas (approved):
Both blockers from my 08-27 review are closed and I verified the result myself at f48fe5e4 rather than reading the diff.

Schema version: measured against a stub implementing the relay's schema — 0/8 payloads accepted at version 3, 8/8 at version 2. The relay source is unchanged today (only `1|2` accepted, 400 before ingest).

Backpressure: real service, real worker, real HTTP; queue filled to 256/256 behind a stalled POST, terminal event dropped and confirmed absent, then delivered exactly once on republish after the drain. Reverting only that hunk fails `TestTelemetryTerminalEventDroppedByFullQueueIsRetriable` with `retried event enqueued 0 times, want 1`, and reverting the suppression itself fails `TestTelemetryTerminalOutcomeReportedOnce` with `enqueued 4 times, want 1` — the tests fail for the right reasons.

End to end: real control plane, registered agent, three real executions — 8 events accepted, 0 rejected, all `telemetry_schema_version: 2` with `usage_context: dev_or_local`, one `execution_completed` per execution across three distinct event IDs, and zero suppression warnings on a clean run.

Local gates at this SHA: gofmt on touched files, web-ui `npm ci && npm run build`, `go build ./...`, `go vet ./...`, the full control-plane suite under `-tags sqlite_fts5` with no failures, `-race` on `./internal/observability/`, and the patch-coverage gate (93.00% on 91 touched lines). CI is green on all 26 checks including `coverage-summary` and `license/cla`.

Approving as reviewer, not author — Santosh should still get a look before it merges.

PR #958, opened 2026-08-25
fix(sdk/go): propagate caller context through MemoryBackend interface

### LINKED ISSUE #433: [Go SDK] ControlPlaneMemoryBackend ignores caller context cancellation
## Summary

All HTTP methods in `ControlPlaneMemoryBackend` use `http.NewRequest` instead of `http.NewRequestWithContext`, so a cancelled execution context cannot cancel in-flight memory I/O, causing 15-second hangs on timeout.

## Context

`control_plane_memory_backend.go` at lines 59, 88, 125, and 153 construct HTTP requests without a context. The `MemoryBackend` interface itself also lacks `context.Context` parameters. When an agent execution is cancelled (user abort, workflow timeout, health-check failure), the in-flight `Set`/`Get`/`Delete`/`List` HTTP calls to the control plane will continue running until the 15s `http.Client` timeout fires. Under load, this means cancelled executions hold HTTP connections and goroutines for up to 15 seconds after they should have been cleaned up.

### In Scope

- Add `context.Context` as the first parameter to all `MemoryBackend` interface methods (`Set`, `Get`, `Delete`, `List`, `SetVector`, `GetVector`, etc.).
- Update all interface implementations (`InMemoryBackend`, `ControlPlaneMemoryBackend`, any test mocks) to accept and propagate the context.
- Use `http.NewRequestWithContext(ctx, ...)` in `ControlPlaneMemoryBackend` for all four HTTP methods.
- Update all callers in the agent SDK to pass the execution context through to memory operations.

### Out of Scope

- Changing the HTTP client timeout value — that is a separate tuning concern.
- Implementing retry logic on context cancellation.
- Fixing the deep-copy race in `InMemoryBackend` — tracked separately in G1.

## Files

- `sdk/go/agent/memory.go` — add `ctx context.Context` to `MemoryBackend` interface and `InMemoryBackend` implementation
- `sdk/go/agent/control_plane_memory_backend.go:59,88,125,153` — switch to `http.NewRequestWithContext`
- `sdk/go/agent/agent.go` — update callers that invoke memory backend methods to pass execution context
- `sdk/go/agent/memory_test.go` — add test: cancelling context before `Get` causes the call to return `context.Canceled`

## Acceptance Criteria

- [ ] `MemoryBackend` interface methods all accept `context.Context` as first parameter
- [ ] `ControlPlaneMemoryBackend` uses `http.NewRequestWithContext` for all requests
- [ ] Cancelling the context passed to a memory operation causes it to return promptly with `context.Canceled` or `context.DeadlineExceeded`
- [ ] All existing memory backend tests pass with the updated interface

## Notes for Contributors

**Severity**: HIGH

This is an interface-breaking change — all callers and mocks must be updated. Search for `MemoryBackend` usages across `sdk/go/` with `grep -r 'MemoryBackend' sdk/go/` to find all sites. The `InMemoryBackend` can accept the context but ignore it (operations are in-memory and non-blocking); the context parameter is there for interface consistency and future use.

7vignesh:
## Summary

Fixes #433. All HTTP methods in `ControlPlaneMemoryBackend` used `http.NewRequest` (without context), so a cancelled execution context could not cancel in-flight memory I/O. This caused 15-second hangs on timeout under load when cancelled executions held HTTP connections and goroutines until the `http.Client` timeout fired.

## Type of change

- [x] Bug fix
- [x] Breaking change

### Interface (`memory.go`)
- Added `context.Context` as the first parameter to all 8 `MemoryBackend` interface methods: `Set`, `Get`, `Delete`, `List`, `SetVector`, `GetVector`, `SearchVector`, `DeleteVector`

### ControlPlaneMemoryBackend (`control_plane_memory_backend.go`)
- All `http.NewRequest` calls replaced with `http.NewRequestWithContext(ctx, ...)`
- Cancelled contexts now abort HTTP calls immediately instead of waiting 15s

### InMemoryBackend (`memory.go`)
- Accepts `context.Context` (ignored, since in-memory ops are instant and non-blocking)
- Interface consistency for callers that pass context through

### Memory / ScopedMemory (`memory.go`)
- All backend calls now pass the caller's `ctx` through (already available from method signatures)

### Breaking change note
Any external `MemoryBackend` implementations must add `context.Context` as the first parameter to all methods. The `InMemoryBackend` pattern (accept and ignore) is the recommended approach for backends that don't do I/O.

## Test plan

- [x] `cd sdk/go && go test ./agent/ -run "Memory|ControlPlane" -v -timeout 60s` (all pass)
- [x] `cd sdk/go && go test ./agent/ -run TestControlPlaneMemoryBackend_ContextCancellation -v` (5 cases: 4 methods return context.Canceled, 1 deadline returns promptly)
- [x] `cd sdk/go && go build ./...`
- [x] `cd sdk/go && go vet ./...`

## Test coverage

- [x] If I removed code, I updated `coverage-baseline.json` in this PR *only if* the removal caused a legitimate regression.

## Checklist

- [x] I have linked any related issues (Fixes #433).

## Related issues / PRs

Fixes #433

## Notes

- `TestWithEventTrigger_StampsCallerFileAndLine` is a pre-existing failure on Windows (drive letter in path confuses the split assertion). Unrelated to this change.

### DIFF
--- a/sdk/go/agent/control_plane_memory_backend.go
+++ b/sdk/go/agent/control_plane_memory_backend.go
@@ -2,6 +2,7 @@ package agent
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -45,7 +46,7 @@ type memoryAPIResponse struct {
 	UpdatedAt string `json:"updated_at"`
 }
 
-func (b *ControlPlaneMemoryBackend) Set(scope MemoryScope, scopeID, key string, value any) error {
+func (b *ControlPlaneMemoryBackend) Set(ctx context.Context, scope MemoryScope, scopeID, key string, value any) error {
 	endpoint, err := url.JoinPath(b.baseURL, "/api/v1/memory/set")
 	if err != nil {
 		return err
@@ -60,7 +61,7 @@ func (b *ControlPlaneMemoryBackend) Set(scope MemoryScope, scopeID, key string, 
 	if err != nil {
 		return err
 	}
-	req, err := http.NewRequest(http.MethodPost, endpoint, reader)
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, endpoint, reader)
 	if err != nil {
 		return err
 	}
@@ -79,13 +80,13 @@ func (b *ControlPlaneMemoryBackend) Set(scope MemoryScope, scopeID, key string, 
 	return nil
 }
 
-func (b *ControlPlaneMemoryBackend) Get(scope MemoryScope, scopeID, key string) (any, bool, error) {
+func (b *ControlPlaneMemoryBackend) Get(ctx context.Context, scope MemoryScope, scopeID, key string) (any, bool, error) {
 	endpoint, err := url.JoinPath(b.baseURL, "/api/v1/memory/get")
 	if err != nil {
 		return nil, false, err
 	}
 
-	req, err := http.NewRequest(http.MethodPost, endpoint, keyScopeReader(key, b.apiScope(scope)))
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, endpoint, keyScopeReader(key, b.apiScope(scope)))
 	if err != nil {
 		return nil, false, err
 	}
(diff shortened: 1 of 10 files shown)


santoshkumarradha (approved):
I revalidated the context-cancellation path locally. Threading the caller context through the backend interface and using request-scoped HTTP contexts fixes the timeout hang cleanly.

PR #962, opened 2026-08-26
fix(openrouter): auto-switch stream=false for non-pcm16 audio formats…

### LINKED ISSUE #584: OpenRouterProvider.generate_audio rejects format='wav' under hardcoded stream=true

fcy222fcy:
… (#584)

- OpenRouter SSE streaming only emits pcm16 audio deltas; requesting mp3/flac/opus with stream=True returns a format-related error.
- Add OpenRouterProvider._nonstream_openrouter_audio helper that calls chat/completions with stream=False and parses the JSON response body (choices[0].message.audio.{data, transcript}).
- generate_audio chat-completions path now selects streaming vs non-streaming based on the requested wire format: pcm16 -> SSE; anything else -> JSON.
- Existing SSE tests updated to use format=pcm16.
- New TestOpenRouterNonStreamAudio covers mp3/flac, HTTP errors, invalid JSON, empty choices, and pcm16 staying on the SSE path (13 tests total, all pass).

Closes: #584

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs only
- [ ] Tests only
- [ ] CI / tooling
- [ ] Breaking change

## Test plan

- [ ]
- [ ]

## Test coverage

  **85%**.

- [ ] I ran tests for the surface(s) I changed locally.
- [ ] If I removed code, I updated `coverage-baseline.json` in

> the PR summary.

## Checklist

- [ ] I have read [CONTRIBUTING.md](CONTRIBUTING.md) (if present) and
- [ ] I have linked any related issues (Fixes #123 / Closes #456).

### DIFF
--- a/sdk/python/agentfield/media_providers.py
+++ b/sdk/python/agentfield/media_providers.py
@@ -2667,6 +2667,87 @@ class OpenRouterProvider(MediaProvider):
         transcript = "".join(transcript_parts)
         return b64_full, transcript
 
+    async def _nonstream_openrouter_audio(
+        self,
+        payload: Dict[str, Any],
+        headers: Dict[str, str],
+        *,
+        timeout: float = 300.0,
+        label: str = "audio",
+    ) -> tuple:
+        """
+        Non-streaming helper for OpenRouter audio (chat-completions) requests.
+
+        Used when the requested wire format is *not* ``pcm16``, because the
+        OpenRouter SSE streaming endpoint only emits ``pcm16`` audio deltas
+        and will reject other formats (mp3 / flac / opus) with a 400.
+
+        The non-streaming response is a single JSON document whose shape
+        mirrors the last SSE event::
+
+            {"choices": [{"message": {"audio": {"data": "...",
+                                                  "transcript": "..."}}]}]}
+
+        Args:
+            payload: JSON body for the chat completions request (``stream``
+                must already be set to ``False`` by the caller).
+            headers: HTTP headers including Authorization
+            timeout: Total request timeout in seconds (default 300)
+            label: Label for error messages ("audio" or "music")
+
+        Returns:
+            Tuple of ``(b64_data: str, transcript: str)``
+        """
+        import json as json_mod
+
+        import aiohttp
+
+        client_timeout = aiohttp.ClientTimeout(total=timeout)
+
+        async with aiohttp.ClientSession(timeout=client_timeout) as session:
+            async with session.post(
+                "https://openrouter.ai/api/v1/chat/completions",
+                json=payload,
+                headers=headers,
+            ) as resp:
+                if resp.status != 200:
+                    body = await resp.text()
+                    raise RuntimeError(
+                        f"OpenRouter {label} request failed ({resp.status}): "
+                        f"{body[:500]}"
+                    )
+                raw = await resp.read()
+
+        try:
+            doc = json_mod.loads(raw.decode("utf-8"))
+        except json_mod.JSONDecodeError as exc:
+            raise RuntimeError(
+                f"OpenRouter {label} non-stream response was not valid JSON: "
+                f"{exc}"
+            ) from exc
+
+        choices = doc.get("choices") or []
+        if not choices:
+            return "", ""
+
+        # OpenRouter non-stream response places the final audio+transcript
+        # under choices[0].message.audio — mirroring the SSE delta schema but
+        # using the full "message" instead of a "delta".
+        message = choices[0].get("message") or {}
+        audio = message.get("audio") or {}
+        b64_data = audio.get("data") or ""
+        transcript = audio.get("transcript") or ""
+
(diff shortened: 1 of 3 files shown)


sdk/python/agentfield/media_providers.py
```diff
@@ -2779,16 +2860,26 @@ async def generate_audio(
         messages = [{"role": "user", "content": text}]
         if system is not None:
             messages.insert(0, {"role": "system", "content": system})
+        # OpenRouter SSE streaming only supports pcm16 wire format — any
+        # other format (mp3 / flac / opus) must use the non-streaming JSON
+        # response path, otherwise OpenRouter rejects the request with a
+        # format-related error (see: Issue #584).
+        use_stream = wire_format == "pcm16"
```
santoshkumarradha, line 2867: This routing change looks right, but it also means the existing SSE integration tests can no longer ask for `mp3`. Right now `tests/test_media_integration.py::TestOpenRouterAudioE2E::test_audio_sse_stream_to_audio_output` and `...test_audio_sse_includes_optional_system_message` still do that, so the required Python jobs go red once the fork workflows are approved. Please update those tests to use `pcm16` for the SSE path, or add a separate non-stream assertion for the new branch.

santoshkumarradha (dismissed):
Thanks for digging into this. I validated the new non-stream audio branch locally with `uv run --project sdk/python --extra dev pytest sdk/python/tests/test_openrouter_audio.py -q`, but after I approved the fork workflows the required Python SDK CI and coverage jobs failed. The concrete blocker is that two integration tests still expect `format="mp3"` to stay on the SSE path: `tests/test_media_integration.py::TestOpenRouterAudioE2E::test_audio_sse_stream_to_audio_output` and `tests/test_media_integration.py::TestOpenRouterAudioE2E::test_audio_sse_includes_optional_system_message`. Please update those to exercise SSE with `pcm16`, or add separate assertions for the new non-stream branch, then rerun CI.

sdk/python/agentfield/media_providers.py
fcy222fcy: Updated both SSE integration cases to use format="pcm16" in commit 7564c285. The existing mp3/flac non-stream coverage remains in test_openrouter_audio.py.

AbirAbbas:
Thanks for the quick turnaround on the pcm16 switch. One assertion was left behind (`test_media_integration.py:447` still expected `mp3`), so I pushed a small commit to your branch that pins both SSE integration cases to `pcm16` and asserts `stream is True` on the sent payload — that way they're explicitly tied to the streaming path rather than leaving the transport implicit. Full suite is green on 3.10 locally with ruff 0.15.22; CI re-approved.

Two non-blocking notes for whenever you touch this next: `raw.decode("utf-8")` in `_nonstream_openrouter_audio` is strict and only `JSONDecodeError` is caught, so a non-UTF8 body escapes as a bare `UnicodeDecodeError`; and `await resp.read()` buffers the whole body before the `MAX_AUDIO_B64_BYTES` check, whereas the streaming helper aborts mid-stream at the cap.

Separately, `generate_music` and the Go/TS SDKs carry the same hardcoded `stream: true`; we're handling those in follow-up PRs so this can close #584 cleanly.

AbirAbbas (approved):
Verified locally on the merged tree (py3.10 + ruff 0.15.22 full suite green) and re-checked that wav/pcm16 behaviour is byte-identical to main. Merging.

PR #963, opened 2026-08-26
fix(sdk/python): stream all OpenRouter music requests and label audio by its real container

AbirAbbas:
## Summary

The first version of this PR generalised #962's format-driven transport routing from
`generate_audio` to `generate_music`. A verification round against the **live** OpenRouter
API showed that premise does not hold for music models, and that both halves of the routing
were wrong there:

- **Music models require `stream: true` for *any* audio output.** A non-streaming request is
  answered with `400 {"error":{"message":"Audio output requires stream: true"}}`, so routing
  `mp3`/`flac`/`opus` through the non-streaming JSON path turned a working
  `generate_music(format="mp3")` into a hard failure.
- **Music models ignore `audio.format` entirely.** `google/lyria-3-*` returns MP3 whatever is
  asked for. Requesting `pcm16` on behalf of a `wav` caller and then wrapping the answer in a
  RIFF/PCM16 header produced a `.wav` file with an MP3 inside it — `ffprobe` reports
  `[mp3float] Header missing` on the result.

So this PR now does something different from where it started:

**1. `generate_music` always streams, and labels the result by what actually came back.**
The requested `format` is forwarded as `audio.format` untouched (harmless — the model ignores
it), the transport is always SSE, and the container is sniffed from the decoded bytes:
`ID3` / an MPEG frame sync → `mp3`, `RIFF`+`WAVE` → `wav`, `fLaC` → `flac`, `OggS` → `ogg`.
A payload with no container signature is raw PCM16 and is wrapped in a 24 kHz mono RIFF
container only when `wav` was requested; otherwise it passes through untouched. When the
detected container differs from the requested one, exactly one warning naming both is logged
and `response.audio.format` reports the container that is really there.

**2. The default music model id is fixed.** `google/lyria-3-pro` is not a valid OpenRouter
model ID (`400 "… is not a valid model ID"`), so `generate_music()` with no `model` argument
fails on `main` today. The live id is `google/lyria-3-pro-preview`. The `48kHz stereo` claim
in the docstrings is also wrong (lyria returns 44.1 kHz stereo MP3) and is replaced by
"whatever the model produced, reported on `.audio.format`".

**3. `generate_audio` fails fast instead of spending a request on a guaranteed 400.**
The gateway rejects `stream: false` for *any* audio-output chat request — the 400 comes back
with `provider_name: null`, i.e. from the gateway rather than a model — and the streaming
transport only ever emits `pcm16` deltas. There is therefore no combination that returns
mp3/flac/opus from a chat-audio model, and `_nonstream_openrouter_audio` (added by #962)
could never succeed. `generate_audio` now raises `ValueError` for any chat-audio format other
than `pcm16`/`wav` **before** the request is sent, and `_nonstream_openrouter_audio` is
removed along with its tests. The `/audio/speech` (TTS) route is untouched and still serves
mp3/flac/opus/aac natively; the wav-via-pcm16 behaviour is unchanged.

Everything else about `generate_music` is unchanged — prompt/duration-hint message shape,
`openrouter/` prefix stripping, duration validation, and the returned `MultimodalResponse`
fields — and there are tests pinning that.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs only
- [ ] Tests only
- [ ] CI / tooling
- [x] Breaking change

> Breaking-change note: `generate_audio(format="mp3"|"flac"|"opus"|"pcm")` against a
> **chat-audio** model now raises `ValueError` instead of issuing a request. That request
> could only ever return a 400, so the change turns a confusing runtime error into an
> actionable one — but it is a different exception type, hence the flag. TTS models on
> `/audio/speech` are unaffected. Also: `generate_music(...).audio.format` now reports the
> container the model actually returned (today `"mp3"` for `google/lyria-3-*`, whatever was
> requested) instead of echoing the requested format, and raw PCM that is not wrapped as wav
> is labelled `"pcm16"`; code branching on `.audio.format == "wav"` will see the change.

## Validation contract

Behaviours this change must exhibit, and the test each maps to. Music tests live in
`sdk/python/tests/test_openrouter_audio.py::TestOpenRouterMusicTransportAndLabelling`,
sniffing units in `TestSniffAudioContainer` / `TestLabelStreamedAudio`, and the
`generate_audio` guard in `TestOpenRouterChatAudioFormatGuard`.

| # | Behaviour | Test |
|---|---|---|
| M1 | Every music request is POSTed to `/chat/completions` with `stream: true`, for every requested format (default, wav, pcm16, mp3, flac, opus). No music request ever sets `stream: false`. | `test_every_format_streams_with_the_requested_wire_format[...]`, `test_http_error_mentions_music_and_status` |
| M2 | The `audio.format` on the wire is exactly what the caller requested — no pcm16 substitution. | `test_every_format_streams_with_the_requested_wire_format[...]` |
| M3 | With no `model` the wire model is `google/lyria-3-pro-preview`; an `openrouter/` prefix is still stripped from an explicit model. | `test_openrouter_generate_music_default_model`, `test_openrouter_generate_music_strips_prefix` |
| M4 | A payload carrying a known container signature is reported with that container and returned byte-for-byte unchanged. | `test_detected_container_wins_over_requested_format[mp3/mpeg-sync/wav/flac/ogg]` |
| M5 | A payload with no container signature is raw PCM16: wrapped as 24 kHz mono RIFF/WAVE when `wav` was requested, returned unchanged with the requested label otherwise. | `test_raw_pcm_with_wav_requested_is_wrapped_at_24khz`, `test_raw_pcm_with_pcm16_requested_is_returned_verbatim`, `test_raw_pcm_with_mp3_requested_passes_through_and_warns` |
| M6 | A detected/requested mismatch logs exactly one warning naming both; agreement and the raw-PCM→wav wrap log nothing. | the three M5 tests, `test_detected_container_wins_over_requested_format`, `test_matching_container_logs_nothing` |
| M7 | Request shape otherwise unchanged (messages with the duration hint, `modalities`, model stripping) and duration validation still precedes the HTTP call. | `test_request_shape_is_identical_across_formats[wav/pcm16/mp3/opus]`, `test_duration_validation_runs_before_any_request` |
| M8 | An empty stream yields `.audio is None` and does not crash. | `test_empty_stream_returns_no_audio` |
| M9 | A non-200 raises `RuntimeError` naming `"music"` and the status code. | `test_http_error_mentions_music_and_status` |
| A1 | A chat-audio model asked for anything but `wav`/`pcm16` raises `ValueError` **before any generation request** (the model-metadata lookup that picks the route still runs), naming the requested format, saying pcm16 is all that route delivers, and mentioning the client-side wav wrap. An unrecognised format still normalises to `wav` rather than raising. | `test_unsupported_format_raises_before_any_request[mp3/flac/opus/pcm]`, `test_unknown_format_still_falls_back_to_wav` |
| A2 | `format="pcm16"` still streams and returns the concatenated base64 verbatim. | `test_pcm16_format_still_uses_sse_streaming` |
| A3 | `format="wav"` still streams as pcm16 and returns a 24 kHz mono RIFF/WAVE container. | `test_wav_streams_as_pcm16_and_returns_riff` |
| A4 | The `/audio/speech` route is untouched — mp3 still goes out as `response_format: mp3`. | `test_audio_speech_passes_speed_and_extra` |

### Tests that were re-pinned, and why

`TestOpenRouterMusicFormatRouting` (added in the first version of this PR) asserted
`stream is False` for mp3/flac/opus. That is exactly the shape the live API rejects, so the
class is replaced rather than adjusted. `TestOpenRouterNonStreamAudio` (from #962) covered a
helper that can never succeed against the real gateway; its cases become the `ValueError`
cases, and its two still-valid cases (pcm16 SSE, wav re-wrap) are kept.
`tests/test_media_integration.py::TestOpenRouterMusicE2E::test_music_generation_returns_audio`
now uses a decodable base64 body and asserts the wire `audio.format` as well as `stream`.

## Verification round

The first version of this PR was verified live and failed; this version was re-verified the
same way. Three layers:

**1. Fake OpenRouter over real HTTP** — a local `ThreadingHTTPServer` speaking real SSE, with
the SDK's `aiohttp` session rewritten only at the hostname, so real sockets, real aiohttp and
the real SSE parser are exercised. The server mimics the gateway: it answers `stream: false`
with the real `400 "Audio output requires stream: true"`, and its payload depends on the
**model**, not on `audio.format` (a lyria-like model always returns ID3-headed bytes; another
model returns raw PCM16). Result on this branch: **0 failing checks / 6 POSTs, all with
`stream: true`**; `generate_audio(format="mp3")` raised `ValueError` and sent **zero** POSTs.
The same harness run against this PR's previous head produces **11 failures**, including the
`400` for mp3/flac/opus and a `RIFF`-wrapped MP3 for the default `wav` call.

**2. Real OpenRouter API, 2 calls**, `google/lyria-3-pro-preview`, prompt
_"a five second upbeat ukulele jingle"_:

| call | `format=` | wire | result |
|---|---|---|---|
| 1 | `mp3` | `stream: true`, `audio.format: mp3` | 1,149,612 B starting `ID3\x03…`, `.audio.format == "mp3"`, no warning. `ffprobe`: `codec_name=mp3 sample_rate=44100 channels=2` |
| 2 | `wav` | `stream: true`, `audio.format: wav` | 703,232 B starting `ID3\x03…` (**not** RIFF-wrapped), `.audio.format == "mp3"`, one warning: `OpenRouter returned mp3 audio for google/lyria-3-pro-preview but format='wav' was requested; labelling the result as 'mp3'.` `ffprobe`: `codec_name=mp3 sample_rate=44100 channels=2` |

Both files decode cleanly. On the previous head, call 1 was a `400` and call 2 produced a file
`ffprobe` refuses (`[mp3float] Header missing`).

**3. Unit tests** for the sniffer itself, including the negative cases (`RIFF` that is not
`WAVE`, short buffers, raw PCM frames).

## Test plan

All commands run from `sdk/python/`.

- `uvx --from ruff==0.15.22 ruff check .` → `All checks passed!`
- `uvx --from ruff==0.15.22 ruff format --check` on every file this PR touches →
  `4 files already formatted` (`agent.py` carries pre-existing repo-wide `ruff format` drift
  that also exists on `main`; the only change to it here is a docstring, so it is left alone)
- `./scripts/run_pytest.sh` (Python 3.12) → `2037 passed, 4 skipped, 38 deselected, 36 warnings in 120.24s`, exit 0; coverage `TOTAL 2002 stmts / 116 miss / 94%`. Also run on Python 3.10 (`uv run --python 3.10 … --no-cov -q`) → exit 0, 0 failures (this project's `-q` run does not print the trailing counts line under 3.10), and the `websockets-compat` job on 3.11 with `websockets==12.0` and `==15.0.1` → `11 passed` on both
- Targeted: `pytest tests/test_openrouter_audio.py tests/test_media_integration.py` →
  96 passed, exit 0

## Test coverage

- [x] New code paths are covered by tests in this PR (no bare additions). Container sniffing,
      the wav wrap, the passthrough branch, the mismatch warning and the `generate_audio`
      guard all have tests.
- [x] If I removed code, I updated `coverage-baseline.json` in this PR *only if* the removal
      caused a legitimate regression and I called it out in the summary above.
      `_nonstream_openrouter_audio` was removed, but `media_providers.py` is not in the
      coverage `--cov` set, so no baseline change is needed.

## Checklist

## Related issues / PRs

Refs #584
Refs #962

### DIFF
--- a/sdk/python/agentfield/agent.py
+++ b/sdk/python/agentfield/agent.py
@@ -4206,7 +4206,7 @@ class Agent(FastAPI):
             ```python
             result = await app.ai_generate_music("upbeat jazz piano solo")
             if result.has_audio:
-                result.audio.save("jazz.wav")
+                result.audio.save(f"jazz.{result.audio.format}")
             ```
         """
         return await self.ai_handler.ai_generate_music(
--- a/sdk/python/agentfield/agent.py
+++ b/sdk/python/agentfield/agent.py
@@ -4189,7 +4189,9 @@ class Agent(FastAPI):
         Generate music from a text prompt.
 
         Routes to a music-capable provider (OpenRouter with models like
-        google/lyria-3-pro). Returns a MultimodalResponse with audio data.
+        google/lyria-3-pro-preview). Returns a MultimodalResponse with audio
+        data; ``.audio.format`` reports the container the model actually
+        produced, which may differ from the requested ``format``.
 
         Args:
             prompt (str): Text description of the music to generate.
--- a/sdk/python/agentfield/agent_ai.py
+++ b/sdk/python/agentfield/agent_ai.py
@@ -2043,12 +2043,14 @@ class AgentAI:
         Generate music from a text prompt.
 
         Routes to a music-capable media provider (currently OpenRouter with
-        models like google/lyria-3-pro). Returns a MultimodalResponse with
-        generated audio data (48kHz stereo).
+        models like google/lyria-3-pro-preview). Returns a MultimodalResponse
+        whose ``.audio.format`` reports the container the model actually
+        produced — today google/lyria-3-* always returns MP3 (44.1 kHz stereo)
+        whatever ``format`` is requested.
 
         Args:
             prompt: Text description of the music to generate
-            model: Music model to use (defaults to "google/lyria-3-pro")
+            model: Music model to use (defaults to "google/lyria-3-pro-preview")
             duration: Duration hint in seconds
             **kwargs: Provider-specific parameters (e.g., format="wav")
 
--- a/sdk/python/agentfield/agent_ai.py
+++ b/sdk/python/agentfield/agent_ai.py
@@ -2060,7 +2060,7 @@ class AgentAI:
         Examples:
             result = await app.ai_generate_music("upbeat jazz piano solo")
             if result.has_audio:
-                result.audio.save("jazz.wav")
+                result.audio.save(f"jazz.{result.audio.format}")
 
             result = await app.ai_generate_music(
                 "calm ambient electronic music",
--- a/sdk/python/agentfield/media_providers.py
+++ b/sdk/python/agentfield/media_providers.py
@@ -2722,8 +2722,7 @@ class OpenRouterProvider(MediaProvider):
             doc = json_mod.loads(raw.decode("utf-8"))
         except json_mod.JSONDecodeError as exc:
             raise RuntimeError(
-                f"OpenRouter {label} non-stream response was not valid JSON: "
-                f"{exc}"
+                f"OpenRouter {label} non-stream response was not valid JSON: {exc}"
             ) from exc
 
         choices = doc.get("choices") or []
@@ -2742,8 +2741,7 @@ class OpenRouterProvider(MediaProvider):
             total_size = len(b64_data)
             if total_size > MAX_AUDIO_B64_BYTES:
                 raise RuntimeError(
-                    f"Audio base64 data exceeded "
-                    f"{MAX_AUDIO_B64_BYTES} byte limit"
+                    f"Audio base64 data exceeded {MAX_AUDIO_B64_BYTES} byte limit"
                 )
 
         return b64_data, transcript
@@ -2969,14 +2967,20 @@ class OpenRouterProvider(MediaProvider):
         """
         Generate music via OpenRouter using a music-capable model.
 
-        Uses SSE streaming chat completions with audio modality, similar to
-        generate_audio but targeting music generation models.
+        Uses chat completions with the audio modality, similar to
+        generate_audio but targeting music generation models. The transport is
+        chosen from the requested format, exactly as in generate_audio:
+        ``wav``/``pcm16`` stream over SSE (``wav`` is requested as ``pcm16`` on
+        the wire and wrapped in a RIFF container client-side), while any other
+        format (mp3 / flac / opus) uses the non-streaming JSON response,
+        because OpenRouter rejects ``stream=true`` for those (Issue #584).
 
         Args:
             prompt: Text description of the music to generate
             model: Music model (defaults to "google/lyria-3-pro")
             duration: Duration hint in seconds (must be >0 and <=600)
-            **kwargs: Additional parameters (timeout overrides default 300s)
+            **kwargs: Additional parameters (``format`` defaults to ``wav``;
+                ``timeout`` overrides the default 300s)
 
         Returns:
             MultimodalResponse with generated audio (48kHz stereo)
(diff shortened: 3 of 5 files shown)


AbirAbbas:
Re-opened to re-trigger CI: the Coverage Summary workflow never started on 0647286c (GitHub-side startup failures on several workflows at 15:06–15:10), so the required gate was stuck pending.

PR #970, opened 2026-08-26
fix(sdk/python): allow skipping container IP detection

### LINKED ISSUE #624: [Python SDK] Unexpected internal endpoints probe on start

mikemikimike:
## Summary

- add an opt-out for container and public IP discovery in restricted deployments
- preserve the existing detection order and default behavior
- add a deterministic regression test proving no network request is made when disabled

Closes #624

## Validation

- `uv run --frozen pytest tests/test_agent_networking.py -q` (15 passed)
- `uv run --frozen ruff check agentfield/agent.py tests/test_agent_networking.py`
- `uv run --frozen ruff format --check agentfield/agent.py tests/test_agent_networking.py`
- `git diff --check`
- Full suite: 11 failures unrelated to this change (Windows shell/encoding assumptions in `test_harness_packaging.py` and `test_run_cli_env.py`); remaining tests passed, 6 skipped.

### DIFF
--- a/sdk/python/agentfield/agent.py
+++ b/sdk/python/agentfield/agent.py
@@ -182,6 +182,14 @@ def _detect_container_ip() -> Optional[str]:
     Returns:
         External IP address if detected, None otherwise
     """
+    if os.getenv("AGENTFIELD_SKIP_IP_DETECTION", "").lower() in {
+        "1",
+        "true",
+        "yes",
+        "on",
+    }:
+        return None
+
     try:
         # Try to get IP from container metadata (works in many hosted environments)
         import requests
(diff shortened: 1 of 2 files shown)


AbirAbbas:
Thanks for jumping on this — and sorry for the collision: #969 went up within seconds of this one and covers the same opt-out plus the case pocesar actually reported in #624 (the probe firing even though a callback URL is already configured — #969 skips detection automatically whenever `callback_url` / `AGENT_CALLBACK_URL` is set, and adds `AGENTFIELD_DISABLE_IP_DETECTION` for containers with no callback URL at all), with a tripwire test that `requests.get` is never reached and a docs entry. Since #969 is already reviewed and queued, I'm closing this one rather than carry two fixes for one issue. One note for next time: about two thirds of this diff is `ruff format` reflow of untouched code in `agent.py` — the repo has pre-existing drift there and CI doesn't gate formatting, so keeping diffs to the hunks you mean to change makes review much faster. Your #947 is in the merge queue, by the way — thanks for that one.

PR #976, opened 2026-08-27
fix(control-plane): complete a synchronous execute from the stored record when no terminal event arrives

AbirAbbas:
## Summary

`POST /api/v1/execute/<node>.<reasoner>` — the README's own "call your agent" curl — can hang for the full 90 s and then return `{"error":"execution timeout after 1m30s"}` even though the execution succeeded and its result is already stored. This makes the synchronous waiter poll the execution record as a fallback so the wait is bounded by the poll interval (500 ms), not by which callback happens to arrive first.

**Pre-existing, not a regression.** `waitForExecutionCompletion` is byte-identical back to v0.1.134 and the hang reproduces with the v0.1.134 SDK against today's control plane; it was found during the post-merge regression sweep of #963–#973 and is unrelated to those changes.

## Mechanism

When an agent answers 202 (async acknowledgment) the sync handler subscribes to the execution event bus and waits for `execution_completed` / `execution_failed`. Two callbacks then race in from the SDK:

1. the `reasoner.completed` **workflow event** (`POST /api/v1/workflow/executions/events`), which `WorkflowExecutionEventHandler` persists as the terminal status — without publishing a lifecycle event;
2. the authoritative **status callback** (`POST /api/v1/executions/<id>/status`), which normally publishes the event the waiter is listening for.

If (1) lands first, (2) is an idempotent terminal→terminal update and the lifecycle event the waiter needs is never published. The existing pre-subscribe store check does not help because the record is still `running` at that instant. Nothing wakes the waiter; it times out 90 s after the result was stored. Any slowdown of the agent's callback ordering (a loaded box, `python -X dev`, a slow agent) exposes it.

## Change

- `waitForExecutionCompletion` now also ticks a store poll (`completionPollInterval`, 500 ms, a package var so tests can shrink it) and returns the stored record as soon as it is terminal. The event bus stays the fast path; nothing else changes — no new events, no double webhooks.
- Test `TestWaitForExecutionCompletion_TerminalStatusWithoutEvent` stores a terminal status and publishes **no** event; the waiter must return within the poll interval. Red with the poll disabled (times out at the test deadline), green with it.

## Validation

- Unit: the five `TestWaitForExecutionCompletion_*` tests pass; the new one fails when the poll is neutralised.
- Live: a real Python agent (`python -X dev`, explicit callback URL) against an isolated control plane built from this branch — sync executes return in ~20 ms, and the control-plane log shows the fallback firing (`completing from the stored record`) on an execution whose event never arrived. The hang was observed against the unfixed rc.6 binary in the same harness (1 of ~70 calls on an idle box; the regression sweep's agent hit it 3/3 under load, and 1/1 on the pre-merge commit).
- Gates: `go build`, `go vet`, `go test ./...`, `golangci-lint` (no new findings vs `main`).

## Type of change

- [x] Bug fix
- [ ] Breaking change

Refs the post-merge regression sweep of #963–#973.

### DIFF
--- a/control-plane/internal/handlers/execute.go
+++ b/control-plane/internal/handlers/execute.go
@@ -1357,9 +1357,24 @@ func (c *executionController) publishExecutionStartedEvent(plan *preparedExecuti
 	)
 }
 
+// completionPollInterval is how often waitForExecutionCompletion re-reads the
+// execution record while it waits on the event bus. Var, not const, so tests
+// can shrink it.
+var completionPollInterval = 500 * time.Millisecond
+
 // waitForExecutionCompletion waits for an execution to complete by subscribing to the event bus.
 // It returns the completed execution record or an error if the execution fails or times out.
 // This is used when agents return HTTP 202 (async acknowledgment) but the sync endpoint needs to wait for completion.
+//
+// The event bus is the fast path, but not every writer of a terminal status
+// publishes a lifecycle event: the SDK's reasoner.completed workflow event
+// persists the terminal state through WorkflowExecutionEventHandler without
+// one, and when the authoritative /status callback lands afterwards it is an
+// idempotent terminal->terminal update. If that ordering wins the race, the
+// only signal a synchronous caller would ever get was the timeout, ninety
+// seconds after its result had been stored. The store is therefore polled as
+// a fallback, so the wait is bounded by completionPollInterval rather than by
+// whichever callback happened to arrive first.
 func (c *executionController) waitForExecutionCompletion(ctx context.Context, executionID string, timeout time.Duration) (*types.Execution, error) {
 	if c.eventBus == nil {
 		return nil, fmt.Errorf("event bus not available")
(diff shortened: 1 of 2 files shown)


AbirAbbas:
Follow-up from the post-merge sweep's second reviewer, which sharpened the picture after this merged:

- The trigger is not load or `python -X dev`: **any reasoner that awaits before returning** (`asyncio.sleep(0.3)` was enough, 12/12) hung the synchronous endpoint on main, while instant reasoners passed. Re-verified with a 2 s reasoner: unfixed rc.6 → 90 s timeout 2/2; **v0.1.134 (last stable, CP + SDK) → the same 90 s timeout**, so this predates the current release train; with this fix → 2.5 s with the result, the store-poll fallback carrying both calls.
- The mechanism note in the description is partly wrong: in the failing traces the `/status` callback lands *after* both workflow events and does write `executions.status=succeeded` — the terminal event simply never reaches the waiter. The poll fallback covers that regardless of which writer is at fault.
- One residue this fix does not address, filed as #978: the `workflow_executions` row stays `running` for awaiting reasoners on both the sync and async routes (also on v0.1.134), so run views show them as permanently running.

PR #977, opened 2026-08-27
test(ts-sdk): cover AgentFieldClient restart paths

mikemikimike:
## Summary
- add regression coverage for restartExecution input validation
- verify structured control-plane errors preserve the HTTP status and encoded execution path

## Validation
- npm test -- --run tests/agentfield_client.test.ts (12 passed)
- npm run lint
- git diff --check
- Full SDK suite: 893 passed, 1 pre-existing Windows path assertion failure in tests/harness_runner.test.ts


AbirAbbas (approved):
Ran the CI-exact gates locally on Node 20 (`npm ci` / `npm run lint` / `npm test`) and the whole suite is green — 894 passed, and the Windows path assertion in `harness_runner.test.ts` doesn't reproduce on Linux. I also mutation-checked both new tests against `AgentFieldClient.restartExecution`: removing the `executionId` guard, dropping the status from the error message, and dropping `encodeURIComponent` from the restart path each make them fail, so they're pinned to real behavior rather than to the implementation's shape. One non-blocking note: these overlap heavily with the existing restartExecution cases in `tests/agentfield_client_coverage.test.ts` — the genuinely new assertion is that no request is made when validation rejects — but that's cheap duplication and not worth holding this up.

PR #979, opened 2026-08-27
fix(control-plane): mirror workflow execution events

### LINKED ISSUE #978: workflow_executions row stays 'running' after a reasoner that awaits completes (sync and async), since at least v0.1.134
## Symptom

For any reasoner that yields to the event loop before returning (`await asyncio.sleep(0.3)` is enough; an LLM call or HTTP fetch behaves the same), the execution completes and `executions` is correct, but the matching `workflow_executions` row never leaves `running`:

```
executions          {status: succeeded, duration_ms: 2001, completed_at: 2026-08-27 00:42:59}
workflow_executions {status: running,   duration_ms: NULL, completed_at: NULL, output_data: NULL}
```

Reproduced identically on:
- current `main` after #976, via `POST /api/v1/execute/<node>.<reasoner>` (which now returns the result promptly) **and** via `POST /api/v1/execute/async/...` + poll;
- the pre-merge commit 42284143;
- **v0.1.134** (control plane + SDK), so this predates the current release train.

Instant reasoners (return without awaiting) are not affected. The run / workflow views that read `workflow_executions` therefore show these executions as permanently running (observed by the post-merge regression sweep's second reviewer).

## Evidence from one run (control plane at debug, isolated home)

All three SDK callbacks arrive and are accepted, in this order:

```
POST /api/v1/workflow/executions/events   200   (reasoner.started,   t+0s)
POST /api/v1/workflow/executions/events   200   (reasoner.completed, t+2s)
POST /api/v1/executions/<id>/status       200   (succeeded,          t+2s, last)
```

`handleStatusUpdate` calls `updateWorkflowExecutionStatus` and, for terminal states, `updateWorkflowExecutionFinalState` (control-plane/internal/handlers/execute.go ~:1006-1010), both of which go through `store.UpdateWorkflowExecution`. Neither produces a "failed to update workflow execution status" error line, and the storage layer logs no update for that execution — only the original `Successfully inserted new workflow execution ... status=running` from execute time. `updateWorkflowExecutionStatus` swallows any error containing "not found" silently (execute.go ~:1070), so a lookup miss inside `UpdateWorkflowExecution` would leave exactly this state with nothing in the log. For instant reasoners the same callbacks do update the row, so the difference is timing/ordering-related rather than a missing call.

## Repro

```python
# slow_agent.py
import asyncio, os
from agentfield import Agent
app = Agent(node_id="slow-demo", agentfield_server=os.environ["AGENTFIELD_SERVER"], callback_url=os.environ["CB"])
@app.reasoner()
async def slow(seconds: float = 2.0) -> dict:
    await asyncio.sleep(seconds); return {"slept": seconds}
if __name__ == "__main__":
    app.run(host="0.0.0.0", port=int(os.environ["PORT"]), auto_port=False)
```

```
AGENTFIELD_LOG_LEVEL=debug af server                                   # fresh home
AGENTFIELD_SERVER=http://127.0.0.1:8080 CB=http://127.0.0.1:8001 PORT=8001 python slow_agent.py
curl -s -X POST http://127.0.0.1:8080/api/v1/execute/slow-demo.slow -H 'Content-Type: application/json' -d '{"input":{"seconds":2}}'
sqlite3 ~/.agentfield/data/agentfield.db "select execution_id,status,duration_ms from workflow_executions order by rowid desc limit 1"
```

## Related

- #976 fixed the separate symptom on the same path (the synchronous caller hanging 90 s while the result was already stored). This issue is what remains after it.

mikemikimike:
## Summary

Fixes #978.

Workflow execution events updated the primary execution record but did not mirror later events to the `workflow_executions` projection when that row already existed. As a result, executions whose reasoners yielded could remain displayed as `running`.

This change applies the event status, result, duration, completion time, and error to the workflow projection as well, creating the projection when an event-based execution has no row. Existing terminal-state protection is preserved.

## Validation

- Added coverage to `TestWorkflowExecutionEventHandler_CreateAndUpdate` for the terminal workflow projection.
- `gofmt` completed successfully.
- `git diff --check` completed successfully.
- The focused Go test could not build in this environment because the required C compiler (`gcc`) is not installed for the SQLite cgo dependency.

### DIFF
--- a/control-plane/internal/handlers/coverage_handlers_92_target_test.go
+++ b/control-plane/internal/handlers/coverage_handlers_92_target_test.go
@@ -163,14 +163,16 @@ func (s *cancelHandlerErrorStorage) StoreWorkflowExecutionEvent(_ context.Contex
 
 type workflowEventStoreStub struct {
 	storage.StorageProvider
-	exec             *types.Execution
-	getErr           error
-	createErr        error
-	updateErr        error
-	storeWorkflowErr error
-	createdExec      *types.Execution
-	storedWorkflow   *types.WorkflowExecution
-	updatedExecution *types.Execution
+	exec              *types.Execution
+	getErr            error
+	createErr         error
+	updateErr         error
+	storeWorkflowErr  error
+	updateWorkflowErr error
+	workflow          *types.WorkflowExecution
+	createdExec       *types.Execution
+	storedWorkflow    *types.WorkflowExecution
+	updatedExecution  *types.Execution
 }
 
 func (s *workflowEventStoreStub) GetExecutionRecord(_ context.Context, _ string) (*types.Execution, error) {
@@ -206,7 +208,23 @@ func (s *workflowEventStoreStub) CreateExecutionRecord(_ context.Context, execut
 
 func (s *workflowEventStoreStub) StoreWorkflowExecution(_ context.Context, execution *types.WorkflowExecution) error {
 	s.storedWorkflow = execution
+	s.workflow = execution
 	return s.storeWorkflowErr
+}
+
+func (s *workflowEventStoreStub) UpdateWorkflowExecution(_ context.Context, _ string, update func(*types.WorkflowExecution) (*types.WorkflowExecution, error)) error {
+	if s.updateWorkflowErr != nil {
+		return s.updateWorkflowErr
+	}
+	if s.workflow == nil {
+		return errors.New("workflow execution not found")
+	}
+	updated, err := update(s.workflow)
+	if err != nil {
+		return err
+	}
+	s.workflow = updated
+	return nil
 }
 
 func (s *workflowEventStoreStub) UpdateExecutionRecord(_ context.Context, _ string, update func(*types.Execution) (*types.Execution, error)) (*types.Execution, error) {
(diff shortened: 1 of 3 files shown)


AbirAbbas (approved):
Verified this end to end against the repro in #978.

Root cause confirmed by reading main plus a live debug log: for any reasoner that yields, the Python SDK `await`s the `reasoner.completed` POST to `/api/v1/workflow/executions/events` (decorators.py `_send_workflow_completion`) before the handler fires the status callback. So the events handler marks the `executions` row terminal first via `applyEventToExecution`, and the status callback then lands in the terminal-noop branch from #951 (execute.go ~955-963) and returns at `if terminalNoop` — before `updateWorkflowExecutionStatus` / `updateWorkflowExecutionFinalState` ever run. `workflow_executions` therefore never leaves `running`. Instant reasoners are fine only because the status callback wins that race. My debug log shows exactly the ordering in the issue: events, events, status last.

Mirroring in the events handler is the right layer — it's the writer that made the row terminal, and the `existing == nil` branch already stored a projection, so the update branch was simply asymmetric. The status-callback path stays correct for SDKs/paths that never emit events: the executions row isn't terminal when the callback arrives, so there's no noop and the projection still gets updated.

Ran locally (Go 1.25.4, cgo/sqlite available — the bit that couldn't run in your environment):

- `go build ./...`, `go vet ./internal/handlers/`, `go test ./internal/handlers/ -count=1` all green; golangci-lint produces the same set of pre-existing findings as main and nothing in the touched files.
- Anti-tautology check: reverting only `workflow_execution_events.go` to main makes `TestWorkflowExecutionEventHandler_CreateAndUpdate` fail with `expected: succeeded / actual: running`, so the new assertions are load-bearing rather than shaped to the code. The `coverage_handlers_92_target_test.go` edits are necessary too — the stub embeds a nil `StorageProvider`, so it needs a real `UpdateWorkflowExecution` now that the handler calls it; no existing expectations were rewritten.
- Live: built `af` from this branch and from main, each against its own isolated HOME and fresh sqlite, with the `slow_agent.py` from the issue.
  - main — sync slow: `workflow_executions` `running`; async slow: `running`; failing reasoner: `running`, no error stored; instant: `succeeded`. 13 of 14 projection rows disagreed with `executions`.
  - this branch — sync, async, failing (`failed` with `error_message` and duration), instant, and a nested `parent` → 2×`child` run via `agent.call` all land terminal with `duration_ms`, `completed_at` and `output_data` populated. 17/17 projection rows agree with `executions`, and no extra projection rows are created.

Two small things worth a follow-up, neither blocking: `applyEventToWorkflowExecution` clears `ErrorMessage` when a terminal event carries an empty error string, where `applyEventToExecution` only does that for `succeeded`; and the mirror doesn't carry `status_reason`, so a projection written by the event path leaves it null. Both are strictly better than the row being stuck at `running`.

PR #980, opened 2026-08-27
feat(ts-sdk): support multimodal AI request content

### LINKED ISSUE #91: [TypeScript SDK] Add helpers for image, audio, and file inputs/outputs
The Python SDK has convenient wrappers for multimodal content (images, audio, files). Add similar helpers to TypeScript.

### Input helpers

```typescript
import { Image, Audio, File } from '@agentfield/sdk';

const img = await Image.fromFile('./photo.jpg');
const audio = await Audio.fromUrl('https://example.com/audio.mp3');
const file = File.fromBuffer(buffer, 'document.pdf');
```

### Output handlers

```typescript
const response = await ctx.ai("Describe this image", { image: img });
if (response.hasAudio) {
  await response.audio.save('./output.mp3');
}
```

## Reference implementation

- Input: `sdk/python/agentfield/multimodal.py`
- Output: `sdk/python/agentfield/multimodal_response.py`

## Files to create

- `sdk/typescript/src/ai/multimodal.ts`
- `sdk/typescript/src/ai/MultimodalResponse.ts`

---

## Documentation (internal team)

This adds new developer-facing APIs for multimodal AI. Update the SDK documentation to include:
- New classes: Image, Audio, File, MultimodalResponse
- Factory methods and their parameters
- Complete examples for vision and audio use cases
- Supported file formats

mikemikimike:
## Summary

Closes #91.

The TypeScript SDK already exposed `Image`, `Audio`, `Video`, and `File` helpers, but `AIClient.generate` and `AIClient.stream` only accepted a plain string prompt. This change adds an optional `content` request field and translates those helpers into AI SDK user message parts.

## Changes

- Add `content?: MultimodalContent[]` to `AIRequestOptions`.
- Preserve the existing string prompt behavior when content is omitted.
- Pass image, audio, video, and file content through both generation entry points.
- Add deterministic provider-mocked coverage for image and audio requests.

## Validation

- `npm test -- --run tests/ai_client_extra.test.ts` — 7 passed.
- `npm run lint` — passed.
- `npm run build` — passed.
- Full test suite: 892 passed, 1 unrelated Windows path-separator assertion failed in `tests/harness_runner.test.ts`.

### DIFF
--- a/sdk/typescript/src/ai/AIClient.ts
+++ b/sdk/typescript/src/ai/AIClient.ts
@@ -22,6 +22,7 @@ import {
 } from './openrouterAttribution.js';
 import { withOpenRouterUsageInclude } from './openrouterUsage.js';
 import { recordAiSdkUsage } from '../usage/aiUsage.js';
+import type { MultimodalContent } from './multimodal.js';
 
 export type ZodSchema<T> = z.Schema<T, z.ZodTypeDef, any>;
 
@@ -70,6 +71,8 @@ export interface AIRequestOptions {
    * - 'tool': Force tool calling mode
    */
   mode?: 'auto' | 'json' | 'tool';
+  /** Additional image, audio, video, or file parts for the user message. */
+  content?: MultimodalContent[];
 }
 
 export type AIStream = AsyncIterable<string>;
@@ -101,13 +104,14 @@ export class AIClient {
   async generate<T = any>(prompt: string, options: AIRequestOptions = {}): Promise<T | string> {
     const { provider, modelName } = this.resolveModelChoice(options);
     const model = this.buildModel(options);
+    const requestPrompt = this.buildPrompt(prompt, options.content);
 
     if (options.schema) {
       const schema = options.schema;
       const call = async () =>
         generateObject({
           model: model,
-          prompt,
+          prompt: requestPrompt,
           output: 'object',
           system: options.system,
           temperature: options.temperature ?? this.config.temperature,
@@ -124,7 +128,7 @@ export class AIClient {
     const call = async () =>
       generateText({
         model: model,
-        prompt,
+        prompt: requestPrompt,
         system: options.system,
         temperature: options.temperature ?? this.config.temperature,
         maxOutputTokens: options.maxTokens ?? this.config.maxTokens
@@ -143,7 +147,7 @@ export class AIClient {
     const model = this.buildModel(options);
     const streamResult = streamText({
       model: model,
-      prompt,
+      prompt: this.buildPrompt(prompt, options.content),
       system: options.system,
       temperature: options.temperature ?? this.config.temperature,
       maxOutputTokens: options.maxTokens ?? this.config.maxTokens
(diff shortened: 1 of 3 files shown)


sdk/typescript/src/ai/AIClient.ts
```diff
@@ -297,6 +301,24 @@ export class AIClient {
...
+    return [{
+      role: 'user' as const,
+      content: [
+        { type: 'text' as const, text: prompt },
+        ...content.map((part) => {
+          if (part.type === 'text') return { type: 'text' as const, text: part.text };
+          if (part.type === 'image_url') return { type: 'image' as const, image: part.imageUrl.url };
+          if (part.type === 'video_url') return { type: 'file' as const, data: part.videoUrl.url, mimeType: 'video/*' };
+          if (part.type === 'input_audio') return { type: 'file' as const, data: part.audio.data, mimeType: `audio/${part.audio.format}` };
+          return { type: 'file' as const, data: part.file.url, mimeType: part.file.mimeType };
```
AbirAbbas: `mimeType` isn't a field the AI SDK knows about. In v6 a file part is `{ type: 'file', data, mediaType }` and `mediaType` is **required** — `FilePart` in `@ai-sdk/provider-utils`, and `filePartSchema` in `ai/dist/index.mjs` has `mediaType: z4.string()` with no `.optional()`. So every audio, video and file part built here fails prompt validation before the request is ever sent:

```
AI_InvalidPromptError: Invalid prompt: The messages do not match the ModelMessage[] schema.
```

I ran the exact objects this function emits through a real `generateText` against a stub model on the pinned `ai@6.0.149`: image parts are fine, but `Audio`, `Video` and `File` all throw. Renaming the key to `mediaType` makes all three pass. The suite doesn't catch it because `generateText`/`streamText` are fully mocked in `ai_client_extra.test.ts`, so the new assertions lock in `mimeType` instead of flagging it — switching those to `mediaType` would turn it into a real guard.

Two follow-ons in the same map:

- `'video/*'` isn't a real IANA type. For `Video.fromFile/fromBuffer/fromBase64` the data URL carries the true type and the SDK overrides the wildcard, but `Video.fromUrl('…/clip.webm')` passes `video/*` through to the provider verbatim.
- `File.fromUrl(url)` leaves `mimeType` undefined, so even after the rename that part still throws — `mediaType` has no default.

Also worth noting `audio/${part.audio.format}` yields `audio/mp3`, while `AUDIO_MIME_TYPES` in `multimodal.ts` already maps `.mp3` to the canonical `audio/mpeg`; reusing that table keeps the two consistent.

Rough fix shape:

```ts
if (part.type === 'video_url') return { type: 'file' as const, data: part.videoUrl.url, mediaType: mediaTypeFromDataUrl(part.videoUrl.url) ?? 'video/mp4' };
if (part.type === 'input_audio') return { type: 'file' as const, data: part.audio.data, mediaType: AUDIO_MIME_TYPES['.' + part.audio.format] ?? `audio/${part.audio.format}` };
return { type: 'file' as const, data: part.file.url, mediaType: part.file.mimeType ?? mediaTypeFromDataUrl(part.file.url) ?? 'application/octet-stream' };
```

One more thing: the `as any` on the returned array is what let this past `tsc --noEmit`. Typing the return as `ModelMessage[]` instead surfaces it immediately — `Property 'mediaType' is missing in type '{ type: "file"; data: string; mimeType: string; }' but required in type 'FilePart'` — and would keep the mapping honest as the SDK moves.

AbirAbbas: Pushed the fix on top in 57ce8a20 so this can land rather than sitting on a review round: file parts now use `mediaType`, audio/video/file derive real IANA types from the maps already in `multimodal.ts` (mp3 -> `audio/mpeg`, `.webm` -> `video/webm`), `File.fromUrl()` without an explicit mime falls back to `application/octet-stream`, and the messages array is typed as `ModelMessage[]` instead of `as any` so tsc guards the keys from here on. Verified against a stub LanguageModelV3 on the pinned ai@6.0.149 that image, audio, video and file all reach the model in both URL- and data-URL-backed variants, and the tests now assert `mediaType`.

AbirAbbas (approved):
The multimodal translation was emitting `mimeType`, which ai@6.0.149 doesn't accept — file parts need `mediaType` — so audio, video and file content failed prompt validation before any request went out, with only images actually working. I pushed a fix on top rather than send it back for a round trip: `mediaType` throughout, real IANA types derived from the maps already in multimodal.ts instead of the `video/*` wildcard, a sensible fallback when `File.fromUrl()` has no mime, and the messages array typed as `ModelMessage[]` instead of `as any` so tsc catches this class of mistake next time. Verified against a stub LanguageModelV3 on the pinned SDK that image, audio, video and file all reach the model in both URL- and data-URL-backed forms, and the tests now assert `mediaType`.

PR #981, opened 2026-08-27
test(go-sdk): cover Gemini harness provider

mikemikimike:
## Summary

- add focused unit coverage for the Go SDK Gemini harness provider
- verify command construction, permission modes, model variant stripping, working-directory precedence, and error mapping

## Validation

- `go test ./harness -run '^TestGeminiProviderBuildsCommandAndMapsSuccess$|^TestGeminiProviderUsesPlanModeAndCwdPrecedence$|^TestGeminiProviderMapsExecutionErrors$' -count=1 -v -timeout 30s` (passed; shell-based cases are skipped on Windows)
- `git diff --check` (passed)

The tests use the repository's existing POSIX shell fixture helper and therefore skip explicitly on Windows. On POSIX CI they execute the provider through the real subprocess path.

### DIFF
--- a/sdk/go/harness/gemini_test.go
+++ b/sdk/go/harness/gemini_test.go
@@ -1,0 +1,55 @@
+package harness
+
+import (
+	"context"
+	"runtime"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGeminiProviderBuildsCommandAndMapsSuccess(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("shell fixture requires POSIX")
+	}
+	dir := t.TempDir()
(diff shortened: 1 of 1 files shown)


AbirAbbas (approved):
Ran the sdk-go CI gates locally and they're clean: `go mod tidy` leaves no go.mod/go.sum drift, `go build ./...`, `go vet ./harness/`, the full `go test ./... -count=1` for sdk/go, and `./harness -race`. I mutation-tested the new assertions rather than trusting them — renaming `--yolo`, flipping plan mode to `--approval-mode default`, dropping the model-variant strip, swapping the Cwd/ProjectDir precedence, changing the missing-binary FailureCrash, disabling the non-zero-exit error mapping, and removing the NumTurns zeroing each made one of these three tests fail, so they're guarding real behavior. Coverage of `gemini.go` Execute goes 85.7% -> 88.1%; the one net-new statement is the "exited with code N and produced no output" branch (existing tests always write stderr, so they take the other side) — the rest overlaps runner_test.go and model_test.go, which is fine for a redundancy-on-purpose provider test.

PR #992, opened 2026-08-27
fix(python-sdk): include PEP 561 type marker

### LINKED ISSUE #984: [Python SDK] Missing py.typed for the SDK
## Summary

`mypy` 2.3.0 will only follow and resolve Agentfield SDK generic and types when you force it to include the overrides

```toml
[[tool.mypy.overrides]]
module = "agentfield.*"
follow_untyped_imports = true
check_untyped_defs = true
follow_imports = "silent"
disallow_untyped_calls = true # this one even makes app.run() to complain
```

## Proposed solution

Add the `py.typed`, the defacto (PEP 561) for signaling a complete type-safe codebase. 
basedPyright also complains on some API surface coming from AF too.

ump45nose:
## Summary

Publish the Python SDK with the PEP 561 `py.typed` marker so mypy, Pyright, and other type checkers recognize the inline type information. The marker is declared explicitly as setuptools package data to keep it in built wheels.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs only
- [ ] Tests only
- [ ] CI / tooling
- [ ] Breaking change

## Test plan

- [x] Built the current SDK wheel before the change and confirmed `agentfield/py.typed` was absent.
- [x] Rebuilt the wheel after the change and confirmed `agentfield/py.typed` is included.
- [x] Parsed `pyproject.toml` with `tomllib` and ran `git diff --check`.

## Test coverage

- [x] I ran the packaging checks for the Python SDK surface changed locally.
- [x] No runtime code path was added; the built-wheel assertion directly covers the package-data change.
- [x] I did not modify `coverage-baseline.json`.

## Checklist

- [x] I have read the contributing and development guidance.
- [x] The commit is signed off and follows conventional-commits style.

## Related issues / PRs

Fixes #984


AbirAbbas (approved):
Verified the packaging claim locally rather than taking it on faith: built the sdist and wheel from this branch and from main, and `agentfield/py.typed` is present in both artifacts here and absent in both on main. Also confirmed the end-to-end effect — mypy against the main wheel reports `Skipping analyzing "agentfield": ... missing library stubs or py.typed marker` and reveals `Any`, while against this branch's wheel it resolves cleanly, including submodules like `agentfield.harness`. Marker file is empty (not `partial`), which is the right signal for a fully-inline-typed package, and the full CI-exact gates pass on 3.10 (ruff 0.15.22 clean, 2197 passed / 4 skipped).

PR #993, opened 2026-08-27
fix(python-sdk): scope LiteLLM cap to Python 3.10

### LINKED ISSUE #982: [Python SDK] Hard semver range doesn't allow updating LiteLLM >=1.98
## Summary

LiteLLM is very fast paced and they have been breaking things for a while even on patch versions, using it as a client library is calculated risk. 
Right now, the pyproject.toml sets a hard semver range < 1.98.0 that fails to install if you ask for a higher version of `litellm` locally. 
The comment inside the project file says the clamp is because of python 3.10, but every modern python version gets affected.

## Proposed solution

Use python markers to only keep that hard limit on python 3.10 (ie. `; python_version < 3.11`).
Keep the old free range (`"litellm"`) to allow newer versions on new python versions (>3.11 <3.14)

## Alternatives considered

`uv` allows overrides, poetry doesn't, so no real alternative besides local patching.

ump45nose:
## Summary

Apply the LiteLLM `<1.98.0` compatibility cap only on Python 3.10, where the documented `ModelResponse` and `typing.NotRequired` failures occur. Python 3.11–3.13 now receive an unconstrained LiteLLM requirement and can install current releases.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs only
- [ ] Tests only
- [ ] CI / tooling
- [ ] Breaking change

## Test plan

- [x] Built the SDK wheel before the change and confirmed the `<1.98.0` cap had no Python marker.
- [x] Rebuilt the wheel and verified its METADATA contains separate Python `<3.11` and `>=3.11` requirements.
- [x] Evaluated both markers with `packaging`: Python 3.10 selects `!=1.97.0,<1.98.0`; Python 3.11 selects the unconstrained requirement.
- [x] Ran pre-commit for `sdk/python/pyproject.toml` and `git diff --check`.

## Test coverage

- [x] I ran the packaging checks for the Python SDK surface changed locally.
- [x] No runtime code path was added; the built-wheel METADATA assertion directly covers this change.
- [x] I did not modify `coverage-baseline.json`.

## Checklist

- [x] I have read the contributing guidance.
- [x] The commit is signed off and follows conventional-commits style.

## Related issues / PRs

Fixes #982

### DIFF
--- a/sdk/python/pyproject.toml
+++ b/sdk/python/pyproject.toml
@@ -31,7 +31,8 @@ dependencies = [
     "pydantic>=2.0",
     # LiteLLM 1.97.0 crashes on Python 3.10 building ModelResponse, and
     # 1.98.0 imports typing.NotRequired (only available in Python 3.11+).
-    "litellm!=1.97.0,<1.98.0",
+    "litellm!=1.97.0,<1.98.0; python_version < '3.11'",
+    "litellm; python_version >= '3.11'",
     "psutil",
     "PyYAML>=6.0",
     "aiohttp>=3.8",


AbirAbbas (approved):
Built the wheel and confirmed METADATA carries two `Requires-Dist: litellm` lines with the `python_version < "3.11"` / `>= "3.11"` markers, and evaluated them with `packaging`: 3.10 selects `!=1.97.0,<1.98.0`, 3.11/3.12/3.13 select the unconstrained one, and both stay in `[project.dependencies]`. Installed `.[dev]` on 3.10 (resolves litellm 1.96.2) and on 3.12 (resolves 1.98.0, the current latest release) and ran the full suite on both — 2197 passed / 4 skipped on each, ruff 0.15.22 clean, plus a smoke of the SDK's litellm surfaces (types.get_litellm_params, get_model_info, media_providers, openrouter_attribution, harness._runner) under 1.98.0. Also confirmed the retained cap is still load-bearing: litellm 1.98.0 declares requires-python `>=3.10,<3.15`, so pip happily installs it on 3.10 and it then dies at import with `cannot import name 'NotRequired' from 'typing'`.

Issue #985, opened 2026-08-27
[Python SDK] Structured logs, event loop and Postgres instance


pocesar:
## Describe the bug

https://github.com/Agent-Field/agentfield/blob/4d4d54e402bc89c6d0e52e8c826d679409214778/sdk/python/agentfield/logger.py#L178-L184

`_emit_structured_record` do 2 things at once: 
1. Dumps to stdout, that depending on the size of the payload (ie >1MB) it will block the event loop if the stdout is for any reason busy or slow. logging default `StreamHandler` uses a `threading.RLock` internally also, for double trouble, besides the Agentfield logger also using another global lock instance https://github.com/Agent-Field/agentfield/blob/4d4d54e402bc89c6d0e52e8c826d679409214778/sdk/python/agentfield/logger.py#L81-L89 In highly concurrent or heavy I/O driven deployments, it's a dice roll until something stalls.
3. The only way to omit this JSON dump is to manually patch `_emit_structured_record` and `_emit_plain` the agentfield logger module during initialization and make it a no-op. 

Ideally it should be an opt-in. There's a ambiguous env var `AGENTFIELD_LOGS_ENABLED` but it's a misnomer, since what it does is to watch for its own process stdout / stderr and intercept that. 

https://github.com/Agent-Field/agentfield/blob/4d4d54e402bc89c6d0e52e8c826d679409214778/sdk/python/agentfield/node_logs.py#L223-L232

## Steps to reproduce

2. Run '...'
3. See error

## Expected behavior

When logs are enabled but no stdout setting (that doesn't exist today, let's say `AGENTFIELD_ONLY_EMIT_LOGS`), only sends to the Postgres instance, through `_dispatch_to_cp` and no `print` line. 

Additionally if locks must be used to ensure order, either use a non-blocking Lock (with a tiny timeout that constantly yields) or use a FIFO queue that is async friendly, and will deal with global locks only when it needs to output. 

## Environment

- Control plane version:
- SDK version (if applicable):
- Deployment environment (local, docker, kubernetes, etc.):

AbirAbbas:
## Triage

**Classification:** bug ✓
**Suggested labels:** bug

**Likely affected files:**
- ./reasoners/helpers.py (matched: reason, help)
- ./cla-signatures.json (matched: json)
- ./.env.example (matched: env)
- ./reasoners/models.py (matched: reason)
- ./reasoners/__init__.py (matched: reason)

---

_Triage by github-buddy_

[...]

PR #994, opened 2026-08-27
feat(sdk): allow disabling structured log stdout mirroring

mikemikimike:
## Summary

Structured execution records were always mirrored to stdout, even when applications only needed control-plane forwarding. This can add large synchronous writes to event-loop workloads and makes it difficult to keep process output clean.

This change adds `AGENTFIELD_LOG_STDOUT`, defaulting to `true` for compatibility. Setting it to `false` suppresses the stdout mirror while preserving control-plane dispatch.

## Validation

- `uv run --frozen pytest sdk/python/tests/test_logger.py -q`
- `uv run --frozen ruff check sdk/python/agentfield/logger.py sdk/python/tests/test_logger.py`
- `uv run --frozen ruff format --check sdk/python/agentfield/logger.py sdk/python/tests/test_logger.py`
- `git diff --check`


AbirAbbas:
Took this over with a small follow-up commit (33a5fc6) rather than sending you round again — the feature is right, just the parsing.

`== "true"` made every value other than the literal `"true"` disable the mirror, so `AGENTFIELD_LOG_STDOUT=1` and `=yes` turned it off, which is the opposite of what they mean everywhere else in the SDK (`_TRUTHY_ENV_VALUES` in `agent.py` accepts `1`/`true`/`yes`). The one that would have bitten people is a set-but-empty variable — a bare `AGENTFIELD_LOG_STDOUT=` in a Compose file or `value: ""` in a k8s manifest — which silently dropped their structured logs, as did any typo. I switched it to the falsy-list convention `node_logs.logs_enabled()` already uses for on-by-default flags: `0`/`false`/`no`/`off` disable, anything else keeps the default, so the flag fails towards keeping records visible.

Confirmed live in a subprocess before and after: unset/`true` mirror, `false`/`0`/`no`/`off` stay silent, and `_dispatch_to_cp` still fires with the mirror off. Also added parametrized coverage for the truthy aliases and the empty/typo cases (your two tests passed against both the old and the corrected parsing, so nothing was guarding that), and moved the docs entry to "Agent Nodes > Python SDK agents" since only the Python SDK reads the variable.

AbirAbbas (dismissed):
Verified locally on the 3.10 leg — full pytest suite green (2213 passed) and ruff clean — plus a subprocess check that the mirror actually goes quiet with the flag off while control-plane dispatch keeps firing. Good fix for #985; I pushed one follow-up commit correcting the env parsing so `1`/`yes` and an empty value no longer silently disable logging.

mikemikimike:
Thanks for taking this over and for the detailed validation. I see commit `33a5fc6` now handles the truthy/falsy aliases and empty-value default correctly, while preserving control-plane dispatch. The subprocess checks and full Python test/ruff results are clear, and I am happy for the PR to proceed with this follow-up.

sdk/python/agentfield/logger.py
```diff
@@ -179,7 +188,8 @@ def _emit_structured_record(self, record: Dict[str, Any]) -> Dict[str, Any]:
         line = json.dumps(
```
pocesar: @mikemikimike quick comment here, the json.dumps will unnecessarily run when it's disabled. move this inside the `if` block

cc @AbirAbbas

mikemikimike:
Follow-up fix pushed in `ff99fa9`.

A deterministic reproduction showed that `_TeeTextIO.write()` blocks the event-loop thread when structured records are mirrored through `node_logs.install_stdio_tee()` and stdout is slow. Structured stdout mirroring is now opt-in (default off), so normal async SDK execution no longer enters that synchronous tee path. Explicit `AGENTFIELD_LOG_STDOUT=true`/`1`/`yes`/`on` preserves local mirroring, while control-plane dispatch remains enabled in either mode.

Validation: `uv run --frozen pytest sdk/python/tests/test_logger.py -q` (22 passed), Ruff check/format, and `git diff --check` passed.

santoshkumarradha (approved):
Tested this locally with `uv run --project sdk/python --extra dev pytest sdk/python/tests/test_logger.py -q`. The env parsing and the skipped serialization path both look right, and the behavior stays fail-safe by default.

PR #995, opened 2026-08-27
fix(control-plane): keep postgres payloads off local disk

### LINKED ISSUE #983: [Control Plane] Even when using postgres mode, payloads are dumped to local disk
## Describe the bug

Kubernetes deployment needs to be restarted from time-to-time to get rid of the accumulating payload files inside the data folder, after 30 days the pod was evicted due to 2GB of disk usage

## Steps to reproduce

1. Set mode to postgres
2. Do some runs
3. Payload is written to both workflow and execution tables and the disk as well

## Expected behavior

Only write to the disk when mode is indeed local

## Environment

- SDK version (if applicable): Any

ump45nose:
## Summary

In Postgres storage mode, execution input and result payloads are already persisted in the database. This change stops wiring the local `FilePayloadStore` in that mode, preventing every run from also accumulating duplicate payload files under the control-plane data directory. Local storage mode keeps its existing file-backed behavior.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs only
- [ ] Tests only
- [ ] CI / tooling
- [ ] Breaking change

## Test plan

- [x] `cd control-plane && go test -v ./internal/server -run "^TestNewPayloadStore$" -count=1`
- [x] `gofmt -w control-plane/internal/server/server.go control-plane/internal/server/server_additional_test.go`
- [x] `git diff --check`

## Test coverage

- [ ] If I removed code, I updated `coverage-baseline.json` in this PR only if the removal caused a legitimate regression and I called it out in the summary above.

## Checklist

- [x] I have read `docs/DEVELOPMENT.md`.

## Related issues / PRs

Fixes #983

### DIFF
--- a/control-plane/internal/server/server.go
+++ b/control-plane/internal/server/server.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
@@ -429,7 +430,7 @@ func NewAgentFieldServer(cfg *config.Config) (*AgentFieldServer, error) {
 		})
 	}
 
-	payloadStore := services.NewFilePayloadStore(dirs.PayloadsDir)
+	payloadStore := newPayloadStore(cfg.Storage.Mode, dirs.PayloadsDir)
 
 	// Configure SSRF-safe webhook client allowlist. Hosts/CIDRs listed here
 	// bypass the private-IP check (e.g. for internal Docker/K8s service names).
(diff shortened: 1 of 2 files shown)


santoshkumarradha (changes requested):
The direction makes sense, but this is still incomplete against current `main`. `savePayload()` in `control-plane/internal/handlers/execute.go` still dereferences `record.URI` unconditionally after `SaveBytes()`, so returning a nil payload-store result in postgres mode can panic on the first persisted payload. `#996` covers that extra path by handling nil saves explicitly, so I would either fold that guard into this PR or defer to the more complete fix there.

ump45nose:
Thanks for the review. #996 has now merged and includes both the `NopPayloadStore` wiring for Postgres mode and the nil-save guard in `savePayload`. That fully supersedes this branch, so I’m closing #995 to avoid duplicate and conflicting changes.

PR #996, opened 2026-08-27
fix(control-plane): skip disk payload writes in postgres storage mode

### LINKED ISSUE #983: [Control Plane] Even when using postgres mode, payloads are dumped to local disk

7vignesh:
## Summary

Fixes #983. When `storage.mode` is `postgres`, the control plane was unconditionally writing execution payloads to the local filesystem via `FilePayloadStore`, even though the same data is already persisted inline in the `input_payload` / `result_payload` BYTEA columns of the execution record. This caused unbounded disk growth (~2GB over 30 days) leading to pod evictions on Kubernetes.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Root cause

`server.go` always instantiated a `FilePayloadStore` regardless of the configured storage mode. Every execution wrote payloads to both the database (inline columns) and the filesystem (via URI), but only the database path was used on read in postgres mode.

## Fix

- Added `NopPayloadStore` - a no-op implementation of `PayloadStore` that returns nil on save
- When `cfg.Storage.Mode == "postgres"`, the server uses `NopPayloadStore` instead of `FilePayloadStore`
- `savePayload` now handles nil records gracefully (prevents panic)
- All read paths (`resolveExecutionData`, `previewPayload`, etc.) already check inline data first and only fall back to URI when inline is empty - since URIs are now nil in postgres mode, the file path is never reached

## Files changed

| File | Change |
|------|--------|
| `services/payload_store.go` | Added `NopPayloadStore` (no-op implementation) |
| `server/server.go` | Use `NopPayloadStore` when mode is postgres |
| `handlers/execute_helpers.go` | Handle nil record from `SaveBytes` |
| `services/payload_store_test.go` | Tests for `NopPayloadStore` |
| `handlers/coverage_handlers_90_additional_test.go` | Test `savePayload` with nop store |

## Test plan

- [x] `go build ./...` - clean
- [x] `go vet ./internal/services/ ./internal/server/ ./internal/handlers/` - clean
- [x] `go test ./internal/services/` - pass (59s)
- [x] `go test ./internal/handlers/` - pass (73s)
- [x] Verified read paths (`resolveExecutionData`, `previewPayload`) short-circuit on nil URI without calling `Open`

## Test coverage

- [x] If I removed code, I updated `coverage-baseline.json` in this PR *only if* the removal caused a legitimate regression.

## Checklist

- [x] I have linked any related issues (Fixes #983).

## Related issues / PRs

Fixes #983

### DIFF
--- a/control-plane/internal/handlers/coverage_handlers_90_additional_test.go
+++ b/control-plane/internal/handlers/coverage_handlers_90_additional_test.go
@@ -501,6 +501,11 @@ func TestExecuteHelpers_AdditionalCoverage(t *testing.T) {
 		require.NotNil(t, uri)
 		assert.True(t, strings.HasPrefix(*uri, "payload://"))
 
+		// NopPayloadStore returns nil record - savePayload must handle gracefully
+		controller.payloads = services.NopPayloadStore{}
+		nopURI := controller.savePayload(context.Background(), []byte(`{"data":true}`))
+		assert.Nil(t, nopURI)
+
 		controller.triggerWebhook("")
 		controller.webhooks = &mockWebhookDispatcher{
 			notifyFunc: func(ctx context.Context, executionID string) error {
--- a/control-plane/internal/handlers/execute_helpers.go
+++ b/control-plane/internal/handlers/execute_helpers.go
@@ -816,6 +816,9 @@ func (c *executionController) savePayload(ctx context.Context, data []byte) *str
 		logger.Logger.Warn().Err(err).Int("bytes", len(data)).Msg("failed to persist payload; proceeding without URI")
 		return nil
 	}
+	if record == nil {
+		return nil
+	}
 	uri := record.URI
 	return &uri
 }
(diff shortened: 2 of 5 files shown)


santoshkumarradha (approved):
Tested the touched paths locally with `go test ./internal/services -count=1` and `go test ./internal/server -run ^'TestNewPayloadStore$' -count=1` from `control-plane`. This is the right fix direction for postgres mode, and it covers the nil-save path that would otherwise be left behind.

Issue #989, opened 2026-08-27
Add minimal and max wait time to both control plane and Python SDK


pocesar:
## Summary

Most of the time you'll have an external service paired with the agent node Kubernetes deployment: Redis, a vector DB, the postgres connector itself. 
Allows some time to pass before starting to teardown everything

## Proposed solution

The idea follows the `maxSigtermDelay` and `minSigtermDelay`: https://github.com/GoogleCloudPlatform/cloud-sql-proxy-operator/blob/main/docs/api.md#authproxycontainerspec

When receiving a shutdown event (SIGTERM, SIGINT, etc) either on the control plane itself (a re-deployment or a new version being deployed) or on agents deployments, wait for at least `n` seconds before fanning out the shutdown signals to in-flight reasoners. And similarly, when an agent (using the python SDK here), it shouldn't kill the process straight away and should wait `n < kill < m`. A Kubernetes `lifecycle.stopSignal: SIGHUP` could also be used to differentiate from a hard `SIGTERM`.

Additionally a `/shutdown` internal endpoint could be exposed so Kubernetes `lifecycle.preStop.httpGet.path: /shutdown` could be used natively to start winding down 

## Alternatives considered

Right now, using a `lifecyle.preStop.sleep` with a obscene high number until it get SIGTERM'd. Some agent reasoners can take 4 minutes to complete, and 90% of the time it gets killed anyway.
The drawback is that new deployments take just as much to be live

AbirAbbas:
## Triage

**Classification:** feature ✓
**Suggested labels:** enhancement

**Likely affected files:**
- ./reasoners/webhook_router.py (matched: out, reasoners, reasoners)
- ./reasoners/models.py (matched: reasoners, reasoners)
- ./reasoners/__init__.py (matched: reasoners, reasoners)
- ./reasoners/conversational.py (matched: reasoners, reasoners)
- ./reasoners/helpers.py (matched: reasoners, reasoners)

---

_Triage by github-buddy_

AbirAbbas:
Partly shipped in v0.1.137, and the rest is in one PR.

[...]

Issue #987, opened 2026-08-27
[Control Plane] Kubernetes new deployments causes every in-flight run to fail and never recover


pocesar:
## Describe the bug

When there's a reasoner that is acting like an orchestrator, and can take 5-10 minutes to complete, it has a high chance of destroying the entire run (which is composed by multiple reasoners and skills) with the "in-flight reasoner cannot be revived" error.

The only workaround is to re-run the same entrypoint reasoner with the same inputs and start again. Happily, when this is done, the run will blast through until the step where it got killed due the persisted intermediary steps and the 1h cache that is enabled.

## Steps to reproduce

2. Run '...'
3. See error

## Expected behavior

When a reasoner or skill gets killed, there should be a way to restart the root run, keeping the same execution ID so the external consumers don't need to implement any sort of retries or re-submission. 
It's a trade-off for the immutability and deterministic behavior that is expected when you call the same reasoner with the same input, but generates different execution / run IDS, should be an opt-in flag

## Environment

- Control plane version: 0.1.127
- SDK version (if applicable): Python SDK
- Deployment environment (local, docker, kubernetes, etc.): Kubernetes

AbirAbbas:
## Triage

**Classification:** bug ✓
**Suggested labels:** bug

**Likely affected files:**
- ./reasoners/helpers.py (matched: reasoner, reasoners, reasoner, reasoner, reasoner, reasoner, help)
- ./reasoners/models.py (matched: reasoner, reasoners, reasoner, reasoner, reasoner, reasoner)
- ./reasoners/__init__.py (matched: reasoner, reasoners, reasoner, reasoner, reasoner, reasoner)
- ./reasoners/conversational.py (matched: reasoner, reasoners, reasoner, reasoner, reasoner, reasoner)
- ./reasoners/repo_config.py (matched: reasoner, reasoners, reasoner, reasoner, reasoner, reasoner)

---

_Triage by github-buddy_

AbirAbbas:
Where this stands after v0.1.137 (released yesterday):

[...]

PR #1011, opened 2026-08-29
fix: end SSE streams on shutdown so af server exits promptly; restore Go SDK notify-then-drain order

AbirAbbas:
Two regressions introduced by #1010 (found by an adversarial review of the merged diff), plus two small related corrections.

## Regressions fixed
- **`af server` waited up to 30s and exited 1 on Ctrl+C whenever the UI was open.** The graceful HTTP drain waited for the dashboard's SSE streams, which only end when the browser disconnects — and `af server` opens the browser by default. rc.12 exited instantly. Streaming handlers now end when shutdown starts (a server-level stream context cancelled before `http.Server.Shutdown`), normal in-flight requests keep the full drain budget, and a drain that merely runs out of time exits 0 with a warning instead of 1 (the macOS tray relaunches on non-zero exit).
- **Go SDK: `shuttingDown` was set before the control-plane notify**, so dispatches arriving during that notify (up to the client's 10s timeout) got a 503 the control plane records as a non-retryable failure; rc.12 accepted and drained that work. Restored notify-then-flag order; the 503 gate on all entry points stays for after the notify.

## Related
- The async-pool drain gets its own budget (max of the remaining shutdown budget and 5s) instead of inheriting an already-expired context after a slow HTTP drain.
- `terminationGracePeriodSeconds` is a Helm value (default 45) instead of a hardcoded template constant; `AGENTFIELD_LOG_MAX_LINE_BYTES` floor wording corrected per SDK.

## Validation contract
- With one SSE client attached and nothing else in flight, `Stop()` returns in well under a second; the stream ends cleanly.
- A non-streaming in-flight request still completes within `AGENTFIELD_SHUTDOWN_TIMEOUT`.
- Drain timeout → exit 0 + warning; real `Stop()` failure → exit 1.
- HTTP budget exhausted → queued async jobs still get ≥5s before being failed.
- Go agent: dispatch during the CP notify is accepted and drained; after the notify, 503.

## Tests
CI-literal gates run locally on the branch (base `cd483045` = rc.13):
  - control-plane: `go build ./...` ✓ · `golangci-lint run` 0 findings ✓ · `go test -tags sqlite_fts5` on server/handlers/config/cmd packages (12) ✓
  - sdk/go: `go build` · `go vet ./agent/...` · `go test ./...` · `go test -race ./agent/...` ✓
  - `helm template deployments/helm/agentfield` renders identically by default ✓

Manual (one isolated control plane): 
  - SSE regression: one client on `/api/ui/v1/executions/events`, then SIGTERM — **main: exited after 30.1s with "Error during shutdown: context deadline exceeded"; this branch: 0.1s, "Server stopped gracefully."**
  - `AGENTFIELD_SHUTDOWN_TIMEOUT=20` (bare seconds) is now honoured by the control plane (was silently ignored).
  - Re-ran the #1010 behavioural checks on this binary: `AGENTFIELD_AGENT_CALL_TIMEOUT=0s` → sync 2s reasoner returns 200 in 2.5s; 9 MiB registration → 413 (Content-Length and chunked), normal execute unaffected; stopped node → 503 `node_unavailable` with `Retry-After: 1` after the hold; SIGTERM to `af server` → full shutdown sequence, exit 0; no leftover listeners.

Refs #987 #989

### DIFF
--- a/control-plane/cmd/af/main.go
+++ b/control-plane/cmd/af/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log"
 	"os"
@@ -258,6 +259,10 @@ func drainOnShutdown(ctx context.Context, stopSignals func(), stop func() error)
 	}
 	fmt.Println("\nShutdown signal received, draining connections...")
 	if err := stop(); err != nil {
+		if errors.Is(err, context.DeadlineExceeded) {
+			log.Printf("Warning: shutdown drain timed out; forced close completed: %v", err)
+			return nil
+		}
 		return err
 	}
 	fmt.Println("Server stopped gracefully.")
--- a/control-plane/cmd/af/main_test.go
+++ b/control-plane/cmd/af/main_test.go
@@ -156,3 +156,11 @@ func TestDrainOnShutdownPropagatesStopError(t *testing.T) {
 		t.Fatalf("got %v, want %v", err, want)
 	}
 }
+
+func TestDrainOnShutdownTreatsTimeoutAsSuccessfulExit(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
+	require.NoError(t, drainOnShutdown(ctx, nil, func() error {
+		return context.DeadlineExceeded
+	}))
+}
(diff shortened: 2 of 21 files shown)


santoshkumarradha (approved):
Reviewed in a detached worktree. The shutdown path changes and the Go SDK lifecycle fix both look sound, and the focused local regression slices I ran matched the green GitHub check surface.

PR #1013, opened 2026-08-30
feat(sdk/go): add OrcaRouter as a first-class named provider

clementguarino06510-glitch:
This adds OrcaRouter as a first-class named provider in the Go SDK's AI package, so AgentField agents can route through it the same way they already route through OpenRouter or Infron — `ORCAROUTER_API_KEY` in the environment, and you're off to the races with `ai.DefaultConfig()`.

OrcaRouter is an OpenAI-compatible AI gateway built for both models and agents. Like OpenRouter, it exposes a provider/model namespace across many models — but it also combines adaptive routing, automatic failover, zero-markup inference, observability, guardrails, and agent-tool governance behind the same endpoint. Adding `orcarouter` as a first-class provider means AgentField users can use that stack directly, without treating OrcaRouter as an anonymous custom base URL. It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

The change mirrors the existing Infron integration (an OpenAI-compatible gateway that was wired in the same way): `Config.IsOrcaRouter()` detection by base URL and `orcarouter/` model prefix, `DefaultConfig()` resolution from `ORCAROUTER_API_KEY`, attribution headers on both the sync and streaming paths, and the `usage: {"include": true}` opt-in so responses carry native `usage.cost`. One deliberate difference: the `orcarouter/` prefix is part of the published model id on OrcaRouter (the gateway rejects the bare id), so it stays on the wire rather than being stripped like Infron's routing-only prefix.

**Files changed:** `sdk/go/ai/config.go`, `sdk/go/ai/client.go`, `sdk/go/ai/model_params.go`, new `sdk/go/ai/orcarouter_attribution.go` (plus a matching attribution test file and README section).

**Verification:** `go build ./...`, `go vet`, and `go test ./...` all pass in `sdk/go` (new tests cover config detection, precedence, attribution fallback/opt-out, the usage opt-in, and prefix preservation). I also ran a live check against `https://api.orcarouter.ai/v1` through the new code path — both the sync and streaming completions returned 200 with populated usage, and native cost came back inside `usage.cost`.

Discord: discord.gg/YEubt8enRA · X: https://x.com/OrcaRouter

I'm an engineer on the OrcaRouter team.

### DIFF
--- a/sdk/go/ai/README.md
+++ b/sdk/go/ai/README.md
@@ -1,10 +1,10 @@
 # Go SDK AI Package
 
-This package provides AI/LLM capabilities for the AgentField Go SDK, supporting both OpenAI and OpenRouter APIs with structured output support.
+This package provides AI/LLM capabilities for the AgentField Go SDK, supporting OpenAI, OpenRouter, Infron, and OrcaRouter APIs with structured output support.
 
 ## Features
 
-- ✅ **OpenAI & OpenRouter Support**: Works with both OpenAI API and OpenRouter for multi-model routing
+- ✅ **OpenAI, OpenRouter, Infron & OrcaRouter Support**: Works with OpenAI API, OpenRouter, Infron, and OrcaRouter for multi-model routing
 - ✅ **Structured Outputs**: JSON schema validation with Go struct support
 - ✅ **Streaming**: Support for streaming responses
 - ✅ **Type-Safe**: Automatic conversion from Go structs to JSON schemas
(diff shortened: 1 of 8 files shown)


santoshkumarradha (approved):
This looks good to me. I spot-checked the Go SDK provider wiring in a detached worktree, `go test ./ai -count=1` passed locally, and the full CI surface is now green after the fork workflows were approved. The remaining blocker looks to be the CLA only, so once that is signed, this should be good to go.

clementguarino06510-glitch:
Thanks for the review and the worktree spot-check of the Go SDK wiring.

Acknowledged on the CLA — the `license/cla` status on the head commit is still `pending` on cla-assistant (badge shows not signed), and that's the only outstanding item. I'll complete the CLA signing on the author side; once the check flips to green, this should be ready to merge. No code changes pending from my side.

AbirAbbas:
Closing this for now to keep the review queue clean: the CLA has been pending since Aug 30 (the `license/cla` check is still red after the acknowledgement here), and the branch now conflicts with `main` in `sdk/go/ai/README.md` and `sdk/go/ai/config.go` after the provider changes that landed since.

The code itself was already approved, so nothing is lost. Once the CLA is signed, reopen (or ping here) and we'll help get the two conflicts resolved so it can land.

Issue #960, opened 2026-08-25
proposal: make opencode harness calls explicit


ddbaron:
# Proposal: make OpenCode harness calls explicit

## Summary

I’d like to propose updating the AgentField OpenCode provider to build a
per-run `OPENCODE_CONFIG_CONTENT` overlay from the existing `harness(...)`
options. Each `opencode run` would select a fixed agent with `--agent`, attach
the system prompt and model to that agent, use `steps: 500`, apply the supplied
reasoning variant, and run with explicit noninteractive permissions.

This would let OpenCode receive the same intent that the caller already
expressed through the harness API, without changing SWE-AF-specific workflows.
SWE-AF is one consumer of this behavior, but the change should stay generic and
live in the AgentField OpenCode provider.

## Problem

The harness already accepts `system_prompt`, `tools`, `model`, `variant`, `cwd`,
`max_turns`, and `permission_mode`. The OpenCode provider currently uses only a
subset of those values when it builds a command similar to:

```text
opencode run --format json --dir <CWD> -m <MODEL> <PROMPT>
```

The system prompt is folded into the positional prompt as ordinary text,
`tools` is not translated into OpenCode permissions, and
`permission_mode` does not establish a reliable noninteractive boundary. As a
result, a run can depend on ambient OpenCode defaults, use an unintended agent,
duplicate system instructions, or wait for a permission approval that nobody
can provide.

## Proposed change

For each standalone OpenCode subprocess, the provider would generate an
`OPENCODE_CONFIG_CONTENT` overlay with one fixed generic agent name, such as
`agentfield-harness`. The deployment can continue to own the shared settings
for plugins, providers, model limits, and compaction. Meaning: if you customize your opencode deployment already, this will pickup your global mcp, plugins, etc... and only superceed what is declared.

The generated agent should have this shape:

```json
{
  "$schema": "https://opencode.ai/config.json",
  "default_agent": "agentfield-harness",
  "agent": {
    "agentfield-harness": {
      "prompt": "<SYSTEM_PROMPT>",
      "model": "<MODEL>",
      "mode": "primary",
      "steps": 500,
      "reasoningEffort": "<VARIANT>",
      "permission": {
        "*": "allow",
        "question": "deny",
        "task": "deny"
      }
    }
  }
}
```

`reasoningEffort` would be omitted when no reasoning variant is supplied. The
AgentField `variant` and OpenCode `reasoningEffort` values represent one
[...]

PR #1023, opened 2026-08-31
feat(sdk/python): isolate OpenCode harness runs

ddbaron:
## Summary

- Configure each standalone Python OpenCode run with a generated `agentfield-harness` agent through `OPENCODE_CONFIG_CONTENT`.
- Preserve caller and ambient OpenCode configuration through a deep merge while giving the generated harness fields precedence.
- Keep credentials out of the generated configuration and preserve the task prompt as the normal positional or stdin prompt.
- Keep AgentField's `max_turns` runner limit separate from OpenCode's fixed `steps` setting.
- Add targeted recursion protection for AgentField orchestration skills while leaving unrelated skills available.
- Retain an opt-in inline system-prompt compatibility path through `AGENTFIELD_OPENCODE_INLINE_SYSTEM_PROMPT=1`.

The overlay also carries the resolved model and OpenCode-specific reasoning variant when available. The provider-agnostic public `variant` API is intentionally kept in the related follow-up rather than duplicated here.

## Validation

- `uv run --extra dev pytest -q tests/test_harness_provider_opencode.py tests/test_harness_agent_wiring.py tests/test_harness_runner.py tests/test_harness_types.py`
- Python compilation
- `git diff --check`

The focused tests pass against the current `main` branch.

Related to the original proposal: https://github.com/Agent-Field/agentfield/issues/960

Related to the provider-agnostic follow-up: https://github.com/Agent-Field/agentfield/pull/1024

The fork preview remains available at https://github.com/ddbaron/agentfield/pull/3.

### DIFF
--- a/README.md
+++ b/README.md
@@ -391,6 +391,8 @@ Two examples already run at this load. The [deep-research engine](https://agentf
 | Tool access control | `tools=["Read", "Write", "Bash"]` |
 | Environment injection | `env={"KEY": "value"}` |
 | System prompt override | `system_prompt="..."` |
+| OpenCode per-run configuration | Preserves caller `OPENCODE_CONFIG_CONTENT` while applying the harness overlay |
+| OpenCode prompt compatibility | `AGENTFIELD_OPENCODE_INLINE_SYSTEM_PROMPT=1` enables the opt-in inline rollback |
 | Multi-layer output recovery | Cosmetic repair → retry → full retry |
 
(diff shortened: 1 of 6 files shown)


santoshkumarradha (approved):
Focused Python harness tests passed locally for me, and the config-overlay path looks sound. I also approved the fork-gated workflows so the full matrix is running now.

PR #1024, opened 2026-08-31
feat(sdk/python): expose provider-agnostic harness variants

ddbaron:
## Summary

- Add an optional provider-agnostic `variant` to `HarnessConfig`, `HarnessRunner.run`, and `Agent.harness`.
- Preserve the configured variant while allowing a per-call value to override it.
- Document that explicit variants take precedence over a `#variant` model suffix where the selected provider supports variants.
- Keep provider-specific interpretation in each provider adapter; this change only exposes and forwards the public API value.

This change is independently applicable to `main`. The OpenCode configuration and recursion-isolation work is kept in the related OpenCode pull request rather than made a prerequisite here.

## Validation

- `uv run --extra dev pytest -q tests/test_harness_agent_wiring.py tests/test_harness_runner.py tests/test_harness_types.py`
- Python compilation
- `git diff --check`

The focused tests pass against the current `main` branch.

Related to the original proposal: https://github.com/Agent-Field/agentfield/issues/960

Related to, but independently applicable from, the OpenCode change: https://github.com/Agent-Field/agentfield/pull/1023

The fork preview remains available at https://github.com/ddbaron/agentfield/pull/4.

### DIFF
--- a/README.md
+++ b/README.md
@@ -393,6 +393,7 @@ Two examples already run at this load. The [deep-research engine](https://agentf
 | System prompt override | `system_prompt="..."` |
 | OpenCode per-run configuration | Preserves caller `OPENCODE_CONFIG_CONTENT` while applying the harness overlay |
 | OpenCode prompt compatibility | `AGENTFIELD_OPENCODE_INLINE_SYSTEM_PROMPT=1` enables the opt-in inline rollback |
+| Provider-agnostic reasoning variants | `variant="high"` or a `#high` model suffix |
 | Multi-layer output recovery | Cosmetic repair → retry → full retry |
 
 #### Connector API (Fleet Management)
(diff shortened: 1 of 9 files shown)


santoshkumarradha (dismissed):
Focused Python harness tests passed locally for me, and the public variant plumbing looks clean. I also approved the fork-gated workflows so the full matrix is running now.

santoshkumarradha:
This still looks good from my side, but `#1023` landed first and this branch is now conflicting with current `main`. Please rebase on the latest `main` and rerun the checks, and once that is back green this should be good to go.

santoshkumarradha:
@ddbaron thanks for this !  @AbirAbbas lets enhance this with tsx and go support as well.

ddbaron:
@santoshkumarradha I rebased this PR onto the latest main and pushed commit, please take another look when convenient.

santoshkumarradha (approved):
Rebased state looks good. I reran the focused Python harness coverage on current main, and the new public `variant` plumbing still reads clean and stays provider-agnostic at this layer.

PR #1035, opened 2026-09-01
fix(sdk/python): coerce complex pydantic type hints in reasoner args

### LINKED ISSUE #1034: [Python SDK] Trouble with cross agent pydantic (un)marshalling
## Describe the bug

Sometimes complex pydantic models in reasoners and skills and multiple arguments can cause the inputs paramters to become a simple `dict`, forcing pydantic model coercion. 

And a side note: sometimes the parameters being put inside `TYPE_CHECK` blocks also builds properly but crash during deployment. 

## Steps to reproduce

Using parameters like: 

- `item: PydanticModel1 | PydanticModel2 | None = None`
- `items: list[PydanticModel1] | list[PydanticModel2] | None = None`
- `maybe_empty_sequence: Sequence[PydanticModel1 | None] = []`

In most cases, those fall through and crash during runtime. Nested JSON under the `input` can cause this as well. 

Using the new RUFF 0.16 UP rules might have to do with this since it prefers using `list[model] | None` instead of `Optional[List[model]]` on newer python versions

## Expected behavior

Inputs and outputs always recursively validate if they are a pydantic model, inputs are ensured to be an instance of the model / dataclass. Outputs survives roundrip losslessly  

## Environment

- SDK version (if applicable): 0.1.130

## Additional context

Another side note: ruff allows to place decorators (flake8-type-checking) in `runtime-evaluated-decorators` and `exempt-modules`. Not sure if this can be set at the AF SDK level that will be inherited or it needs to be set manually for every project

7vignesh:
## Summary

Fixes #1034. Reasoner/skill argument coercion in the Python SDK only handled a bare Pydantic model or a 2-arg `Optional[model]`. Complex type hints fell through and the caller received raw `dict`s instead of model instances.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Reproduction

Before this change, these all left the value as a raw `dict` / list of dicts:

```python
def r(item: M1 | M2 | None = None): ...              # -> dict, not M1
def r(items: list[M1] | None = None): ...            # -> [dict, dict]
def r(items: list[M1] | list[M2] | None = None): ... # -> [dict]
def r(seq: Sequence[M1 | None] = ()): ...            # -> [dict, None]
```

## Root cause

`pydantic_utils.py` detected a model only via `is_pydantic_model(actual_type)` after unwrapping a strict 2-arg `Optional`. A 3+ arg union, or a container whose element is a model, never matched, so `convert_dict_to_model` was never called. Separately, `convert_function_args` wrapped Pydantic v2 `ValidationError` with keyword args its v2 constructor rejects, so the `except Exception` fallback swallowed real validation failures and returned the raw dict.

## Fix

- `type_hint_involves_model()` recurses through `Union` / `list` / `Sequence` / `tuple` / `dict` args to detect a model anywhere in a hint.
- `_convert_with_type_hint()` validates via `pydantic.TypeAdapter(hint).validate_python(value)`, which resolves any nested / union / optional / container shape losslessly. Hints with no model are returned untouched, so plain `int` / `str` / `dict` params still pass through unchanged.
- `convert_function_args` now propagates validation errors (wrapped with the parameter name) instead of hiding them; non-validation errors (e.g. unresolved forward refs) still fall back to the original args.
- `convert_dict_to_model` uses `model_class.model_validate` (the correct v2 API).
- `should_convert_args` uses the same recursive detection so conversion triggers for the new shapes.

## Test plan

- [x] `cd sdk/python && python -m pytest tests/test_pydantic_utils.py` - 15 passed
- [x] `python -m pytest tests/test_skill_pydantic_models.py tests/test_decorators.py` - pass
- [x] Full unit suite: 2316 passed (remaining failures are `litellm`-not-installed and live-server environmental issues, pre-existing)
- [x] `python -m ruff check` and `ruff format --check` on the changed files - clean

New tests cover: multi-arg union of models, `list[model]`, union of `list[model]`, `Sequence[model | None]`, nested-model roundtrip, container validation-error propagation, and non-model pass-through for complex hints.

## Checklist

- [x] I have linked any related issues (Fixes #1034).

## Related issues / PRs

Fixes #1034


sdk/python/agentfield/pydantic_utils.py
```diff
@@ -138,29 +172,14 @@ def convert_function_args(
...
-                    ) from e
-            else:
-                # Not a Pydantic model, keep original value
-                converted_kwargs[param_name] = value
+            try:
+                converted_kwargs[param_name] = _convert_with_type_hint(value, type_hint)
+            except ValidationError as e:
+                # Add parameter context to the error, preserving the original
+                # ValidationError as the cause.
+                raise ValueError(
```
santoshkumarradha: This changes the surfaced exception type from `ValidationError` to `ValueError`, but the existing callers in `agent.py` and `decorators.py` only catch `ValidationError` around `convert_function_args()`. With this change, invalid model payloads can skip the current safe-input handling path. Can we preserve `ValidationError` here, or update those callers in the same PR?

santoshkumarradha (changes requested):
Thanks for chasing the nested-type coercion gaps here. I found one behavior regression that needs to be fixed before this is safe to merge: invalid model payloads now get rethrown as `ValueError`, but the current SDK call sites still only intercept `pydantic.ValidationError` when they convert reasoner and skill inputs. That means a bad payload can now miss the existing safe-validation path and either fall back to the raw dict or surface through the wrong error path. If you preserve `ValidationError` here, or update the callers in the same PR, this looks close.

7vignesh:
Good catch - fixed in 656bbd94.

You're right: wrapping the failure as `ValueError` bypassed the `except ValidationError` handlers in `agent.py` (reasoner + skill) and `decorators.py` that convert a bad payload into a safe `_HandlerInputError`, so it would have fallen through to the raw-dict fallback.

I went with option 1 (preserve `ValidationError`): `convert_function_args` now lets the `ValidationError` raised by `TypeAdapter.validate_python` propagate unchanged instead of re-wrapping it, and the top-level `except` re-raises `ValidationError` rather than the previous `(ValidationError, ValueError)`. No caller changes needed - all three call sites keep intercepting it exactly as before.

Verified end to end:

```python
def f(m: MyModel): ...
convert_function_args(f, (), {"m": {"x": "bad"}})   # -> raises pydantic.ValidationError

def g(items: list[MyModel]): ...
convert_function_args(g, (), {"items": [{"x": "bad"}]})  # -> raises pydantic.ValidationError
```

Both the direct-model and nested-container error paths now surface `ValidationError`. Updated the two tests to assert `pytest.raises(ValidationError)` explicitly. `test_pydantic_utils.py`, `test_skill_pydantic_models.py`, and `test_decorators.py` all pass; ruff check/format clean.

santoshkumarradha (approved):
Thanks for turning the review fix quickly. I rechecked the updated coercion path in a detached worktree: , ============================= test session starts ==============================
platform darwin -- Python 3.13.9, pytest-9.1.1, pluggy-1.6.0
rootdir: /Users/santoshkumarradha/Documents/agentfield/code/platform/agentfield
collected 0 items

============================ no tests ran in 0.00s =============================, ============================= test session starts ==============================
platform darwin -- Python 3.13.9, pytest-9.1.1, pluggy-1.6.0
rootdir: /Users/santoshkumarradha/Documents/agentfield/code/platform/agentfield
collected 0 items

============================ no tests ran in 0.00s =============================, and E902 No such file or directory (os error 2)
--> agentfield/pydantic_utils.py:1:1

E902 No such file or directory (os error 2)
--> tests/test_pydantic_utils.py:1:1

Found 2 errors. all passed locally. The remaining blocker on GitHub looks like the  workflow run, not this patch itself.

santoshkumarradha (approved):
Thanks for turning the review fix quickly. I rechecked the updated coercion path in a detached worktree: uv sync --python python3.13 --extra dev, uv run pytest tests/test_pydantic_utils.py, uv run pytest tests/test_skill_pydantic_models.py tests/test_decorators.py, and uv run ruff check agentfield/pydantic_utils.py tests/test_pydantic_utils.py all passed locally. The remaining blocker on GitHub looks like the coverage-summary workflow run, not this patch itself.

santoshkumarradha (dismissed):
I rechecked the latest head after the merge-from-main refresh. There is no new authored code beyond the previously reviewed SDK fix, the current CI surface is green, and the required coverage and CLA checks are both passing. This is good to go on my end.

sdk/python/agentfield/pydantic_utils.py
AbirAbbas: Verified this end to end at 4d6158da before resolving. ValidationError does propagate now — but tracing where it lands, the safe-input path you were pointing at didn't actually exist yet:

- agent.py turned it into `_HandlerInputError` in both the reasoner and the skill endpoint, and nothing caught that, so an invalid payload for one of the newly covered shapes came back as 500 "Internal Server Error" instead of the 422 the SDK builds. Reproduced against a real Agent over TestClient: `M1 | M2 | None` with `{"item": {"x": "bad"}}`, `list[M1]` with a bad element, and the same call on a skill — all 500, handler never entered.
- decorators.py:379 still did `raise ValidationError(msg, model=...)`, the same pydantic-v2 constructor misuse you flagged here. It was dead code before this PR (the error never escaped conversion); with the fix it is live, so the in-process reasoner-to-reasoner path failed with `TypeError: ValidationError.__new__() got an unexpected keyword argument 'model'`.

Both are fixed in the two commits I pushed, with HTTP-level tests that fail on the old callers (500 == 422, and the TypeError).

One more behaviour regression I found while checking: `def reasoner(m: MyModel = None)` — implicit Optional, no `Optional[...]` — used to arrive as None and started raising, because `apply_defaults()` fills None and `TypeAdapter(MyModel).validate_python(None)` rejects it. That is a 500 on every call to such a reasoner. Top-level None now passes through unchanged; None elements inside a container are still validated, and a null for a required parameter is still rejected by `_validate_handler_input` first.

Also cached the TypeAdapter (512-entry lru_cache, uncached fallback for unhashable hints): it was rebuilt per call, 55 us of adapter construction on a `list[Model] | None` parameter versus 1.2 us to validate with a cached one.

Full sdk-python suite (2387 passed), ruff, and the patch-coverage gate are green at that SHA.

AbirAbbas (approved):
Verified locally at 4d6158da, not just read.

What I checked, against a real Agent driven through its ASGI app (TestClient), plus the origin/main implementation side by side in the same harness:

- `M1 | M2 | None`, `Optional[list[M1]]`, `Sequence[M1 | None]` and a skill taking `list[M1]` all arrive as model instances now; on main they arrive as raw dicts. Explicit null still passes through as None.
- `dict[str, Any]`, `int`, `str` parameters are untouched — `should_convert_args` is False for those signatures and the payload keeps its identity, so no arbitrary dict gets validated by accident.
- String/forward-ref hints resolve through `get_type_hints`; an unresolvable one logs and falls back to the original args instead of blowing up.
- Invalid payloads now come back as 422 `{"detail": ...}` with the handler never entered, on both the reasoner and the skill endpoint (they were 500s before the last two commits), and the in-process decorator path raises pydantic's ValidationError rather than a TypeError.
- Anti-tautology: reverting `pydantic_utils.py` to main makes six of the new tests fail (`should_convert_args ... is True` -> False, and DID NOT RAISE ValidationError); reverting `agent.py`/`decorators.py` makes the three new caller tests fail with `assert 500 == 422` and the constructor TypeError.

Gates at this SHA: ruff clean, full sdk-python suite 2387 passed / 4 skipped, sdk-python coverage 94.7%, patch-coverage gate passed; all 21 CI checks green (the one `coverage (control-plane)` failure was a runner-level abort, green on re-run).

Perf note for the record: coercion of a `list[Model] | None` argument costs about 75 us per call more than no coercion, which is ~3-5% of the reasoner round trip I measured (1.3-2.7 ms) — and the TypeAdapter is cached now, so it is close to free after the first call.

PR #1041, opened 2026-09-05
fix(control-plane): compare execution timestamps by instant on SQLite (#1040)

### LINKED ISSUE #1040: [Control Plane] SQLite cleanup can time out fresh executions with non-UTC timestamps
## Summary

   When AgentField uses local SQLite storage on a host with a non-UTC timezone, the execution-cleanup reaper can mark a fresh, active execution as timed out.

   The affected rows contain timestamps with a local timezone offset, while the cleanup cutoff is generated in
 UTC. SQLite can compare these timestamp values lexicographically as text instead of comparing the represented
 instants.

   ## Observed behavior

   A disposable SWE-AF planning run on AgentField `v0.1.138-rc.9` was marked with:

   ```text
   execution timed out (no activity)
 ```

 The configured stale timeout was 30 minutes, but the execution was marked timed out approximately 108 seconds
 after it started. Later successful status callbacks were rejected with HTTP 409 because the execution had already
 become terminal.

 Example values from the affected path:

 ```text
   stored updated_at: 2026-09-03 06:52:18.193275-05:00
   UTC equivalent:    2026-09-03 11:52:18.193275Z
   cleanup cutoff:    2026-09-03 11:24:05+00:00
 ```

 The stored timestamp is newer than the cutoff by instant, so it should not be considered stale. However, the
 textual comparison evaluates incorrectly:

 ```sql
   SELECT
     '2026-09-03 06:52:18.193275-05:00'
     <=
     '2026-09-03 11:24:05+00:00';
   -- 1

   SELECT
     julianday('2026-09-03 06:52:18.193275-05:00')
     <=
     julianday('2026-09-03 11:24:05+00:00');
   -- 0
 ```

 ## Likely cause

 The stale-selection queries use:

 ```sql
   COALESCE(updated_at, created_at, started_at) <= ?
 ```

 and:

 ```sql
   ORDER BY COALESCE(updated_at, created_at, started_at) ASC
 ```

 The cutoff is generated with time.Now().UTC(), but several workflow timestamp write paths still use time.Now()
 without normalizing to UTC, including:

 - control-plane/internal/storage/local.go
 - control-plane/internal/handlers/execution_notes.go

 The affected selection paths appear to include:

 - MarkStaleExecutions
 - MarkStaleWorkflowExecutions
 - RetryStaleWorkflowExecutions

 ## Expected behavior

 An execution should be selected for cleanup only when its timestamp represents an instant older than the
 configured stale cutoff, regardless of the timezone offset used when it was stored.

 ## Suggested fix

 1. Normalize all persisted execution created_at and updated_at values with time.Now().UTC().
 2. For SQLite, compare timestamp expressions using an instant-aware function such as:
    ```sql
      julianday(COALESCE(updated_at, created_at, started_at))
      <= julianday(?)
    ```
 3. Apply the same normalization to the corresponding ORDER BY expressions and retry-selection query.
 4. Keep the PostgreSQL query path database-specific rather than using SQLite's julianday() there.
 5. Add regression tests using equal instants represented with different timezone offsets, covering both stale
    marking and retry selection.

 ## Environment

 - AgentField: v0.1.138-rc.9
 - Storage: local SQLite
 - Cleanup interval: 5 minutes
 - Stale timeout: 30 minutes
 - Reproducer: disposable SWE-AF planning run using an OpenCode-backed reasoner

## Local Workaround
Running `TZ=UTC` with `agentfield serve` I don't experience the error. This forced sqllite to use UTC for all values so the local TZ comparison issue doesn't surface.

7vignesh:
## Summary

On a non-UTC host using local SQLite storage, the stale-execution reaper could mark a fresh, active execution as timed out within seconds of starting (the issue reports ~108s against a 30-minute timeout). Later successful status callbacks were then rejected with HTTP 409 because the row had already become terminal. This PR compares execution timestamps by instant instead of by text, and persists execution/workflow timestamps in UTC so the offset is never stored.

Fixes #1040.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs only
- [ ] Tests only
- [ ] CI / tooling
- [ ] Breaking change

Root cause: SQLite compares timestamp columns as text. Write paths persisted timestamps with `time.Now()`, so on a non-UTC host they carried a local zone offset (e.g. `...-05:00`). A fresh local value sorts lexically before a UTC cutoff even though its instant is newer, so the reaper's `COALESCE(updated_at, created_at, started_at) <= cutoff` test matched rows that were not actually stale.

Fix has two layers:

1. Read path: wrap the timestamp comparison and `ORDER BY` in `julianday()` for SQLite so the comparison is instant-aware and offset-correct. Postgres uses `timestamptz` (already instant-aware) and has no `julianday()`, so the Postgres query path is left unchanged via a dialect-aware helper. This also corrects rows already persisted with an offset.
2. Write path: normalize execution and workflow `created_at`/`updated_at` to `time.Now().UTC()` so newly stored rows never carry a local offset.

Applies to `MarkStaleExecutions`, `MarkStaleWorkflowExecutions`, and `RetryStaleWorkflowExecutions`.

Behavior that does NOT change: the set of rows reaped/retried for correctly-stored UTC timestamps, all status transitions, the executions-table sync, the retry logic, and the parent-child skip guards. Only the timestamp comparison semantics change (text -> instant).

## Test plan

- `cd control-plane && go test ./internal/storage/ -run 'Stale|Retry|NonUTC' -count=1`
- `cd control-plane && go test ./internal/storage/ ./internal/handlers/ -count=1`

New regression tests in `internal/storage/stale_execution_tz_test.go`:
- Fresh execution stored with a `-05:00` offset must NOT be reaped (fails before the fix, passes after).
- Genuinely stale execution stored with a `-05:00` offset must still be reaped.
- Same fresh-not-reaped guard for `MarkStaleWorkflowExecutions` and `RetryStaleWorkflowExecutions`.

I confirmed the two fresh-timestamp tests fail on the unpatched code and pass after the fix, and that all existing stale/retry and NULL-fallback reaper tests still pass.

## Test coverage

- [x] I did not remove code, so no baseline changes are needed.

## Checklist

- [x] I have read the development docs.
- [x] I have linked the related issue (Fixes #1040).

## Related issues / PRs

Fixes #1040

### DIFF
--- a/control-plane/internal/handlers/execution_notes.go
+++ b/control-plane/internal/handlers/execution_notes.go
@@ -135,7 +135,9 @@ func AddExecutionNoteHandler(storageProvider ExecutionNoteStorage, ownershipEnfo
 
 			// Add the new note
 			execution.Notes = append(execution.Notes, note)
-			execution.UpdatedAt = time.Now()
+			// Persist in UTC so the stored text carries no local zone offset,
+			// which would break the stale-execution reaper on SQLite (#1040).
+			execution.UpdatedAt = time.Now().UTC()
 
 			return execution, nil
 		})
--- a/control-plane/internal/storage/execution_records.go
+++ b/control-plane/internal/storage/execution_records.go
@@ -12,6 +12,25 @@ import (
 	"github.com/Agent-Field/agentfield/control-plane/pkg/types"
 )
 
+// staleTimestampExpr returns the SQL expression used to compare an execution's
+// most recent activity timestamp against the stale cutoff.
+//
+// SQLite stores timestamps as text and compares them lexically. When a host
+// runs in a non-UTC timezone, time.Now() persists values carrying a local zone
+// offset (e.g. "...-05:00"), and a fresh local value can sort BEFORE a UTC
+// cutoff even though its instant is newer, wrongly reaping active executions
+// (#1040). Wrapping the value in julianday() forces an instant-aware numeric
+// comparison that is offset-correct.
+//
+// Postgres columns are timestamptz and already compare by instant, and
+// julianday() does not exist there, so Postgres keeps the plain expression.
+func (ls *LocalStorage) staleTimestampExpr(col string) string {
+	if ls.requireSQLDB().Mode() == "postgres" {
+		return col
+	}
+	return "julianday(" + col + ")"
+}
+
 // maxNodesForDepthCalc caps the number of executions for which we compute DAG depth to avoid heavy queries.
 const maxNodesForDepthCalc = 1000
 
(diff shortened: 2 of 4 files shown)


santoshkumarradha (approved):
This looks good. The SQLite path now compares timestamps by instant, existing offset-bearing rows are covered, and new writes are normalized to UTC. The focused storage and handler suites pass locally, and CI plus CLA are green.

Issue #291, opened 2026-03-19
[Epic] Add Cursor as a Harness Provider


santoshkumarradha:
## Overview

Add Cursor as a first-class harness provider in both the Python and Go SDKs. Cursor's CLI (`agent`) supports headless execution with JSON/streaming output, session resume, model selection, and MCP support — making it a natural fit alongside our existing providers (claude-code, codex, gemini, opencode).

### ToS Compliance

✅ Cursor's Terms of Service (Jan 13, 2026) **do not prohibit** automated/programmatic CLI use:
- Section 1.5 Use Restrictions cover reverse engineering, derivative works, reselling — not automation
- Section 1.7 explicitly acknowledges auto-code execution (`--force` flag)
- Official headless docs at cursor.com/docs/cli/headless endorse "scripts and automation workflows"
- Cloud Agents API (beta) is designed for programmatic agent management

### CLI Mapping

| AgentField Option | Cursor CLI Flag |
|---|---|
| Non-interactive mode | `agent -p --force --trust` |
| JSON output | `--output-format json` |
| NDJSON streaming | `--output-format stream-json` |
| Working directory | `--workspace <path>` |
| Model selection | `--model <model>` |
| Session resume | `--resume <chatId>` |
| Permission mode | `--mode plan\|ask\|agent` |
| MCP auto-approve | `--approve-mcps` |
| Auth | `--api-key <key>` or `CURSOR_API_KEY` env |

### Child Issues

- [ ] #292 — **Python SDK**: Add `CursorProvider` to `sdk/python/agentfield/harness/providers/`
- [ ] #293 — **Go SDK**: Add `CursorProvider` to `sdk/go/harness/`
- [ ] #294 — **Example**: Create Cursor harness usage example
- [ ] #295 — **Cloud Agents**: Add Cloud Agent API providers (Cursor, Copilot, Claude Agent SDK)
- [ ] Agent-Field/website-docs#56 — **Docs**: Add Cursor harness documentation

### Installation

Users must install Cursor CLI separately:
```bash
curl https://cursor.com/install -fsS | bash
```

Or set custom binary path via `cursor_bin` config field.

PR #1042, opened 2026-09-06
feat(python-sdk): add CursorProvider to the harness

### LINKED ISSUE #292: [harness] Add Cursor CLI provider to Python SDK
## Parent Epic
Part of #291 — Add Cursor as a Harness Provider

## Summary

Add `CursorProvider` to the Python SDK harness system. This follows the exact same pattern as the existing `ClaudeCodeProvider`, `CodexProvider`, `GeminiProvider`, and `OpenCodeProvider`.

### New
- `sdk/python/agentfield/harness/providers/cursor.py` — `CursorProvider` class

### Modify
- `sdk/python/agentfield/harness/providers/_factory.py` — Add `"cursor"` to `SUPPORTED_PROVIDERS`, wire factory
- `sdk/python/agentfield/harness/_runner.py` — Add `cursor_bin` to config (if not already generic)
- `sdk/python/agentfield/harness/providers/__init__.py` — Export

### Tests
- `sdk/python/tests/harness/test_cursor.py`

### CLI Command Construction

```python
cmd = ["agent", "-p", "--force", "--trust", "--output-format", "stream-json"]

if options.cwd:
    cmd.extend(["--workspace", options.cwd])
if options.model:
    cmd.extend(["--model", options.model])
if options.resume_session_id:
    cmd.extend(["--resume", options.resume_session_id])
if options.permission_mode == "plan":
    cmd.extend(["--mode", "plan"])

# API key via env
env["CURSOR_API_KEY"] = api_key

cmd.append(prompt)
```

### Output Parsing

Use `stream-json` mode (NDJSON). Parse events:
- `{"type": "result", "subtype": "success", "result": "<text>", "session_id": "<uuid>", "duration_ms": N}` — final result
- Tool call events for streaming (optional first pass — can use `json` mode if simpler)

### Session Resume

`session_id` from result event maps to `--resume <chatId>` on next invocation. Store in `RawResult.session_id`.

### Config

Add to `HarnessConfig`:
```python
cursor_bin: Optional[str] = None  # Override binary path (default: "agent")
```

### Permission Mode Mapping

| AgentField | Cursor CLI |
|---|---|
| `"auto"` | `--force` (default, already in base flags) |
| `"plan"` | `--mode plan` |
| `None` | `--mode ask` |

## Acceptance Criteria

- [ ] `CursorProvider` implements `HarnessProvider` protocol
- [ ] Factory resolves `"cursor"` to `CursorProvider`
- [ ] Prompt passed as last positional argument
- [ ] JSON/stream-json output parsed correctly
- [ ] Session resume works via `--resume`
- [ ] `cursor_bin` config allows custom binary path
- [ ] Schema validation uses same `.agentfield_output.json` pattern
- [ ] Tests cover: basic execution, session resume, error handling, timeout
- [ ] Follows exact patterns from existing providers (reference: `codex.py`)

dchaudhari7177:
Closes #292. Part of #291.

Follows `codex.py` as the reference, per the last acceptance item: same `RawResult`/`Metrics` mapping, same failure-type classification, same `env`/`cwd` handling, same `parse_jsonl` / `extract_final_text` / `extract_token_usage` pipeline.

### Four places I did not copy blindly

**The binary is `agent`, not `cursor`.** Cursor ships its headless agent under that name, so `cursor_bin` defaults to `"agent"` — the one provider whose default does not match its own name. There is a test pinning it, because that mismatch looks like a typo and is the sort of thing a later edit would helpfully "correct".

**An unset `permission_mode` maps to `--mode plan`, not `--mode ask`.** The issue's table says `None` → `--mode ask`, and I think that is wrong in this context: `ask` waits for an interactive answer, and a subprocess has nobody to give one, so the run would **hang until the harness timeout** rather than fail. Planning is the safe reading of "no mode stated", and it fails fast. `"auto"` → `--force` and `"plan"` → `--mode plan` are as specified. Happy to change it if `ask` is deliberate.

**`CURSOR_API_KEY` only when supplied.** Setting it unconditionally would clobber an inherited key with an empty string. It goes through the environment, never argv — a test asserts the key does not appear in the command, since argv is visible in `ps`.

**`--resume` omitted on an empty id.** The first turn of a session has no chat id, and `--resume ""` is an error rather than a no-op.

### Wiring

`SUPPORTED_PROVIDERS`, `build_provider`, `HarnessConfig.cursor_bin`, the `provider` field's description, and the `_runner` option passthrough.

### Tests — 19, all passing

Basic execution and exact command construction; prompt-last positioning (a flag after it would be read as part of the prompt); session resume in both directions — `--resume` out, `session_id` back; the three permission modes, parametrised; api-key present and absent; binary-missing → `HarnessProviderUnavailable`; timeout → `FailureType.TIMEOUT` not `CRASH`; non-zero exit **with** output (not an error, matching codex) and **without** (error); kill-by-signal; malformed NDJSON lines skipped; cost estimation into metrics.

`test_harness_factory.py`, `test_harness_provider_codex.py`, `test_harness_defaults.py` and `test_harness_packaging.py` still pass — 56 total, nothing regressed.

`ruff check` clean on every file I touched. `ruff format --check` clean on the new files; `agentfield/types.py` reports as unformatted, but it does so on unmodified `main` too, so I left it rather than mixing an unrelated reformat into this diff.

### Not in scope here

`.agentfield_output.json` schema validation is handled generically by the runner rather than per-provider, so there was nothing provider-specific to add — flagging that rather than silently ticking the box. If Cursor needs its own hook there, that is a separate change.

### DIFF
--- a/sdk/python/agentfield/harness/providers/cursor.py
+++ b/sdk/python/agentfield/harness/providers/cursor.py
@@ -1,0 +1,168 @@
+"""Cursor provider using CLI subprocess (agent -p --output-format stream-json)."""
+
+from __future__ import annotations
+
+import time
+from typing import Any, Dict, List, Optional
+
+from agentfield.harness._availability import ensure_cli_available, provider_unavailable
+from agentfield.harness._cli import (
+    estimate_cli_cost,
+    extract_final_text,
+    extract_token_usage,
+    parse_jsonl,
+    resolve_model_and_variant,
+    run_cli,
+    strip_ansi,
+)
+from agentfield.harness._result import FailureType, Metrics, RawResult
+
+
+class CursorProvider:
+    """Cursor CLI provider. Invokes `agent -p --output-format stream-json`.
+
+    The binary is named ``agent``, not ``cursor``: Cursor's CLI ships its
+    headless agent under that name, which is why the default here does not
+    match the provider name the way it does for codex or gemini.
+    """
+
+    def __init__(self, bin_path: str = "agent"):
+        self._bin = bin_path
+
+    async def execute(self, prompt: str, options: dict[str, object]) -> RawResult:
+        ensure_cli_available("cursor", self._bin)
+        # -p is headless (print) mode. --trust accepts the workspace without
+        # an interactive prompt, which a subprocess can never answer.
+        cmd = [self._bin, "-p", "--trust", "--output-format", "stream-json"]
+
+        # Agent root: project_dir is the canonical field, fall back to cwd,
+        # matching the codex provider. Cursor spells it --workspace.
+        root = options.get("project_dir") or options.get("cwd")
+        if isinstance(root, str):
+            cmd.extend(["--workspace", root])
+
+        # permission_mode -> Cursor's --force / --mode. Unlike codex there is
+        # no sandbox dimension: --force is "act without asking", `--mode plan`
+        # plans without editing, and `--mode ask` would block forever in a
+        # subprocess, so an unset mode maps to plan rather than ask.
+        permission_mode = options.get("permission_mode")
+        if permission_mode == "plan":
+            cmd.extend(["--mode", "plan"])
+        elif permission_mode == "auto":
+            cmd.append("--force")
+        else:
+            cmd.extend(["--mode", "plan"])
+
+        model_value, _variant_value = resolve_model_and_variant(options)
+        if model_value:
+            cmd.extend(["--model", model_value])
+
+        # Session resume: the chat id comes back as session_id on the result
+        # event, and goes out as --resume on the next call.
+        resume_session_id = options.get("resume_session_id")
+        if isinstance(resume_session_id, str) and resume_session_id:
+            cmd.extend(["--resume", resume_session_id])
+
+        # Prompt last, as a positional argument.
+        cmd.append(prompt)
+
+        env: Dict[str, str] = {}
+        env_value = options.get("env")
+        if isinstance(env_value, dict):
+            env = {
+                str(key): str(value)
+                for key, value in env_value.items()
+                if isinstance(key, str) and isinstance(value, str)
+            }
+
+        # The CLI reads its credential from the environment. Only set it when
+        # the caller supplied one, so an inherited CURSOR_API_KEY is not
+        # clobbered with an empty string.
+        api_key = options.get("api_key")
+        if isinstance(api_key, str) and api_key:
+            env["CURSOR_API_KEY"] = api_key
+
+        cwd: Optional[str] = root if isinstance(root, str) else None
+        start_api = time.monotonic()
+
+        try:
+            stdout, stderr, returncode = await run_cli(cmd, env=env, cwd=cwd)
+        except FileNotFoundError as exc:
+            raise provider_unavailable("cursor", self._bin) from exc
+        except TimeoutError as exc:
+            return RawResult(
+                is_error=True,
+                error_message=str(exc),
+                failure_type=FailureType.TIMEOUT,
+                metrics=Metrics(),
+            )
+
+        api_ms = int((time.monotonic() - start_api) * 1000)
+        events = parse_jsonl(stdout)
+        result_text = extract_final_text(events)
+
+        num_turns = 0
+        total_cost: Optional[float] = estimate_cli_cost(
+            model=model_value or "",
+            prompt=prompt,
+            result_text=result_text,
+        )
+        session_id = ""
+        messages: List[Dict[str, Any]] = events
+
+        for event in events:
+            event_type = event.get("type")
+            if event_type == "assistant":
+                num_turns += 1
+            elif event_type == "result":
+                # The result event carries the chat id to resume with. Read it
+                # from any event that has one, since a stream may carry the id
+                # on an earlier system event too.
+                session_id = str(event.get("session_id", "")) or session_id
+            elif not session_id and event.get("session_id"):
+                session_id = str(event.get("session_id", ""))
+
+        tokens = extract_token_usage(events)
+
+        clean_stderr = strip_ansi(stderr.strip()) if stderr else ""
+
+        if returncode < 0:
+            failure_type = FailureType.CRASH
+            is_error = True
+            error_message: str | None = (
+                f"Process killed by signal {-returncode}. stderr: {clean_stderr[:500]}"
+                if clean_stderr
+                else f"Process killed by signal {-returncode}."
+            )
+        elif returncode != 0 and result_text is None:
+            failure_type = FailureType.CRASH
+            is_error = True
+            error_message = (
+                clean_stderr[:1000]
+                if clean_stderr
+                else (f"Process exited with code {returncode} and produced no output.")
+            )
+        else:
+            failure_type = FailureType.NONE
+            is_error = False
+            error_message = None
+
+        return RawResult(
+            result=result_text,
+            messages=messages,
+            metrics=Metrics(
+                duration_api_ms=api_ms,
+                num_turns=num_turns,
+                total_cost_usd=total_cost,
+                session_id=session_id,
+                input_tokens=tokens["input_tokens"],
+                output_tokens=tokens["output_tokens"],
+                cache_read_tokens=tokens["cache_read_tokens"],
+                cache_creation_tokens=tokens["cache_creation_tokens"],
+                model=model_value,
+            ),
+            is_error=is_error,
+            error_message=error_message,
+            failure_type=failure_type,
+            returncode=returncode,
+        )
--- a/sdk/python/agentfield/harness/_availability.py
+++ b/sdk/python/agentfield/harness/_availability.py
@@ -30,6 +30,15 @@ PROVIDER_SPECS = {
         install_command="npm install -g @openai/codex",
         auth_env_vars=("OPENAI_API_KEY",),
     ),
+    "cursor": ProviderSpec(
+        # Cursor ships its headless CLI as `agent`, not `cursor`, so this is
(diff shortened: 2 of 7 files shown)


santoshkumarradha (approved):
This looks good on my side. The provider follows the existing harness contract, keeps the API key out of argv, handles session resume and failure modes, and the focused provider suite plus Ruff pass locally. I also checked the command flags against the installed Cursor CLI.

santoshkumarradha (changes requested):
One integration point is still missing. Adding `cursor` to `SUPPORTED_PROVIDERS` makes `harness_doctor()` index `PROVIDER_SPECS["cursor"]`, but this PR does not add that spec. The full Python CI reproduces it as `test_doctor_checks_all_supported_providers_with_no_args` failing with `KeyError: cursor`. Please add the Cursor availability/doctor spec and cover the doctor report. The focused provider tests and Ruff are otherwise clean.

dchaudhari7177:
Good catch — thank you, that was a real integration hole and the reproducer was exact.

`PROVIDER_SPECS["cursor"]` is added in `c85837d`. The detail worth flagging: it is the one spec whose `binary` does not match its provider name. Cursor ships its headless CLI as `agent`, so the doctor probes `agent --version`; specifying `cursor` there would have made `harness_doctor()` report the provider as not installed on a machine where it is. There's a comment on the spec pointing at the matching note on `CursorProvider.bin_path` so the two stay in step.

Three tests rather than one:

1. `test_doctor_probes_cursor_through_its_agent_binary` — full `ProviderHealth` equality, and asserts the probe command is `["agent", "--version"]`.
2. `test_doctor_reports_a_missing_cursor_cli_with_its_install_command` — the `binary_not_found` path.
3. `test_every_supported_provider_has_a_doctor_spec` — `SUPPORTED_PROVIDERS <= set(PROVIDER_SPECS) | {"claude-code"}`. This is the one that matters beyond this PR: `harness_doctor()` indexes `PROVIDER_SPECS` directly, so the next provider added to the set without a spec is a runtime `KeyError`, not an import error. `claude-code` is the documented exception since it takes the `_claude_health` wrapper branch instead of a CLI probe.

Verification:

```
pytest tests/test_harness_doctor.py tests/test_harness_provider_cursor.py   28 passed
ruff check agentfield tests                                                 All checks passed
ruff format --check                                                         clean
```

The 7 `test_harness_ai_schema_repair.py` failures in my local run are present on `main` with this branch stashed, so they're not from this change.

On the CLA: it's showing not-signed on this PR — I'll get that sorted.

santoshkumarradha (dismissed):
Thanks for fixing the missing doctor integration. I rechecked the updated head: the Cursor provider and doctor tests pass locally, and Ruff is clean. The provider spec now probes the correct `agent` binary and the coverage guard will catch future provider/spec drift. This looks good once the CLA and CI are green.

santoshkumarradha:
The updated implementation and all CI checks look good. The CLA is the only remaining step; once that’s signed, this should be good to go.

AbirAbbas (commented):
Went through this against the real Cursor CLI rather than just reading it: installed the current release (2026.09.08-6caf4ff) and diffed every flag the provider emits against `agent --help`. All of them are accepted — `-p`, `--trust`, `--output-format stream-json`, `--workspace`, `--mode plan`, `--force`, `--model`, `--resume <id>`, prompt last, and `CURSOR_API_KEY` read from the environment. (Unknown flags and bad `--mode` values are rejected by the arg parser *before* the auth check, so a run that gets as far as "Authentication required" proves the rest parsed.) The doctor-spec commit does what it claims too: dropping `_availability.py` back to main makes four tests in `test_harness_doctor.py` fail with `KeyError: 'cursor'`, including the generic all-providers one. ruff, the full Python suite (2386 passed) and the coverage/patch gates are green at c85837dd.

One thing I think should change before this lands — the default `permission_mode` mapping. Details inline.

Two smaller notes, take or leave:

- `docs/harness-providers.md` still has the install table without a `cursor` row; a user reading the docs won't find out the binary is called `agent`.
- The `assistant` event in `RESULT_STREAM` is `{"type":"assistant","content":"working"}`, but the shipped CLI emits `{"type":"assistant","message":{"role":"assistant","content":[{"type":"text","text":...}]},"session_id":...}`. Nothing breaks today because the final text and the chat id both come off the `result` event, but the fixture isn't pinning the shape the CLI actually produces.

sdk/python/agentfield/harness/providers/cursor.py
```diff
@@ -0,0 +1,168 @@
...
+        # no sandbox dimension: --force is "act without asking", `--mode plan`
+        # plans without editing, and `--mode ask` would block forever in a
+        # subprocess, so an unset mode maps to plan rather than ask.
+        permission_mode = options.get("permission_mode")
+        if permission_mode == "plan":
+            cmd.extend(["--mode", "plan"])
+        elif permission_mode == "auto":
+            cmd.append("--force")
+        else:
+            cmd.extend(["--mode", "plan"])
```
AbirAbbas, line 54: An unset `permission_mode` mapping to `--mode plan` makes the provider read-only by default, and I don't think the premise behind it holds.

The CLI never defaults to `ask`: `--mode` takes only `plan|ask`, and omitting it leaves `-p` in its normal execution mode, which the CLI's own help describes as "Print responses to console (for scripts or non-interactive use). Has access to all tools, including write and shell." There's also no hang to guard against — in headless mode the CLI auto-rejects approval requests instead of waiting on them.

The cost isn't cosmetic. The harness produces schema output by asking the agent to *write a file* (`_schema.build_prompt_suffix`: "You MUST use your Write tool to create this file: ..."), and `HarnessConfig.permission_mode` defaults to `None`. So `app.harness(prompt, schema=Model)` with `provider="cursor"` runs under `--mode plan` — "read-only/planning (analyze, propose plans, no edits)" — and can never create the output file. Plain task runs silently edit nothing too.

It's also the only adapter here that does this. codex, gemini, pi and claude-code all leave the CLI at its own default when `permission_mode` is unset, and grok goes the other way explicitly (`bypassPermissions`, with the comment "Harness runs are unattended; prefer non-interactive approvals").

Dropping the `else` is all it takes:

```python
permission_mode = options.get("permission_mode")
if permission_mode == "plan":
    cmd.extend(["--mode", "plan"])
elif permission_mode == "auto":
    cmd.append("--force")
```

Worth a test that asserts no `--mode` and no `--force` when `permission_mode` is unset, since the current parametrized case pins the opposite.

dchaudhari7177: You are right on every point. I had assumed an unset `--mode` meant `ask` and never checked that against the CLI help. Fixed in `01141cb`, exactly as you suggested: the `else` is gone, so an unset `permission_mode` emits neither `--mode` nor `--force`.

The `(None, ["--mode", "plan"])` case that pinned the wrong behaviour is removed. It is replaced by `test_an_unset_permission_mode_leaves_the_cli_default`, parametrized over `{}` and `{"permission_mode": None}`, which asserts neither flag appears. With `cursor.py` stashed back to the previous head, both cases fail; with the fix, `test_harness_provider_cursor.py` + `test_harness_doctor.py` pass (29). The comment in `cursor.py` now says the unset case follows the CLI default like the other adapters, and that schema runs depend on it.

dchaudhari7177:
Took both smaller notes as well, in the same commit (`01141cb`):

- **Docs:** `docs/harness-providers.md` now has a `cursor` row in the install table (`curl https://cursor.com/install -fsS | bash`, required CLI `agent`, auth `agent login` or `CURSOR_API_KEY`). The note below the table now says `cursor`, like `grok`, is Python-SDK only, and that the binary is `agent`, not `cursor`.
- **Fixture:** the `assistant` event in `RESULT_STREAM` now has the shape the CLI emits (`{"type":"assistant","message":{"role":"assistant","content":[{"type":"text","text":…}]},"session_id":…}`). Final text and chat id still come off the `result` event, and every test passes against the new shape.

Ruff (the pinned `v0.15.22`) is clean on both Python files.

Thanks for testing this against the real CLI. The `--mode plan` default would have broken every schema run.

santoshkumarradha (approved):
Thanks for correcting the default permission behavior. Leaving the mode unset preserves the Cursor CLI default and allows schema runs to write their output, while the explicit plan and auto mappings remain intact. The focused provider, doctor, factory, defaults, and packaging tests pass locally, and Ruff is clean.

sdk/python/agentfield/harness/providers/cursor.py
AbirAbbas: Verified on 01141cb6 merged with current main: an unset `permission_mode` now emits neither `--mode` nor `--force`, and putting the old else branch back makes `test_an_unset_permission_mode_leaves_the_cli_default` fail on both parametrizations. Resolving.

AbirAbbas (approved):
Checked this against the installed Cursor CLI (2026.09.08) rather than just the tests: every flag the provider can emit is in `agent --help`, bare `-p` is the write+shell mode as the fix assumes, the key only ever travels via `CURSOR_API_KEY`, and the stream-json fixture matches the CLI's bundled output formatter. Full sdk-python gates pass on the merge with main (ruff 0.15.22, run_pytest.sh 2428 passed, websockets 12/15 compat), and the harness lines the patch gate can't see are 100% covered by the cursor/doctor tests. Code is good to go.

AbirAbbas:
Closing only because the CLA is still unsigned (status has been pending since Sep 6, reminder on Sep 8) and we keep the open queue to things that can merge. The code itself is approved by both of us and needs no further changes. Sign it at https://cla-assistant.io/Agent-Field/agentfield?pullRequest=1042 and reopen this PR (or ping me) and it goes straight into the merge queue.

PR #1044, opened 2026-09-07
fix(sdk/go): return isolated session definitions

emecii:
## Summary

`Agent.SessionDefinitions()` returned copied `SessionDefinition` structs, but
their slice and metadata fields still shared state with the registered
definitions. This makes each returned definition an isolated snapshot,
including recursively nested metadata maps and slices, overlapping slice views,
and the `TurnDetection` configuration and its optional scalar pointers.
Nil, zero, and false values are preserved.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs only
- [ ] Tests only
- [ ] CI / tooling
- [ ] Breaking change

## Test plan

Local validation by Codex on Go 1.23.12 / macOS arm64 (`TMPDIR=/private/tmp`):

- [x] `cd sdk/go && go test ./... -count=1` (all eight SDK packages)
- [x] Focused session/turn-detection race tests repeated 20 times
- [x] `go build ./...`
- [x] `cd sdk/go && go vet ./agent`
- [x] `gofmt -d agent/session.go agent/session_test.go` (no output)
- [x] `go mod tidy -diff` (no diff)
- [x] The new regression fails against the previous `SessionDefinitions()` implementation and passes with this change.

## Test coverage

- [x] I ran tests for the changed Go SDK surface locally.
- [ ] The coverage gate is green in CI before requesting review.

## Checklist

- [x] I have read [CONTRIBUTING.md](docs/CONTRIBUTING.md).
- [x] Commit follows conventional-commits style.
- [ ] Related issue: no existing issue covers this focused bug fix.

## AI assistance

This change was prepared with OpenAI Codex. The validation listed above was run locally.
Repository-wide scripts were not completed because the local Python environment lacks pytest/ruff.
Hosted checks for the updated head await maintainer workflow approval.

### DIFF
--- a/sdk/go/agent/session.go
+++ b/sdk/go/agent/session.go
@@ -1,5 +1,7 @@
 package agent
 
+import "reflect"
+
 type SessionDefinition struct {
 	Name         string         `json:"name"`
 	Provider     string         `json:"provider"`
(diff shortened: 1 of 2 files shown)


santoshkumarradha (dismissed):
Looks good. The snapshot now isolates all session slices plus nested map and slice metadata, including cycles, while preserving nil values. The focused defensive-snapshot test passes locally.

santoshkumarradha:
The implementation and all CI checks look good. The CLA is the only remaining step; once that’s signed, this should be good to go.

sdk/go/agent/session.go
```diff
@@ -97,7 +99,81 @@ func (a *Agent) RegisterSession(name string, provider string, transport string,
...
+		iter := value.MapRange()
+		for iter.Next() {
+			result.SetMapIndex(iter.Key(), cloneSessionMetadataValue(iter.Value(), copied))
+		}
+		return result
+	case reflect.Slice:
+		if value.IsNil() {
+			return reflect.Zero(value.Type())
+		}
+		ref := sessionMetadataCopyReference{kind: value.Kind(), typeOf: value.Type(), ptr: value.Pointer()}
```
AbirAbbas: `value.Pointer()` on a slice is the address of the backing array's first element, not an identity for the slice itself. So a slice and any prefix of it hash to the same `sessionMetadataCopyReference`, and whichever one the map range visits first gets cached and then handed to both.

```go
shared := []any{"a", "b", "c"}
a.RegisterSession("alias", "openai", "webrtc", agent.WithSessionMetadata(map[string]any{
    "all":  shared,
    "head": shared[:1],
}))
d := a.SessionDefinitions()[0]
// d.Metadata["head"] comes back as ["a","b","c"], or d.Metadata["all"] comes back as ["a"]
```

I ran that 400 times against this branch: `head` came back with the wrong length 354 times and `all` 46 times (map iteration order picks the winner). The same program on `main` is 0/400, because nothing is copied there and each slice keeps its own header. So the snapshot that goes into the registration payload can silently gain or drop metadata entries that `main` gets right, and which way it goes changes run to run.

Keying on the slice header rather than the array start fixes it:

```go
type sessionMetadataCopyReference struct {
	kind   reflect.Kind
	typeOf reflect.Type
	ptr    uintptr
	length int
	cap    int
}
```

with `length: value.Len(), cap: value.Cap()` set in the slice branch (left zero for maps). I tried exactly that locally: the overlap case goes to 0/400 wrong and cycle detection still works, including a self-referential `s[0] = s`, because the back-edge has the same ptr/len/cap.

Worth a regression test for it as well - register `{"all": s, "head": s[:1]}` and assert both lengths survive `SessionDefinitions()`.

AbirAbbas (commented):
I built a standalone program against this branch's `sdk/go` that registers session metadata with a map inside a slice inside a map, a typed `[]string`, a `map[string]string`, `[]byte`, a reference cycle and nil values, mutates every level of the `SessionDefinitions()` result and then re-reads the registry. On this branch every one of those stays isolated, nils keep their typed-nil-ness, the cycle stays a cycle and terminates, and the JSON of the definition is unchanged; the same program run against `main` leaks all of them, so the fix does what it says. `go test -race ./agent/...` is clean and patch coverage on the touched lines is 80%.

One real problem though, left inline: two metadata slices that share a backing array collide in the copy memo and one of them comes back with the wrong length. It's nondeterministic and silent, and it's a way for the payload to be wrong that doesn't exist on `main`, so I'd rather see it fixed here than shipped.

Separately, and fine to leave as is: the `default:` branch hands back pointers, structs with reference fields and arrays as they are, so `map[string]any{"cfg": &Cfg{...}}` still aliases into the registry. That's outside what this PR set out to do, but a line on `cloneSessionMetadataValue` saying only maps and slices are deep-copied would save the next reader the experiment.

sdk/go/agent/session.go
emecii: Fixed in 03e0491b. Slice memo keys now include length and capacity, with a regression for shared backing arrays. `go test -race ./agent/...` passes.

emecii:
CLA signed.

santoshkumarradha (dismissed):
Looks good. The updated memo key preserves overlapping slice views, and both defensive snapshot regressions pass locally under the race detector. I also approved the pending fork workflows now that the CLA is signed.

emecii:
Merged current `main` (`2180e30c`) in `146e5fa3` to resolve the conflict with #1056. Both existing snapshot regressions and the upstream turn-detection registration test are retained unchanged, including the reviewed slice memo key fix.

The integration also needs to detach the newly introduced `TurnDetection` struct and its five scalar pointers. Added that copy while preserving nil, zero, and false values. The new regression reproduced aliasing before the fix and now passes for defaults, explicit zero/false, semantic VAD, and no turn detection.

Validation with Go 1.23.12 on macOS arm64 (`TMPDIR=/private/tmp`): full Go SDK `go test ./... -count=1`, 20 repeated focused race runs (`Session|TurnDetection`), build, vet, tidy with no diff, and formatting all pass. The repository-wide script remains unavailable locally because pytest is missing; `make fmt tidy` also stops on missing ruff. No unrelated formatting changes are included. The effective diff against current main remains limited to the two Go session files.

santoshkumarradha (approved):
I rechecked the rebased head. The snapshot clone still preserves shared and cyclic metadata safely, distinguishes overlapping slice views, and now includes the current turn-detection fields from main. The focused Go tests, vet, and diff check pass locally.

santoshkumarradha:
recheck

AbirAbbas (approved):
Re-verified on the merged tree (146e5fa3 + current main): sdk-go coverage surface and patch gate pass (84% of the touched session.go lines), `go test -race ./agent/...` clean, and the three regressions each fail when their fix is reverted in isolation — pointer-only slice keys break the overlapping-slice test, dropping the TurnDetection clone breaks its detach test, main's session.go breaks the snapshot test. Nothing new landed on main under sdk/go/agent since your merge, so no further fields need copying. CLA status was missing on the new head; I triggered the recheck and it's green now.

PR #1046, opened 2026-09-08
fix(storage): workflow reaper consults execution activity clock

ddbaron:
## Proposal

Keep the workflow cleanup path from reaping a workflow while its paired execution still has recent activity. The query should check the workflow timestamp and the timestamp on the paired active execution before declaring the workflow stale.

## Observed failure

Each execution has two activity timestamps. Heartbeats and status writes update `executions.updated_at`. `MarkStaleWorkflowExecutions` reads `workflow_executions.updated_at`, which remains unchanged during a leaf wait.

Four coder runs were reaped 10 to 13 minutes after starting on a 10-minute fuse. Heartbeats were arriving every 90 seconds. One run was reaped 57 seconds after its latest heartbeat. File commits continued after the reap, and the late completion updates returned HTTP 409 because the records were already terminal.

## Implementation

This PR makes the workflow reaper join each workflow row to its paired active execution. It requires the workflow timestamp and the execution timestamp to be older than the cutoff before selecting the workflow for reaping.

The query still allows cleanup when the paired execution is absent or already terminal. The existing workflow-to-execution terminal synchronization remains in place.

The regression tests cover two cases:

1. Recent activity on `executions.updated_at` keeps an old workflow row running.
2. Rows that are old on both timestamps are reaped, and both records receive the existing terminal timeout state.

The change is on the read path, so existing heartbeat and note writes do not create another database write. Updating both timestamps on every activity event is another reasonable implementation. I am open to changing to that approach if it fits the control plane better.

## Validation and evidence

The storage and handler tests for this path pass, along with `go vet` and formatting checks. At the time of this update, upstream checks report 27 passed, 0 failed, and 22 skipped.

The local reproduction is documented in the comment below. It includes the sampled execution IDs, timestamps, the 57-second interval between a heartbeat and a reap, the HTTP 409 completion errors, and the post-fix smoke results.

This PR is a working implementation for maintainer review. The goal is to keep the existing stale cleanup behavior for genuinely inactive workflows while allowing recent activity in the paired execution to count.

### DIFF
--- a/control-plane/internal/storage/execution_records.go
+++ b/control-plane/internal/storage/execution_records.go
@@ -1185,7 +1185,13 @@ func parseTimeString(value string) (time.Time, error) {
 // is reaped first, which makes its parent childless and eligible on the next
 // sweep, and so on up. Nothing is stuck forever; it just takes one sweep per
 // level.
+// The conditional UPDATE re-evaluates the staleness predicates so a row that
+// gains activity after selection is left alone.
 func (ls *LocalStorage) MarkStaleExecutions(ctx context.Context, staleAfter time.Duration, limit int) (int, error) {
+	return ls.markStaleExecutions(ctx, staleAfter, limit, nil)
+}
+
+func (ls *LocalStorage) markStaleExecutions(ctx context.Context, staleAfter time.Duration, limit int, afterCandidateSelection func()) (int, error) {
 	if limit <= 0 {
 		return 0, nil
 	}
@@ -1197,6 +1203,7 @@ func (ls *LocalStorage) MarkStaleExecutions(ctx context.Context, staleAfter time
 
 	db := ls.requireSQLDB()
 	tsExpr := ls.staleTimestampExpr("COALESCE(updated_at, created_at, started_at)")
+	executionUpdateTSExpr := ls.staleTimestampExpr("COALESCE(e.updated_at, e.created_at, e.started_at)")
 	cutoffExpr := ls.staleTimestampExpr("?")
 	rows, err := db.QueryContext(ctx, `
 		SELECT execution_id, started_at
@@ -1236,6 +1243,12 @@ func (ls *LocalStorage) MarkStaleExecutions(ctx context.Context, staleAfter time
 		return 0, nil
 	}
 
+	// Package tests use this seam to make the candidate-selection-to-update
+	// interleaving deterministic; production callers leave it nil.
+	if afterCandidateSelection != nil {
+		afterCandidateSelection()
+	}
+
 	tx, err := db.BeginTx(ctx, nil)
 	if err != nil {
 		return 0, fmt.Errorf("begin stale execution transaction: %w", err)
@@ -1243,9 +1256,16 @@ func (ls *LocalStorage) MarkStaleExecutions(ctx context.Context, staleAfter time
 	defer rollbackTx(tx, "MarkStaleExecutions")
 
 	updateStmt, err := tx.PrepareContext(ctx, `
-		UPDATE executions
+		UPDATE executions AS e
 		SET status = ?, error_message = ?, completed_at = ?, duration_ms = ?, updated_at = ?
-		WHERE execution_id = ? AND status IN ('running', 'pending', 'queued')`)
+		WHERE e.execution_id = ?
+		  AND e.status IN ('running', 'pending', 'queued')
+		  AND `+executionUpdateTSExpr+` <= `+cutoffExpr+`
+		  AND NOT EXISTS (
+		      SELECT 1 FROM executions c
+		      WHERE c.parent_execution_id = e.execution_id
+		        AND c.status IN ('running', 'pending', 'queued')
+		  )`)
 	if err != nil {
 		return 0, fmt.Errorf("prepare stale execution update: %w", err)
 	}
@@ -1273,6 +1293,7 @@ func (ls *LocalStorage) MarkStaleExecutions(ctx context.Context, staleAfter time
 			durationMS,
 			now,
 			rec.id,
+			cutoff,
 		)
 		if err != nil {
 			return 0, fmt.Errorf("update stale execution %s: %w", rec.id, err)
--- a/control-plane/internal/storage/execution_records.go
+++ b/control-plane/internal/storage/execution_records.go
@@ -1295,8 +1295,10 @@ func (ls *LocalStorage) MarkStaleExecutions(ctx context.Context, staleAfter time
 }
 
 // MarkStaleWorkflowExecutions updates workflow executions stuck in non-terminal states
-// when their updated_at timestamp exceeds the staleAfter threshold. This catches orphaned
-// child executions whose parent failed without cascading cancellation.
+// when both the workflow and its paired execution activity timestamps exceed the
+// staleAfter threshold. This catches orphaned child executions whose parent failed
+// without cascading cancellation while allowing activity recorded in either table
+// to keep the execution alive.
 //
 // See MarkStaleExecutions for the updated_at invariant, the COALESCE fallback
 // rationale, and why a row with a non-terminal child is skipped rather than reaped.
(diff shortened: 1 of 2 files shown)


ddbaron:
## Evidence from the local reproduction

I read these values directly from the local SQLite database and the execution records. The rows used UTC timestamps.

Three sampled coder executions showed the problem:

- `exec_20260908_063449_vawzsj9v`: the workflow row started at `11:34:49Z` and was reaped at `11:47:46Z`, 12m57s later. The paired execution row continued receiving heartbeat notes through `11:50:13Z`, including a `coder complete` note three minutes after the workflow had been reaped.
- `exec_20260908_064218_mrs177j6`: the workflow row was reaped 10m28s after its start. The execution row continued receiving heartbeat updates for about 27 minutes after that.
- `exec_20260908_065048_xrszw2m7`: the workflow row was reaped about 12 minutes after its start. The execution row continued receiving activity after the reap.

The heartbeat was fresh when the cleanup ran. For example, one `harness/heartbeat` note arrived at `11:46:49Z` and the workflow was reaped at `11:47:46Z`, 57 seconds later. The late completion updates returned HTTP 409 with `execution is already in a terminal state`.

During the investigation I checked the other likely explanations. All 66 same-day rows used `+00:00` offsets, the `julianday` comparison behaved as expected, the 10-minute fuse matched the configuration, environment, and database, and every timed-out execution had a model recorded.

After the node heartbeat and workflow-clock changes, a full build survived a 17-minute silent coder interval with no no-activity reap. A separate 27-execution short-fuse round also completed with no timeout signatures.

control-plane/internal/storage/execution_records.go
```diff
@@ -1311,21 +1313,29 @@ func (ls *LocalStorage) MarkStaleWorkflowExecutions(ctx context.Context, staleAf
...
+		SELECT w.execution_id, w.started_at
 		FROM workflow_executions w
-		WHERE status IN ('running', 'pending', 'queued', 'waiting')
-		  AND `+tsExpr+` <= `+cutoffExpr+`
-		  AND COALESCE(approval_status, '') != 'pending'
+		LEFT JOIN executions e
+		  ON e.execution_id = w.execution_id
+		 AND e.status IN ('running', 'pending', 'queued', 'waiting')
+		WHERE w.status IN ('running', 'pending', 'queued', 'waiting')
+		  AND `+workflowTSExpr+` <= `+cutoffExpr+`
```
santoshkumarradha, line 1354: These timestamps are checked during candidate selection, but the later UPDATE only rechecks status. If a heartbeat updates executions.updated_at between those statements, the workflow is still timed out. Could we repeat the stale predicates in the conditional UPDATE, count only RowsAffected > 0, and add a test for that interleaving?

santoshkumarradha (dismissed):
The focused storage tests and CI pass, and the join handles heartbeats that arrive before candidate selection. There is still a narrow concurrent-heartbeat race in the final update that can produce the same false timeout, so I think we should close that before merging.

control-plane/internal/storage/execution_records.go
AbirAbbas: Verified this at the current head and pushed one follow-up commit for the sibling reaper.

Your ask is covered for the workflow reaper by ddbaron's `bcdfb064`: the conditional UPDATE now repeats the workflow activity clock, the paired non-terminal execution's clock, `approval_status`, and the child-liveness guard, and only `RowsAffected > 0` is counted and mirrored into `executions`. The regression test is `TestMarkStaleWorkflowExecutions_ActivityAfterSelectionSkipsUpdate`, which lands a real execution note through `UpdateExecutionRecord` in the window between candidate selection and the update.

`MarkStaleExecutions` had the same shape — its UPDATE also only re-checked status — so my commit gives it the same treatment (activity clock and child guard re-evaluated, same post-selection seam). Tests: `TestMarkStaleExecutions_ActivityAfterSelectionSkipsUpdate`, plus `TestMarkStaleExecutions_SeamWithoutActivityStillReaps` as the control.

Anti-tautology, with only the two UPDATE guards stripped and the seams and tests kept:

```
--- FAIL: TestMarkStaleWorkflowExecutions_ActivityAfterSelectionSkipsUpdate  expected 0, actual 1
--- FAIL: TestMarkStaleExecutions_ActivityAfterSelectionSkipsUpdate          expected 0, actual 1
--- PASS: TestMarkStaleExecutions_SeamWithoutActivityStillReaps
--- PASS: TestMarkStaleWorkflowExecutions_SilentRowsReapAndSyncExecution
```

End to end against a real control plane (20s stale fuse, 5s sweep, a real agent node, three dispatched executions, `POST /executions/note` every 5s as the heartbeat):

- on `main`: the heart-beating run's workflow row goes `timeout` 31s in and its late completion callback gets HTTP 409.
- on this branch: the same run is still `running` at 91s — `workflow_executions.updated_at` frozen at dispatch, `executions.updated_at` advancing with every note — and the late completion returns 200. The silent run was still reaped at 30s and the one whose heartbeats stopped at 25s was reaped at 51s, so the reaper is not just switched off.

AbirAbbas:
@ddbaron pushed one commit on top of your branch so this can land. `MarkStaleExecutions` had the same select-then-update race you just closed in the workflow reaper — its conditional UPDATE also only re-checked status — so it now re-evaluates the activity clock and the child guard the same way, behind the same post-selection seam. Your two commits are untouched, and the two tests you wrote still pass unchanged.

I reproduced your report end to end on a real control plane with a 20s stale fuse and a 5s sweep, driving a real agent node and using `POST /executions/note` as the heartbeat: on `main` the heart-beating run's workflow row is reaped 31s in and the completion callback gets 409; on this branch the same run is still `running` at 91s with `workflow_executions.updated_at` frozen at dispatch and `executions.updated_at` advancing, and the completion returns 200 — while a silent run is still reaped at 30s and one whose heartbeats stop at 25s is reaped at 51s.

Separately, and not caused by this PR: on the PostgreSQL backend neither reaper runs at all. `sqlTx` has no `PrepareContext` override, so the `?` placeholders in these prepared UPDATEs are never rebound to `$n`. Against a real PostgreSQL 14 with the storage layer in postgres mode, on `main` (v0.1.138-rc.13):

```
MarkStaleExecutions         -> n=0 err=prepare stale execution update: ERROR: syntax error at or near "," (SQLSTATE 42601)
MarkStaleWorkflowExecutions -> n=0 err=prepare stale workflow execution update: ERROR: syntax error at or near "," (SQLSTATE 42601)
```

and the row that is stale on both clocks stays `running`. So stale cleanup is a no-op in cloud mode today, and this fix only changes behaviour on the local/SQLite backend until that is addressed. It wants its own PR — and whoever picks it up should look at `RetryStaleWorkflowExecutions` in the same pass, because it still selects on the workflow clock alone and runs before the reaper, so switching the path on would start re-dispatching live runs wherever `max_retries > 0`.

AbirAbbas (approved):
Verified at 8c2395ec.

Gates run locally on the control-plane surface exactly as CI runs them (gofmt on the touched files, web UI `npm ci && npm run build`, `go build ./...`, `go vet ./...`, the full `go test -tags sqlite_fts5 ./...` coverage sweep, and the patch gate): all green, patch coverage 100% on the 63 touched lines, storage package coverage 86.6%.

Anti-tautology: with only the two conditional-UPDATE guards stripped and the seams and tests kept, both interleaving tests fail with expected 0 / actual 1 while `SeamWithoutActivityStillReaps` and `SilentRowsReapAndSyncExecution` still pass. Against origin/main's storage source, the PR's own `ExecutionActivityProtectsWorkflow` fails with the original symptom.

Manual: real control plane with a 20s stale fuse and a 5s sweep, a real agent node, three dispatched executions, notes as heartbeats. On main the heart-beating run is reaped at t=31s and its late completion gets 409; on this branch it is still running at t=91s with the workflow clock frozen and the execution clock advancing, completion returns 200, the silent run is still reaped at t=30s and the one whose heartbeats stop at t=25s is reaped at t=51s.

Caveat recorded in the PR comment, not a blocker for this change: the reapers never execute on the Postgres backend today (`sqlTx` has no `PrepareContext`, so `?` is never rebound), so this only changes behaviour on the local backend until that is fixed separately.

PR #1051, opened 2026-09-10
fix(storage): rebind postgres tx prepares and guard stale retries

ddbaron:
## Problem

On the PostgreSQL backend the stale cleanup path does nothing. Both reapers return `n=0` with:

```
prepare stale execution update: ERROR: syntax error at or near "," (SQLSTATE 42601)
```

`sqlTx` embeds `*sql.Tx` and overrides `Exec`, `Query`, and `QueryRow` so `?` placeholders are rebound to `$n`, but it never overrode `PrepareContext` or `Prepare`. A prepared statement created inside a transaction therefore went to the driver with `?` placeholders intact and failed to prepare on PostgreSQL, which is why `MarkStaleExecutions` and `MarkStaleWorkflowExecutions` never reaped anything in cloud mode.

`RetryStaleWorkflowExecutions` has a related gap. It selects on the workflow clock alone and runs before the reaper, so once the path works it would start re-dispatching workflows whose paired execution is still active.

## Change

- Adds `sqlTx.PrepareContext` and the deprecated `sqlTx.Prepare`, both rebinding the query before delegating to `*sql.Tx`, matching what `sqlDatabase` already did.
- Makes `RetryStaleWorkflowExecutions` consult the paired `executions` activity clock, the same way `MarkStaleWorkflowExecutions` does.
- Repeats the status, retry-count, workflow-clock, and paired-execution-clock predicates in the conditional retry update, so activity arriving between selection and update leaves the workflow untouched. `RowsAffected > 0` gating and retry-count semantics are unchanged.

## Verification

Focused validation passes:

```sh
cd control-plane
go test ./internal/storage ./internal/handlers -count=1
go vet ./...
gofmt -d internal/storage/sql_helpers.go internal/storage/execution_records.go internal/storage/coverage_misc_helpers_test.go internal/storage/retry_stale_test.go
git diff --check
```

The full control-plane suite was also run. Storage and handlers pass; unrelated pre-existing failures remain in `internal/core/services` and `internal/packages` (environment and path expectations). No unrelated package was changed.

### Live PostgreSQL

Run against a throwaway PostgreSQL 16 container, loopback only on a dynamically allocated port, per-test throwaway database, container and volume removed afterwards.

The storage suite reads `POSTGRES_TEST_URL`. With it set, `go test ./internal/storage/ -run TestPostgres -v -count=1` gives **18 passed, 0 failed, 1 skipped** (the skip is the pre-existing unconditional `TestPostgresStorage_DatabaseCreation`). The whole storage package also passes with a live database configured.

Evidence that this pins the fix rather than merely passing next to it:

- Pre-fix reproduction: sending the reaper's `?` placeholders straight to PostgreSQL fails with `SQLSTATE 42601`, the error reported above.
- Post-fix, the text handed to the driver is rebound, with `$1..$8` for the workflow and retry UPDATEs and `$1..$6` for the execution sync UPDATE.
- End to end on the live database: a workflow stale on both clocks is reaped (`status=timeout`), a workflow whose paired execution has fresh activity survives (`status=running`), the retry guard re-dispatches a genuinely stale workflow (`status=pending retry_count=1`) and leaves a freshly heartbeating one alone, and the terminal-paired-execution branch still reaps.
- Removing the two `sqlTx.Prepare*` overrides on this branch makes the three new live tests fail with the original error.

Known harness gaps left untouched: `coverage_postgres_test.go` sits behind the `integration` build tag with a hardcoded DSN, and `locks_test.go`'s `setupPostgresTestStorage` returns nil unconditionally, so both skip regardless of `POSTGRES_TEST_URL`. `go test -race` for this package also aborts in the `boltdb/bolt` v1.3.1 dependency during SQLite initialization, independent of this change.

One deployment note: the workflow-side reaper only started consulting the execution clock in #1046, so this fix changes behaviour only on a control plane that includes that change.

### DIFF
--- a/control-plane/internal/storage/execution_records.go
+++ b/control-plane/internal/storage/execution_records.go
@@ -1573,8 +1573,13 @@ func (ls *LocalStorage) markAgentExecutionsOrphaned(ctx context.Context, agentNo
 
 // RetryStaleWorkflowExecutions finds stale workflow executions that haven't exceeded
 // maxRetries and resets both workflow_executions and executions back to "pending"
-// so the paired records stay in sync for the retry path.
+// so the paired records stay in sync for the retry path. The candidate predicates
+// are repeated in the conditional update so activity after selection prevents a retry.
 func (ls *LocalStorage) RetryStaleWorkflowExecutions(ctx context.Context, staleAfter time.Duration, maxRetries int, limit int) ([]string, error) {
+	return ls.retryStaleWorkflowExecutions(ctx, staleAfter, maxRetries, limit, nil)
+}
+
+func (ls *LocalStorage) retryStaleWorkflowExecutions(ctx context.Context, staleAfter time.Duration, maxRetries int, limit int, afterCandidateSelection func()) ([]string, error) {
 	if limit <= 0 || maxRetries <= 0 {
 		return nil, nil
 	}
@@ -1584,17 +1589,22 @@ func (ls *LocalStorage) RetryStaleWorkflowExecutions(ctx context.Context, staleA
 
 	cutoff := time.Now().UTC().Add(-staleAfter)
 	db := ls.requireSQLDB()
-	tsExpr := ls.staleTimestampExpr("COALESCE(updated_at, created_at, started_at)")
+	workflowTSExpr := ls.staleTimestampExpr("COALESCE(w.updated_at, w.created_at, w.started_at)")
+	executionTSExpr := ls.staleTimestampExpr("COALESCE(e.updated_at, e.created_at, e.started_at)")
 	cutoffExpr := ls.staleTimestampExpr("?")
 
 	rows, err := db.QueryContext(ctx, `
-		SELECT execution_id
-		FROM workflow_executions
-		WHERE status IN ('running', 'pending', 'queued')
-		  AND retry_count < ?
-		  AND `+tsExpr+` <= `+cutoffExpr+`
-		ORDER BY `+tsExpr+` ASC
-		LIMIT ?`, maxRetries, cutoff, limit)
+		SELECT w.execution_id
+		FROM workflow_executions w
+		LEFT JOIN executions e
+		  ON e.execution_id = w.execution_id
+		 AND e.status IN ('running', 'pending', 'queued', 'waiting')
+		WHERE w.status IN ('running', 'pending', 'queued')
+		  AND w.retry_count < ?
+		  AND `+workflowTSExpr+` <= `+cutoffExpr+`
+		  AND (e.execution_id IS NULL OR `+executionTSExpr+` <= `+cutoffExpr+`)
+		ORDER BY `+workflowTSExpr+` ASC
+		LIMIT ?`, maxRetries, cutoff, cutoff, limit)
 	if err != nil {
 		return nil, fmt.Errorf("query retriable workflow executions: %w", err)
 	}
@@ -1616,6 +1626,12 @@ func (ls *LocalStorage) RetryStaleWorkflowExecutions(ctx context.Context, staleA
 		return nil, nil
 	}
 
+	// Package tests use this seam to make the candidate-selection-to-update
+	// interleaving deterministic; production callers leave it nil.
+	if afterCandidateSelection != nil {
+		afterCandidateSelection()
+	}
+
 	tx, err := db.BeginTx(ctx, nil)
 	if err != nil {
 		return nil, fmt.Errorf("begin retry transaction: %w", err)
@@ -1626,13 +1642,29 @@ func (ls *LocalStorage) RetryStaleWorkflowExecutions(ctx context.Context, staleA
 	retryReason := "auto-retry after stale timeout"
 
 	workflowStmt, err := tx.PrepareContext(ctx, `
-		UPDATE workflow_executions
+		UPDATE workflow_executions AS w
 		SET status = 'pending',
 		    retry_count = retry_count + 1,
 		    error_message = ?,
 		    completed_at = NULL,
 		    updated_at = ?
-		WHERE execution_id = ? AND status IN ('running', 'pending', 'queued')`)
+		WHERE w.execution_id = ?
+		  AND w.status IN ('running', 'pending', 'queued')
+		  AND w.retry_count < ?
+		  AND `+workflowTSExpr+` <= `+cutoffExpr+`
+		  AND (
+		      NOT EXISTS (
+		          SELECT 1 FROM executions e
+		          WHERE e.execution_id = w.execution_id
+		            AND e.status IN ('running', 'pending', 'queued', 'waiting')
+		      )
+		      OR EXISTS (
+		          SELECT 1 FROM executions e
+		          WHERE e.execution_id = w.execution_id
+		            AND e.status IN ('running', 'pending', 'queued', 'waiting')
+		            AND `+executionTSExpr+` <= `+cutoffExpr+`
+		      )
+		  )`)
 	if err != nil {
 		return nil, fmt.Errorf("prepare retry statement: %w", err)
 	}
@@ -1653,7 +1685,7 @@ func (ls *LocalStorage) RetryStaleWorkflowExecutions(ctx context.Context, staleA
 
 	var retried []string
 	for _, id := range ids {
-		result, err := workflowStmt.ExecContext(ctx, retryReason, now, id)
+		result, err := workflowStmt.ExecContext(ctx, retryReason, now, id, maxRetries, cutoff, cutoff)
 		if err != nil {
 			return retried, fmt.Errorf("retry workflow execution %s: %w", id, err)
 		}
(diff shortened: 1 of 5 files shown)


ddbaron:
@santoshkumarradha  This closes the gap that came up when #1046 merged. In that thread you noted that on the PostgreSQL backend neither stale reaper runs at all, because `sqlTx` has no `PrepareContext` override, so the `?` placeholders in these prepared UPDATEs were never rebound to `$n`, and you asked whoever picked it up to look at `RetryStaleWorkflowExecutions` in the same pass.

What I understood from that:

1. Give `sqlTx` the same prepare-time rebinding that `sqlDatabase` already has, so prepared statements created inside a transaction work on PostgreSQL.
2. Check the retry path, since it still selected on the workflow clock alone and runs before the reaper, so switching the path on would start re-dispatching live runs wherever retries are enabled.

What I did:

- Added `sqlTx.PrepareContext` and the deprecated `sqlTx.Prepare`, both rebinding before delegating to `*sql.Tx`.
- Made `RetryStaleWorkflowExecutions` consult the paired execution activity clock the same way the workflow reaper does, and repeat the stale predicates in its conditional update so activity arriving between selection and update is respected. `RowsAffected` gating and retry-count semantics are unchanged.
- Verified it against a live PostgreSQL 16 rather than leaving the connection-backed tests skipped. I reproduced the original `SQLSTATE 42601` failure first, then drove both reapers and the retry path end to end and confirmed that a row stale on both clocks is reaped while one with a fresh paired execution survives. Removing the two `Prepare*` overrides makes the new live tests fail with the original error. Counts and the captured statements are in the description.

Two pre-existing gaps I deliberately left alone: a Postgres test file sits behind the `integration` build tag with a hardcoded DSN, and `locks_test.go`'s `setupPostgresTestStorage` returns nil unconditionally, so those tests skip however `POSTGRES_TEST_URL` is set.

If you would rather this were done differently, tell me and I will change it.

control-plane/internal/storage/execution_records.go
```diff
@@ -1626,13 +1642,29 @@ func (ls *LocalStorage) RetryStaleWorkflowExecutions(ctx context.Context, staleA
...
+		      NOT EXISTS (
+		          SELECT 1 FROM executions e
+		          WHERE e.execution_id = w.execution_id
+		            AND e.status IN ('running', 'pending', 'queued', 'waiting')
+		      )
+		      OR EXISTS (
+		          SELECT 1 FROM executions e
+		          WHERE e.execution_id = w.execution_id
+		            AND e.status IN ('running', 'pending', 'queued', 'waiting')
+		            AND `+executionTSExpr+` <= `+cutoffExpr+`
```
santoshkumarradha, line 1667: This protects the workflow UPDATE, but the paired execution is updated by a later statement that only rechecks `status`. On Postgres, a heartbeat can land after this predicate is evaluated and before that second UPDATE; the retry then overwrites the fresh execution back to `pending`. Please repeat the execution staleness predicate on the second update (with a regression for the between-statements interleaving), or lock/update the pair so the decision is atomic.

santoshkumarradha (changes requested):
The Postgres prepare fix and the selection-to-update guard are both solid. There is still one heartbeat window in the retry path that needs closing before this is safe to merge.

control-plane/internal/storage/execution_records.go
ddbaron: Fixed in 5d8575b6. The retry now repeats the execution staleness predicate on the second update, so a heartbeat landing between the two statements can no longer be flattened back to pending. The regression interleaves a heartbeat between them and fails against the previous revision, and it runs both on SQLite and against a live Postgres instance. CI is green. Thanks for the precise pointer.

control-plane/internal/storage/execution_records.go
```diff
@@ -1665,7 +1703,7 @@ func (ls *LocalStorage) RetryStaleWorkflowExecutions(ctx context.Context, staleA
 			continue
 		}
 
-		if _, err := executionStmt.ExecContext(ctx, retryReason, now, id); err != nil {
+		if _, err := executionStmt.ExecContext(ctx, retryReason, now, id, cutoff); err != nil {
```
santoshkumarradha: Please check RowsAffected here and avoid committing the workflow retry when the execution update loses to a fresh heartbeat. In the new regression, this statement updates zero rows, but the workflow has already moved to pending, retry_count is incremented, and the ID is appended to retried. That leaves workflow=pending and execution=running while reporting a successful retry, which can dispatch duplicate work. The two state changes need an all-or-nothing outcome for this candidate.

santoshkumarradha (changes requested):
The new staleness predicate closes the overwrite window, but the retry still needs to be atomic across both records. Right now the heartbeat case is reported as retried even though the paired execution remains running.

control-plane/internal/storage/execution_records.go
ddbaron: Both records now move together per candidate. The execution update repeats the staleness predicate, so if it matches zero rows the workflow row is not committed, the retry count stays as it was, and the candidate is not reported as retried. A heartbeat arriving between the two statements can no longer leave a retried workflow with a running execution.

New tests cover the interleavings:

- a heartbeat landing between the two statements: the workflow keeps the heartbeat timestamp, stays running, and the candidate is absent from the returned set;
- a three-candidate batch with heartbeats on the first and last: only the middle one moves, and the returned ids match just that one;
- a statement failure mid-batch: the rollback discards the candidate that had already moved, so nothing counts as retried;
- both reapers still separate stale rows from fresh ones;
- live Postgres holds a concurrent transaction open, so the heartbeat commits between the two statements.

`go test ./control-plane/internal/storage/...` passes. The live Postgres tests run when `POSTGRES_TEST_URL` is set, and skip with that reason when it is not.

santoshkumarradha (approved):
Thanks, this closes the retry race cleanly. The per-candidate savepoint makes the workflow and execution updates all-or-nothing, preserves the existing workflow-only recovery paths, and reports only candidates that actually moved. The focused storage suite and go vet pass locally, including the new batch and interleaving regressions.

santoshkumarradha (approved):
Reconfirmed on the current head. The transactional rebind and savepoint-based stale retry update keep the workflow and paired execution atomic, preserve the workflow-only recovery cases, and report only candidates that actually moved. Required CI and CLA are green.

ddbaron:
@santoshkumarradha Thanks for the re-confirmation. Since opening this I've had the branch running as my local control plane (up since Sep 10) and running the storage suite against a live Postgres 16 here, and both reapers have stayed quiet. 

Is there anything else you need from me before this lands, or is it just waiting on your merge window?

santoshkumarradha:
Nothing else needed from you right now. The code, CI, coverage, and CLA are all good. GitHub is still holding this behind the repository review and merge-queue gate, so it is not waiting on another change from your side.

control-plane/internal/storage/execution_records.go
```diff
@@ -1584,17 +1591,22 @@ func (ls *LocalStorage) RetryStaleWorkflowExecutions(ctx context.Context, staleA
...
-		  AND retry_count < ?
-		  AND `+tsExpr+` <= `+cutoffExpr+`
-		ORDER BY `+tsExpr+` ASC
-		LIMIT ?`, maxRetries, cutoff, limit)
+		SELECT w.execution_id
+		FROM workflow_executions w
+		LEFT JOIN executions e
+		  ON e.execution_id = w.execution_id
+		 AND e.status IN ('running', 'pending', 'queued', 'waiting')
+		WHERE w.status IN ('running', 'pending', 'queued')
```
AbirAbbas, line 1604: One gap left between this path and the reaper it now mirrors: `MarkStaleWorkflowExecutions` refuses to touch a parent while any child workflow is still non-terminal (the `NOT EXISTS (SELECT 1 FROM workflow_executions c WHERE c.parent_execution_id = w.execution_id AND c.status IN (...))` clause, plus the `approval_status != 'pending'` guard), but the retry candidate query here and the repeated predicate in the UPDATE below don't. So with `max_retries > 0`, a parent whose own clocks are stale but whose child is actively running gets reset to `pending` and `retry_count+1` while the child is still going, which re-dispatches the parent on top of live work. The reaper explicitly leaves that same row alone.

Repro on SQLite (same fixtures as retry_stale_test.go): parent workflow+execution backdated 2h, child workflow+execution with `ParentExecutionID = parent` created now. `MarkStaleWorkflowExecutions(ctx, 30*time.Minute, 100)` returns 0 and the parent stays `running`; `RetryStaleWorkflowExecutions(ctx, 30*time.Minute, 3, 100)` returns `[parent]` and the parent is `pending` with `retry_count = 1`.

Fix shape: add the same two clauses to both the candidate SELECT and the per-candidate workflow UPDATE predicate (they need to match so the re-check still closes the selection-to-update window), and pin it with a regression that runs the reaper and the retry against the same stale-parent/fresh-child pair and expects both to leave it alone. It's default-off (`max_retries` defaults to 0), but it's the one case where this path would now do the opposite of the reaper on Postgres, so I'd rather it go in with this PR.

AbirAbbas (commented):
Went through this on the merge with main against a live Postgres 14 as well as SQLite: both reapers and the retry path now prepare cleanly (reverting only sql_helpers.go brings back 42601 in seven live tests), the heartbeat-between-statements and batch regressions fail with the right symptom when the retry body is put back to main's logic, the savepoint path leaves the outer tx usable on Postgres, and the interleaving tests are stable over 30 SQLite / 10 Postgres repeats. Storage patch coverage is 85% on the changed production lines. One thing inline that I think needs to land before merge.

PR #1052, opened 2026-09-10
ci(python-sdk): enforce mypy with module-by-module allowlist ratchet (#971)

### LINKED ISSUE #971: [Python SDK] Adopt mypy in CI (module-by-module ratchet)
Split out of #620, where the reporter asked for "mypy and ruff for strict checks of async / blocking code". The ruff half shipped in #812 (`ASYNC` ruleset, enforced in CI on 3.10/3.11/3.12). The mypy half was never adopted: `sdk/python/mypy.ini` exists but nothing in `.github/workflows/sdk-python.yml` runs it, and `mypy --config-file mypy.ini agentfield/` currently reports ~194 errors across ~28 files.

## What "done" looks like

- A `mypy` step in `sdk/python` CI that is **required** to pass.
- A per-module allowlist (`[mypy-agentfield.<module>]` with `ignore_errors = True`) for modules that are not yet clean, so the gate can be turned on immediately and the allowlist shrunk module by module.
- No `# type: ignore` sprinkling to get to green — fix the type or allowlist the module.

## Suggested order

1. Land the CI step with everything allowlisted (zero behaviour change, proves the wiring).
2. Clean the async-heavy modules first, since that is the reporter's concern: `run_async.py`, `lock_utils.py`, `async_lifecycle.py`, `result_cache.py`, `async_execution_manager.py`, `http_connection_manager.py`, `memory_events.py`, `logger.py`.
3. Then `client.py` and `agent.py`, which carry most of the remaining count.

## Validation contract

- CI fails if a non-allowlisted module gains a new mypy error.
- Removing a module from the allowlist while it still has errors fails CI.
- `pip install .[dev]` provides the pinned mypy version used in CI.

Refs #620.

7vignesh:
## Summary

Adopts the mypy half of #620 as phase 1 of #971. The ruff half shipped in #812, but the mypy half was never wired up: `sdk/python/mypy.ini` existed and nothing ran it. This adds a required `mypy` step to the Python SDK CI with a module-by-module allowlist, so the gate can be turned on immediately (zero behavior change) and the allowlist shrunk module by module.

Closes #971 (phase 1). Refs #620.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Docs only
- [ ] Tests only
- [x] CI / tooling
- [ ] Breaking change

## What changed

- Pin `mypy==1.18.2` in the `dev` extras (`pyproject.toml`) and regenerate `uv.lock`.
- Add a per-module `ignore_errors = True` allowlist to `mypy.ini` for the 34 modules that currently have type errors (217 errors total), with a header comment describing the ratchet and how to shrink it.
- Add a `Type check (mypy)` step to `.github/workflows/sdk-python.yml` running `mypy --config-file mypy.ini agentfield/`.

No `# type: ignore` sprinkles were added, and no `agentfield/` source was modified.

## Validation contract (from #971)

- [x] CI fails if a non-allowlisted module gains a new mypy error. Verified: injecting a type error into a clean module (`agent_cli.py`) makes mypy exit 1.
- [x] Removing a module from the allowlist while it still has errors fails CI. Verified: removing `agentfield.types` re-surfaces its errors (exit 1).
- [x] `pip install .[dev]` provides the pinned mypy version used in CI.

## Test plan

- [x] `cd sdk/python && mypy --config-file mypy.ini agentfield/` -> `Success: no issues found in 79 source files` (exit 0)
- [x] `cd sdk/python && ruff check .` -> `All checks passed!`
- [x] `cd sdk/python && uv lock --check` -> clean after regenerating `uv.lock`
- [x] `python -c "import agentfield"` -> imports OK
- [x] Ratchet invariants verified (see Validation contract above)

## Test coverage

- [x] No production code changed; this is CI/tooling only, so coverage is unaffected.

## Checklist

- [ ] Commits are signed and follow conventional-commits style. (conventional-commits: yes; commit signing not configured in my environment)

## Related issues / PRs

Closes #971 (phase 1). Refs #620, #812.

## Notes for reviewers

- The mypy step runs in the existing `lint-and-test` matrix (Python 3.10-3.13). `mypy.ini` pins `python_version = 3.10`, so results are identical across matrix rows; running it in every row is redundant but harmless. Happy to move it to a single dedicated job if you prefer.
- Phase 2 (cleaning the async-heavy modules first, per the issue's suggested order) is intentionally out of scope for this PR to keep it a pure, reviewable wiring change.

### DIFF
--- a/sdk/python/pyproject.toml
+++ b/sdk/python/pyproject.toml
@@ -100,6 +100,7 @@ dev = [
     "syrupy>=5,<6",
     "hypothesis>=6.88,<7",
     "pytest-socket>=0.6,<0.8",
+    "mypy==1.18.2",
 ]
 
 [tool.pytest.ini_options]
--- a/.github/workflows/sdk-python.yml
+++ b/.github/workflows/sdk-python.yml
@@ -56,6 +56,11 @@ jobs:
         run: |
           ruff check .
 
+      - name: Type check (mypy)
+        working-directory: sdk/python
+        run: |
+          mypy --config-file mypy.ini agentfield/
+
       - name: Format (auto-fix)
         working-directory: sdk/python
         run: |
--- a/sdk/python/mypy.ini
+++ b/sdk/python/mypy.ini
@@ -12,3 +12,113 @@ ignore_missing_imports = True
 
 [mypy-tests.*]
 disallow_untyped_defs = False
+
+# --- mypy ratchet allowlist (issue #971) ---
+# These modules are not yet type-clean. The CI gate runs mypy on the whole
+# package with these modules' errors suppressed, so the gate can be enforced
+# now and the allowlist shrunk module by module. To clean a module: fix its
+# type errors, then delete its section below. Do NOT add "# type: ignore"
+# sprinkles to reach green. New non-allowlisted modules must stay clean.
+
+[mypy-agentfield.agent]
+ignore_errors = True
+
+[mypy-agentfield.agent_ai]
+ignore_errors = True
+
+[mypy-agentfield.agent_discovery]
+ignore_errors = True
+
+[mypy-agentfield.agent_server]
+ignore_errors = True
+
+[mypy-agentfield.agent_serverless]
+ignore_errors = True
+
+[mypy-agentfield.agent_utils]
+ignore_errors = True
+
+[mypy-agentfield.agent_vc]
+ignore_errors = True
+
+[mypy-agentfield.agent_workflow]
+ignore_errors = True
+
+[mypy-agentfield.async_execution_manager]
+ignore_errors = True
+
+[mypy-agentfield.cancel]
+ignore_errors = True
+
+[mypy-agentfield.client]
+ignore_errors = True
+
+[mypy-agentfield.connection_manager]
+ignore_errors = True
+
+[mypy-agentfield.crypto]
+ignore_errors = True
+
+[mypy-agentfield.decorators]
+ignore_errors = True
+
+[mypy-agentfield.did_auth]
+ignore_errors = True
+
+[mypy-agentfield.did_manager]
+ignore_errors = True
+
+[mypy-agentfield.harness._cli]
+ignore_errors = True
+
+[mypy-agentfield.harness._runner]
+ignore_errors = True
+
+[mypy-agentfield.harness.providers.aforge]
+ignore_errors = True
+
+[mypy-agentfield.harness.providers.grok]
+ignore_errors = True
+
+[mypy-agentfield.harness.providers.opencode]
+ignore_errors = True
+
+[mypy-agentfield.harness.providers.pi]
+ignore_errors = True
+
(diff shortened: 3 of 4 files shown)


sdk/python/pyproject.toml
```diff
@@ -100,6 +100,7 @@ dev = [
     "syrupy>=5,<6",
     "hypothesis>=6.88,<7",
     "pytest-socket>=0.6,<0.8",
+    "mypy==1.18.2",
```
santoshkumarradha, line 103: Please include the typing support needed by the locked dev environment (for example `types-requests`) or explicitly configure these existing imports. After `uv sync --python python3.13 --extra dev`, the exact new command fails in `agentfield/multimodal.py` and `agentfield/agent_field_handler.py` with `[import-untyped]` errors for `requests`. GitHub is green because the workflow uses `pip install .[dev]` and resolved a newer unpinned `requests` than `uv.lock`, so it is not validating the reproducible environment this PR updates.

santoshkumarradha (changes requested):
The ratchet structure makes sense, but the checked-in environment cannot currently run the new gate.

7vignesh:
Fixed in e57eaf74. You're right - the gate wasn't reproducible.

Reproduced exactly: after `uv sync --python 3.13 --extra dev`, the locked `requests 2.33.1` ships no stubs, so mypy failed with `[import-untyped]` in `multimodal.py` and `agent_field_handler.py`. CI was green only because `pip install .[dev]` pulled a newer `requests` that bundles its own stubs.

Went with your first suggestion: pinned `types-requests==2.33.0.20260906` in the `dev` extras and regenerated `uv.lock`. I chose the stub package over allowlisting those two modules because it keeps them type-clean (allowlisting would suppress all their errors, not just the requests import), and over switching CI to `uv sync` because the stub pin fixes both paths without restructuring the workflow.

Verified mypy `Success: no issues found` against both the locked `requests 2.33.1` and a newer pip-resolved `requests`; `ruff check` clean; `uv lock --check` clean.

santoshkumarradha (approved):
Thanks for fixing the reproducible typing environment. I rechecked the updated head with Python 3.13: the locked dev sync, full mypy gate, Ruff, and diff check all pass. The new allowlist entry matches the module added on current main, so this is ready from review.

AbirAbbas (approved):
Ran the workflow's exact steps on the merge with main in three environments: fresh pip 3.10, fresh pip 3.13, and the locked `uv sync --python 3.13 --extra dev` env Santosh used. ruff 0.15.22, `mypy --config-file mypy.ini agentfield/` (80 files, clean) and run_pytest.sh pass in all of them, `uv lock --check` is clean, and the types-requests pin installs on 3.10/3.11/3.12/3.13. The gate is real: a type error injected into an unlisted module fails CI, one in a listed dirty module is suppressed, which is the ratchet contract. All 35 listed modules exist on main. Good to merge.

PR #1058, opened 2026-09-16
feat(examples): add optional Serply search backend to deep_research

googio:
## Summary

The Deep Research example supports Tavily by default and Parallel Search by explicit opt-in (#863). This adds a third explicit option, `SEARCH_PROVIDER=serply`, which answers from Google's result pages. The default Tavily path is untouched, and nothing changes for anyone who does not set the variable.

Serply is a distinct surface rather than a third general index: Tavily and Parallel run their own crawls, while Serply returns what Google returns for the query. For a research agent that is the difference between "what this index has" and "what a person searching would see".

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Docs only
- [ ] Tests only
- [ ] CI / tooling
- [ ] Breaking change

## How it is built

Same shape as the Parallel backend, so there is nothing new to learn in this file:

- `_serply_results` normalizes one response into the existing `{title, url, content, raw_content}` rows, skipping entries with no usable link or title.
- `_execute_serply_search` is one async `aiohttp` call per query, with per-query failure isolation into `{"error": ..., "query": ...}`.
- One `elif` in `execute_search`. The existing `if/else` became `provider = _search_provider()` plus three branches so the variable is read once.
- No `requirements.txt` change. `aiohttp` is imported inside the function, matching the Parallel path exactly.

Two behaviours of this API are handled explicitly rather than assumed, both of them found by calling it:

1. **`num` is an approximate upper bound, ceiling 10.** A response can carry more rows than requested, so the client asks for the ceiling and trims to the example's own five-result budget instead of trusting the count it gets back. That keeps the result volume level with the Tavily path, which uses `max_results=5`.
2. **Errors arrive as a non-2xx with a JSON `detail` string.** `_serply_error` surfaces that text, so a bad key reports `Serply returned HTTP status 401: Invalid API key` rather than a bare status code. It falls back to the plain status line when the body is missing, empty, or not JSON.

## Test plan

- [x] `PYTHONDONTWRITEBYTECODE=1 python3 -m unittest discover -s examples/python_agent_nodes/deep_research/tests -p 'test_*.py' -v` (16 tests, all pass; 7 of them new)
- [x] `git diff --check` and `python3 -m compileall -q examples/python_agent_nodes/deep_research`
- [x] `ruff check` and `ruff format --check` on `examples/python_agent_nodes/deep_research` (ruff 0.15.1, clean)
- [x] Live request to `https://api.serply.io/v1/search`, driven through `execute_search` itself rather than a fixture

The new tests mirror the Parallel ones in `tests/test_search_backends.py`: provider selection, normalization and skipping of unusable rows, malformed payload rejection, the error message builder, dispatch routing, and one fake-`aiohttp` case that asserts the request parameters, the `X-Api-Key` header, the trim to five, and failure isolation across two queries.

The live run, for the record:

```text
execute_search(["agentfield agent backend", "deep research agents"]) with SEARCH_PROVIDER=serply
  -> 10 rows, every row exactly {title, url, content, raw_content}
  -> 5 per query, so the client-side trim is doing its job

_execute_serply_search(["x"]) with a deliberately wrong key
  -> [{'error': 'Serply returned HTTP status 401: Invalid API key', 'query': 'x'}]
```

## Test coverage

`coverage.yml` filters on `control-plane/**`, `sdk/**` and the coverage scripts, and its own comment names "examples-only PRs" as a case it skips while still reporting the required check as passed. `sdk-python.yml` is filtered to `sdk/python/**`. So this PR does not touch a covered surface and no CI job runs these tests. That is exactly why they were run locally and why the live call is quoted above rather than summarized.

- [x] I did not lower or otherwise change `coverage-baseline.json`.
- [x] The coverage gate check is expected green by path filter, not by silencing.

## Checklist

- [x] I have read [docs/CONTRIBUTING.md](docs/CONTRIBUTING.md) and [docs/DEVELOPMENT.md](docs/DEVELOPMENT.md).
- [x] Commits follow conventional-commits style. They are not GPG-signed; the `main` ruleset has no signature rule and #863 merged carrying an unsigned commit, but say the word and I will re-sign.
- [ ] I have linked any related issues (none; this follows #863 rather than an issue).

## Related issues / PRs

Follows #863, which introduced `SEARCH_PROVIDER` and the multi-backend structure this slots into.

Disclosure: I work with Serply, which operates this endpoint. Happy to narrow the scope, drop the `main.py` label line, or close this if a third backend is not a direction you want for the example.


santoshkumarradha (approved):
This is a clean optional backend and keeps Tavily as the default. The response normalization, API-key handling, per-query error isolation, client-side result cap, and provider routing are covered well. I also ran the 16 focused tests, Ruff, formatting, and the diff check locally. The fork workflows are approved and running now.

PR #1062, opened 2026-09-18
fix(storage): mirror reaper parent/child and approval guards in stale retry

ddbaron:
Follow-up to #1051, from the merge-time review finding by @AbirAbbas (2026-09-18T14:17:45Z, inline on `control-plane/internal/storage/execution_records.go:1604`): `RetryStaleWorkflowExecutions` mirrors the reaper's clocks and staleness predicates but not its parent/child and approval-status guards.

## Bug

`MarkStaleWorkflowExecutions` refuses to touch a parent while any child workflow is non-terminal and skips rows whose `approval_status` is pending. The retry candidate query and its per-candidate workflow UPDATE had neither guard, so with `max_retries > 0` a parent whose own clocks are stale but whose child is still running was reset to `pending` with `retry_count` incremented and re-dispatched on top of live work — the same row the reaper explicitly leaves alone.

Repro (from the review, SQLite, same fixtures as `retry_stale_test.go`): parent workflow+execution backdated 2h, child workflow+execution with `ParentExecutionID` set to the parent created now. `MarkStaleWorkflowExecutions(ctx, 30*time.Minute, 100)` returns 0 and the parent stays `running`; `RetryStaleWorkflowExecutions(ctx, 30*time.Minute, 3, 100)` returned `[parent]` and left it `pending` with `retry_count = 1`.

## Fix

Add the reaper's two guards to `RetryStaleWorkflowExecutions` in both the candidate SELECT and the per-candidate workflow UPDATE (matching clauses so the re-check still closes the selection-to-update window):

- `COALESCE(w.approval_status, '') != 'pending'`
- `NOT EXISTS (SELECT 1 FROM workflow_executions c WHERE c.parent_execution_id = w.execution_id AND c.status IN ('running', 'pending', 'queued', 'waiting'))`

The clauses are byte-identical to `MarkStaleWorkflowExecutions`'s (all four occurrences hash to one byte sequence) and placed at the same spots, so the two paths cannot drift. Nothing else changed in the retry statements: RowsAffected gating, retry-count semantics, savepoint behavior, and the waiting-state exclusion are untouched. The diff touches only `execution_records.go` and `retry_stale_test.go`.

Regressions:
- `TestStaleParentWithLiveChildIsSparedByReaperAndRetry` — the review's repro as a test: the reaper and the retry run against the same stale-parent/fresh-child fixture and both must refuse the parent (status, `retry_count`, error message, `completed_at`, `updated_at`, and the paired execution all checked). Proven to bite: with the retry guards reverted it fails with `Should be empty, but was [exec-parent-live-child]`.
- `TestRetryStaleWorkflowExecutions_ApprovalStatusGuard` — a pending approval is spared while a decided (`approved`) approval is still retried.
- `TestRetryStaleWorkflowExecutions_TerminalChildDoesNotShieldParent` — a terminal child is not live work, so its stale parent stays retriable (guards against stranding parents forever).

## How this change was produced

Implemented via a managed SWE-AF `swe-planner.implement_issue` run (execution `exec_20260918_164238_g3umxbyj`, succeeded) with deeper QA enabled, then integrated and independently re-validated on this branch:

- **Coder** (`openrouter/deepseek/deepseek-v4.1-flash#max`): mirrored both guards into the SELECT and UPDATE byte-identically (verified by diff/cmp against the reaper's clauses), added the stale-parent/live-child regression; focused + full SQLite suite pass, gofmt/go vet clean, and the pre-fix bite proof recorded.
- **Code reviewer** (`openrouter/meta/muse-spark-1.3-contributor#xhigh`): approved, no blocking issues — byte-identical clauses in required positions, only the two allowed files touched, retry-count/savepoint/waiting-state semantics preserved.
- **QA** (`openrouter/z-ai/glm-5.3-flash#max`): passed; byte-compared all 4 guard occurrences into one unique byte sequence, verified placement and unchanged semantics, and filled the two coverage gaps above (approval-status, terminal-child boundary).
- **Synthesis**: APPROVE; verifier pass against the acceptance criteria passed.

## Tests (validation of the integrated branch)

- SQLite: `go test ./internal/storage/` full package — pass; the three new regressions and the focused reaper/retry set — all pass; `go vet` and `gofmt` clean on the touched files.
- Live PostgreSQL 16 (`pgvector/pgvector:pg16` in a local throwaway loopback container, per-test throwaway databases, container and volume removed afterwards): `TestPostgresStaleReaper*`, `TestPostgresMarkStale*`, `TestPostgresRetryStale*` — all pass, including `TestPostgresStaleReaperPreparedStatementsAreRebound`, which drives the modified retry statements through PostgreSQL prepare, and the heartbeat-between-statements / batch twins that exercise the changed UPDATE end to end.


santoshkumarradha (approved):
This cleanly mirrors the reaper guards in both retry phases, and the regression coverage checks the important live-child, pending-approval, and terminal-child boundaries. The full storage test package, vet, and diff checks pass locally, and GitHub CI plus CLA are green.

